Issue #124, opened 2021-03-12
Invalid Stack name suggested when csproj file has non-english characters


ppittle:
**Describe the bug**
Tool will suggest an invalid AWS Stack name if the csproj file has spaces.

For example: `withSpecial!こんにちは世界Characters` receives a suggested AWS Stack name of `with spaces`.   That suggestion is not valid:

```
Unhandled exception. Amazon.JSII.Runtime.JsiiException: Stack name must match the regular expression: 
/^[A-Za-z][A-Za-z0-9-]*$/, got 'withSpecial!こんにちは世界Characters'
```

**To Reproduce**
1. `mkdir "withSpecial!こんにちは世界Characters"`
2. `cd "withSpecial!こんにちは世界Characters"`
3. `dotnet new mvc`
4. `dotnet aws deploy`
5. press <kbd>enter</kbd> to accept recommended Aws Stack Name
5. confirm suggestion of Elastic Beanstalk recipe
5. press <kbd>enter</kbd> to accept default recommendations

**Logs**
n/a

**Expected behavior**
tool can deploy `withSpecial!こんにちは世界Characters.csproj` to elastic beanstalk.

**Environment (please complete the following information):**
 - CLI Version: [eg. 1.0.0] `0.1.1-beta.20210312012242.78ce3cc`
 - Include the output of `dotnet --info` `5.0.201`
 - OS Info: `Windows 10`

**Additional context**
n/a

PR #169, opened 2021-03-23
bug/4878 validate application name

ppittle:
- [x] Needs #171 need to merge first

image

*Issue #, if available:*
- #124 
- #123 
- DOTNET-4878

*Description of changes:*
 sanitize cloud application name recommendations and also validate user input

### DIFF
--- a/src/AWS.Deploy.Orchestration/Utilities/CloudApplicationNameGenerator.cs
+++ b/src/AWS.Deploy.Orchestration/Utilities/CloudApplicationNameGenerator.cs
@@ -1,0 +1,87 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.\r
+// SPDX-License-Identifier: Apache-2.0
+
+using System;
+using System.Collections.Generic;
+using System.IO;
+using System.Linq;
+using System.Text.RegularExpressions;
+using AWS.Deploy.Common;
+using AWS.Deploy.Common.IO;
+
+namespace AWS.Deploy.Orchestration.Utilities
+{
+    public interface ICloudApplicationNameGenerator
+    {
+        /// <summary>
+        /// Generates a valid candidate for <see cref="CloudApplication.Name"/> based on <paramref name="target"/>.
+        /// Name  is checked to ensure it is unique, ie <paramref name="existingApplications"/> doesn't already have it.
+        /// </summary>
+        /// <exception cref="ArgumentException">
+        /// Thrown if can't generate a valid name from <paramref name="target"/>.
+        /// </exception>
+        string GenerateValidName(ProjectDefinition target, List<CloudApplication> existingApplications);
+
+        /// <summary>
+        /// https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/cfn-using-console-create-stack-parameters.html
+        /// </summary>
+        bool IsValidName(string name);
+    }
+
+    public class CloudApplicationNameGenerator : ICloudApplicationNameGenerator
+    {
+        private readonly IFileManager _fileManager;
+        private readonly IDirectoryManager _directoryManager;
+        /// <summary>
+        /// https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/cfn-using-console-create-stack-parameters.html 
+        /// </summary>
+        private readonly Regex _validatorRegex = new ("^[a-zA-Z][a-zA-Z0-9-]{0,127}$", RegexOptions.Compiled);
+
+        public CloudApplicationNameGenerator(IFileManager fileManager, IDirectoryManager directoryManager)
+        {
+            _fileManager = fileManager;
+            _directoryManager = directoryManager;
+        }
+
+        public string GenerateValidName(ProjectDefinition target, List<CloudApplication> existingApplications)
+        {
+            // generate recommendation
+            var recommendedPrefix = "deployment";
+
+            if (_fileManager.Exists(target.ProjectPath))
+                recommendedPrefix = Path.GetFileNameWithoutExtension(target.ProjectPath);
+            else if (_directoryManager.Exists(target.ProjectPath))
+                recommendedPrefix = Path.GetDirectoryName(target.ProjectPath);
+
+            // sanitize recommendation
+            recommendedPrefix =
+                new string(
+                    recommendedPrefix
+                        .ToCharArray()
+                        .SkipWhile(c => !char.IsLetter(c) && ((int)c) < 127)
+                        .Where(c =>
+                            char.IsNumber(c) ||
+                            char.IsLetter(c) && ((int)c) < 127 ||
+                            c == '-')
+                        .ToArray());
+
+            // make sure the recommendation doesn't exist already in existingApplications
+            var recommendation = recommendedPrefix;
+            var suffix = 1;
+            while (suffix < 100)
+            {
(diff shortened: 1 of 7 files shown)


philasmar (approved):

src/AWS.Deploy.Orchestration/Utilities/CloudApplicationNameGenerator.cs
```diff
@@ -0,0 +1,87 @@
...
+            // make sure the recommendation doesn't exist already in existingApplications
+            var recommendation = recommendedPrefix;
+            var suffix = 1;
+            while (suffix < 100)
+            {
+                if (existingApplications.All(x => x.StackName != recommendation) && IsValidName(recommendation))
+                    return recommendation;
+
+                recommendation = $"{recommendedPrefix}{suffix++}";
+            }
```
ganeshnj, line 77: We should use something like GUID which is unique but shorter in length. Perhaps first two segments of GUID or last segment should be good enough for our use case.

src/AWS.Deploy.Orchestration/Utilities/CloudApplicationNameGenerator.cs
```diff
@@ -0,0 +1,87 @@
...
+            _directoryManager = directoryManager;
+        }
+
+        public string GenerateValidName(ProjectDefinition target, List<CloudApplication> existingApplications)
+        {
+            // generate recommendation
+            var recommendedPrefix = "deployment";
+
+            if (_fileManager.Exists(target.ProjectPath))
+                recommendedPrefix = Path.GetFileNameWithoutExtension(target.ProjectPath);
```
ganeshnj, line 52: Customer can have a directory starting with - (dash) or number which isn't accepted by CloudFormation.

src/AWS.Deploy.Orchestration/Utilities/CloudApplicationNameGenerator.cs
```diff
@@ -0,0 +1,87 @@
...
+        /// </summary>
+        private readonly Regex _validatorRegex = new ("^[a-zA-Z][a-zA-Z0-9-]{2,126}$", RegexOptions.Compiled);
+
+        public CloudApplicationNameGenerator(IFileManager fileManager, IDirectoryManager directoryManager)
+        {
+            _fileManager = fileManager;
+            _directoryManager = directoryManager;
+        }
+
+        public string GenerateValidName(ProjectDefinition target, List<CloudApplication> existingApplications)
```
ganeshnj, line 46: There is no check on the length of the application name.
```
The stack name is an identifier that helps you find a particular stack from a list of stacks. A stack name can contain only alphanumeric characters (case-sensitive) and hyphens. It must start with an alphabetic character and can't be longer than 128 characters.
```

https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/cfn-using-console-create-stack-parameters.html

I think the generated name should match `^[a-zA-Z][-a-zA-Z0-9]{1,128}$`, I'm not saying to use regex.

ganeshnj (changes requested):

src/AWS.Deploy.Orchestration/Utilities/CloudApplicationNameGenerator.cs
ganeshnj: We can do the same validation on customer entered application for early feedback than trying to deploy and fail.

src/AWS.Deploy.Orchestration/Utilities/CloudApplicationNameGenerator.cs
ppittle: You are correct.  Down below the `recommendedPrefix` is [sanitized](https://github.com/aws/aws-dotnet-deploy/pull/169/files#diff-a0db7211bc877b00e7c00a810aa3a6a1459080bdc78592fed5541b10b404759bR66) to remove invalid characters at the start of the recommendation.


Here's the test case covering it: https://github.com/aws/aws-dotnet-deploy/pull/169/files#diff-a0db7211bc877b00e7c00a810aa3a6a1459080bdc78592fed5541b10b404759bR66

src/AWS.Deploy.Orchestration/Utilities/CloudApplicationNameGenerator.cs
ppittle: 1.  Length: `_validatorRegex` does check length: `{2,126}`  I was a little more restrictive than the regex you mentioned which uses `{1,128}`.
2. Early feedback.  Feedback and validation is done as the user enters/selects an Application Name.  They can't even see recommendations until a valid application name is selected.  See the screenshot in the PR Description.

src/AWS.Deploy.Orchestration/Utilities/CloudApplicationNameGenerator.cs
ppittle: Are you recommending this?

| Current | Guid Proposal |
| --- | --- |
| Select the AWS stack: <li>WebApp</li><li>WebApp1</li><li>WebApp2</li>| Select the AWS stack: <li>WebApp</li><li>WebApp-de8bab2b</li><li>WebApp-e8da26de</li> |

I think I'd rather stick with the numbers.  I find it a bit more approachable and it has a natural sort order.

Do you want to make the case for the Guid Proposal?

src/AWS.Deploy.Orchestration/Utilities/CloudApplicationNameGenerator.cs
ppittle: Initial regex was a bit too conservative.  Expanded to allow Names are short as 1 character and as long as 128

src/AWS.Deploy.Orchestration/Utilities/CloudApplicationNameGenerator.cs
ganeshnj: Can we check this with team just to have second opinion?

normj: I'm with PJ on this. This is a user facing name that the user might have to type in. Nobody wants to type in a guid. It also follows the pattern of new projects creating in VS.

src/AWS.Deploy.Orchestration/Utilities/CloudApplicationNameGenerator.cs
```diff
@@ -0,0 +1,87 @@
...
+            // sanitize recommendation
+            recommendedPrefix =
+                new string(
+                    recommendedPrefix
+                        .ToCharArray()
+                        .SkipWhile(c => !char.IsLetter(c) && ((int)c) < 127)
+                        .Where(c =>
+                            char.IsNumber(c) ||
+                            char.IsLetter(c) && ((int)c) < 127 ||
+                            c == '-')
```
ganeshnj, line 65: this makes me more comfortable
`
'a' <= c <= 'z' || 'A' <= c <= 'Z' || '0' <= c <= '9' || c == '-'
`

ganeshnj (approved):

src/AWS.Deploy.Orchestration/Utilities/CloudApplicationNameGenerator.cs
ppittle: interesting suggestion, but think I'm going to leave it as is

PR #184, opened 2021-04-05
Feature/recipe info user agent metric

ganeshnj:
*Issue #, if available:*

*Description of changes:*
`AWS_EXECUTION_ENV` value for a Beanstalk deployment: `aws-dotnet-deploy-cli_0.4.2_AspNetAppElasticBeanstalkLinux_0.1.0`

### DIFF
--- a/src/AWS.Deploy.CLI/Program.cs
+++ b/src/AWS.Deploy.CLI/Program.cs
@@ -222,7 +222,6 @@ namespace AWS.Deploy.CLI
         /// </summary>
         private static void SetExecutionEnvironment()
         {
-            const string envName = "AWS_EXECUTION_ENV";
             const string awsDotnetDeployCLI = "aws-dotnet-deploy-cli";
 
             var assemblyVersion = typeof(Program).Assembly
@@ -233,14 +232,14 @@ namespace AWS.Deploy.CLI
             var envValue = new StringBuilder();
 
             // If there is an existing execution environment variable add this tool as a suffix.
-            if(!string.IsNullOrEmpty(Environment.GetEnvironmentVariable(envName)))
+            if(!string.IsNullOrEmpty(Environment.GetEnvironmentVariable(EnvironmentVariableKeys.AWS_EXECUTION_ENV)))
             {
-                envValue.Append($"{Environment.GetEnvironmentVariable(envName)}_");
+                envValue.Append($"{Environment.GetEnvironmentVariable(EnvironmentVariableKeys.AWS_EXECUTION_ENV)}_");
             }
 
             envValue.Append($"{awsDotnetDeployCLI}_{assemblyVersion?.InformationalVersion}");
 
-            Environment.SetEnvironmentVariable(envName, envValue.ToString());
+            Environment.SetEnvironmentVariable(EnvironmentVariableKeys.AWS_EXECUTION_ENV, envValue.ToString());
         }
     }
 }
--- a/src/AWS.Deploy.CLI/Utilities/CommandLineWrapper.cs
+++ b/src/AWS.Deploy.CLI/Utilities/CommandLineWrapper.cs
@@ -2,12 +2,14 @@
 // SPDX-License-Identifier: Apache-2.0
 
 using System;
+using System.Collections.Generic;
 using System.Diagnostics;
 using System.Runtime.InteropServices;
 using System.Text;
 using System.Threading;
 using System.Threading.Tasks;
 using Amazon.Runtime;
+using AWS.Deploy.Common;
 using AWS.Deploy.Orchestration;
 using AWS.Deploy.Orchestration.Utilities;
 
@@ -36,6 +38,7 @@ namespace AWS.Deploy.CLI.Utilities
             bool streamOutputToInteractiveService = true,
             Action<TryRunResult> onComplete = null,
             bool redirectIO = true,
+            IDictionary<string, string> environmentVariables = null,
             CancellationToken cancelToken = default)
         {
             StringBuilder strOutput = new StringBuilder();
@@ -64,6 +67,8 @@ namespace AWS.Deploy.CLI.Utilities
             processStartInfo.EnvironmentVariables["AWS_ACCESS_KEY_ID"] = credentials.AccessKey;
             processStartInfo.EnvironmentVariables["AWS_SECRET_ACCESS_KEY"] = credentials.SecretKey;
             processStartInfo.EnvironmentVariables["AWS_REGION"] = _awsRegion;
+
+            UpdateEnvironmentVariables(processStartInfo, environmentVariables);
 
             if (credentials.UseToken)
             {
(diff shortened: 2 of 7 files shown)


src/AWS.Deploy.CLI/Utilities/CommandLineWrapper.cs
```diff
@@ -114,6 +121,27 @@ public class CommandLineWrapper : ICommandLineWrapper
             }
         }
 
+        private static string BuildAWSExecutionEnvValue(string awsExecutionEnv, ProcessStartInfo processStartInfo)
+        {
+            var awsExecutionEnvBuilder = new StringBuilder();
+            if (processStartInfo.EnvironmentVariables.ContainsKey(EnvironmentVariableKeys.AWS_EXECUTION_ENV)
+                && !string.IsNullOrEmpty(processStartInfo.EnvironmentVariables[EnvironmentVariableKeys.AWS_EXECUTION_ENV]))
+            {
+                awsExecutionEnvBuilder.Append(processStartInfo.EnvironmentVariables[EnvironmentVariableKeys.AWS_EXECUTION_ENV]);
```
ppittle, line 130: do you need some kind of separator?  Space character or semi colan perhaps?

src/AWS.Deploy.Orchestration/EnvironmentVariableKeys.cs
```diff
@@ -0,0 +1,13 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+namespace AWS.Deploy.Orchestration
+{
+    /// <summary>
+    /// Contains shared environment variable keys
+    /// </summary>
+    public static class EnvironmentVariableKeys
```
ppittle, line 9: Would this make more sense in Common?  Is it specific to Orchestrator?

src/AWS.Deploy.CLI/Utilities/CommandLineWrapper.cs
```diff
@@ -36,6 +36,7 @@ public class CommandLineWrapper : ICommandLineWrapper
             bool streamOutputToInteractiveService = true,
             Action<TryRunResult> onComplete = null,
             bool redirectIO = true,
+            string awsExecutionEnv = "",
```
ppittle, line 39: should `awsExecutionEnv` really be passed in as a method parameter?  Wouldn't it always be the same for the lifetime of the `CommandLineWrapper`?  Would it make more sense to treat it like `_awsCredentials` or `_awsRegion` and pass it to the constructor?

src/AWS.Deploy.Orchestration/Utilities/ICommandLineWrapper.cs
```diff
@@ -41,6 +45,7 @@ public interface ICommandLineWrapper
             bool streamOutputToInteractiveService = true,
             Action<TryRunResult> onComplete = null,
             bool redirectIO = true,
+            string awsExecutionEnv = "",
```
ppittle, line 48: I'm thinking `awsExecutionEnv` makes more sense as a constructor parameter to `CommandLineWrapper` (there's a separate thread for that discussion).

However, _if_ you do keep `awsExecutionEnv` as method parameter, think you should also update `TryRunWithResults` for consistency.

src/AWS.Deploy.CLI/Utilities/CommandLineWrapper.cs
ppittle: Looked over usage, you're passing in Recipe details, which aren't known until after the user selects a recipe.  

Hmm ... still feels like an odd method parameter on `ICommandLineWrapper.Run` though.  But not sure what to advise.

src/AWS.Deploy.CLI/Utilities/CommandLineWrapper.cs
ganeshnj: we separate using _, check the description, it has example.

src/AWS.Deploy.Orchestration/EnvironmentVariableKeys.cs
ganeshnj: it is also used in CLI but given it has reference to Orchestrator it works. It could be moved to Common. I will do it.

src/AWS.Deploy.CLI/Utilities/CommandLineWrapper.cs
ppittle: Do you need a `.Append("_")`?

ganeshnj: We don't want _ at the beginning. StringBuilder takes the existing value from env var, and then add _ if there is an already specified AWS_EXECUTION_ENV.

This append is basically taking the existing value, there is no need of append.

src/AWS.Deploy.Orchestration/CdkProjectHandler.cs
```diff
@@ -27,6 +27,8 @@ public CdkProjectHandler(IOrchestratorInteractiveService interactiveService, ICo
 
         public async Task CreateCdkDeployment(OrchestratorSession session, CloudApplication cloudApplication, Recommendation recommendation)
         {
+            var recipeInfo = $"{recommendation.Recipe.Id}_{recommendation.Recipe.Version}";
```
philasmar, line 30: we should think about storing this info globally without relying on the developer to pass this every time they do a Run in CommandLineWrapper

src/AWS.Deploy.CLI/Utilities/CommandLineWrapper.cs
ganeshnj: Storing a parameter is causing another problem of state management. I don't see saving this value to a shared object a workable solution.

src/AWS.Deploy.Orchestration/CdkProjectHandler.cs
ganeshnj: recipe info is passed only for `cdk deploy`, we don't have any other-use also where we want to pass it.

See https://github.com/aws/aws-dotnet-deploy/pull/184/files#r610240199 for shared state.

src/AWS.Deploy.CLI/Program.cs
```diff
@@ -222,7 +222,6 @@ private static async Task<int> Main(string[] args)
         /// </summary>
         private static void SetExecutionEnvironment()
         {
-            const string envName = "AWS_EXECUTION_ENV";
             const string awsDotnetDeployCLI = "aws-dotnet-deploy-cli";
```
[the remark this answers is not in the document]
ganeshnj: Well no, this is the value which doesn't belong to keys.

ppittle (approved):

philasmar (approved):

PR #199, opened 2021-04-26
fix: Child option settings IDs are displayed instead of names and DependsOn property not respected

philasmar:
*Issue #, if available:*
* Child option settings IDs are displayed instead of names
* DependsOn property not respected in Child option settings
* Child option settings of type Bool are not being displayed

*Description of changes:*
**Before**
PR199-Before

**After**
PR199-After

### DIFF
--- a/src/AWS.Deploy.Common/Recommendation.cs
+++ b/src/AWS.Deploy.Common/Recommendation.cs
@@ -107,12 +107,28 @@ namespace AWS.Deploy.Common
 
         public T GetOptionSettingValue<T>(OptionSettingItem optionSetting, bool ignoreDefaultValue = false)
         {
-            return optionSetting.GetValue<T>(_replacementTokens, ignoreDefaultValue);
+            var displayableOptionSettings = new Dictionary<string, bool>();
+            if (optionSetting.Type == OptionSettingValueType.Object)
+            {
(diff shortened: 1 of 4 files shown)


src/AWS.Deploy.Common/Recommendation.cs
```diff
@@ -105,14 +105,30 @@ public OptionSettingItem GetOptionSetting(string jsonPath)
             return optionSetting;
         }
 
-        public T GetOptionSettingValue<T>(OptionSettingItem optionSetting, bool ignoreDefaultValue = false)
+        public T GetOptionSettingValue<T>(OptionSettingItem optionSetting, bool ignoreDefaultValue = false, bool onlyDisplayableChildren = false)
```
ganeshnj, line 108: Can you add tests that covers the new param.

ganeshnj (changes requested):

src/AWS.Deploy.Common/Recommendation.cs
philasmar: I added 2 tests in this PR https://github.com/aws/aws-dotnet-deploy/pull/200.
We don't have an option setting that needs this in this PR.

normj (approved):

ganeshnj (approved):

PR #210, opened 2021-05-10
feat: enable nullability and set C# version to 9

philasmar:
*Issue #, if available:*
https://sim.amazon.com/issues/DOTNET-5005

*Description of changes:*
enable nullability and set C# version to 9


src/AWS.Deploy.Common/Recipes/AvailableRuleItem.cs
```diff
@@ -1,37 +0,0 @@
...
-// SPDX-License-Identifier: Apache-2.0
-
-using System.Collections.Generic;
-
-namespace AWS.Deploy.Common.Recipes
-{
-    /// <summary>
-    /// Container for the types of rules that can be checked.
-    /// </summary>
-    public class AvailableRuleItem
```
ganeshnj, line 11: where did this go?

src/AWS.Deploy.Common/Recipes/MSPropertyRule.cs
```diff
@@ -1,23 +0,0 @@
-﻿// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
-// SPDX-License-Identifier: Apache-2.0
-
-using System.Collections.Generic;
-
-namespace AWS.Deploy.Common.Recipes
```
ganeshnj, line 6: and this also, aren't we using them?

src/AWS.Deploy.Recipes/CdkTemplates/AspNetAppEcsFargate/Configurations/VpcConfiguration.cs
```diff
@@ -20,5 +20,22 @@ public class VpcConfiguration
         /// then use an existing VPC by referencing through <see cref="VpcId"/>
         /// </summary>
         public string VpcId { get; set; }
+
+#nullable disable warnings
```
ganeshnj, line 24: Can you add a comment why we disable the warning?

src/AWS.Deploy.Recipes/CdkTemplates/AspNetAppElasticBeanstalkLinux/Configurations/BeanstalkApplicationConfiguration.cs
```diff
@@ -7,5 +7,20 @@ public class BeanstalkApplicationConfiguration
     {
         public bool CreateNew { get; set; }
         public string ApplicationName { get; set; }
+
+#nullable disable warnings
```
ganeshnj, line 11: Can you add a comment why we disable the warning?

ganeshnj (changes requested):

normj (approved):
Approve assuming you address Ganesh comments.

src/AWS.Deploy.CLI/Commands/TypeHints/BeanstalkApplicationCommand.cs
```diff
@@ -27,21 +27,20 @@ public async Task<object> Execute(Recommendation recommendation, OptionSettingIt
...
             var userResponse = _consoleUtilities.AskUserToChooseOrCreateNew(applications, "Select Elastic Beanstalk application to deploy to:", userInputConfiguration);
 
-            return new BeanstalkApplicationTypeHintResponse
-            {
-                CreateNew = userResponse.CreateNew,
-                ApplicationName = userResponse.SelectedOption?.ApplicationName ?? userResponse.NewName
-            };
+            return new BeanstalkApplicationTypeHintResponse(
+                userResponse.CreateNew,
+                userResponse.SelectedOption?.ApplicationName ?? userResponse.NewName!
```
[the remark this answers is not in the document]
philasmar: I have updated it to now throw an exception if a new name is not provided.

src/AWS.Deploy.CLI/Commands/TypeHints/BeanstalkEnvironmentCommand.cs
```diff
@@ -32,8 +32,8 @@ public async Task<object> Execute(Recommendation recommendation, OptionSettingIt
                 options: environments.Select(env => env.EnvironmentName),
                 title: "Select Elastic Beanstalk environment to deploy to:",
                 askNewName: true,
-                defaultNewName: currentValue.ToString());
-            return userResponse.SelectedOption ?? userResponse.NewName;
+                defaultNewName: currentValue.ToString() ?? "");
+            return userResponse.SelectedOption ?? userResponse.NewName!;
```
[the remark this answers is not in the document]
philasmar: I have updated it to now throw an exception if a new name is not provided.

src/AWS.Deploy.Common/Recipes/AvailableRuleItem.cs
philasmar: This class was not used at all. There were no references for it. I also ran the integ tests and they passed.

src/AWS.Deploy.Common/Recipes/MSPropertyRule.cs
philasmar: This class was not used at all. There were no references for it. I also ran the integ tests and they passed.

src/AWS.Deploy.Recipes/CdkTemplates/AspNetAppEcsFargate/Configurations/VpcConfiguration.cs
philasmar: I added a comment for all classes that are doing this.

src/AWS.Deploy.Recipes/CdkTemplates/AspNetAppElasticBeanstalkLinux/Configurations/BeanstalkApplicationConfiguration.cs
philasmar: I added a comment for all classes that are doing this.

normj (approved):

src/AWS.Deploy.Orchestration/Data/AWSResourceQueryer.cs
```diff
@@ -102,7 +106,7 @@ public async Task<List<KeyPairInfo>> ListOfEC2KeyPairs()
             return response.KeyPairs;
         }
 
-        public async Task<string> CreateEC2KeyPair(string keyName, string saveLocation)
+        public async Task<string> CreateEC2KeyPair(string? keyName, string saveLocation)
```
ganeshnj, line 109: https://docs.aws.amazon.com/AWSEC2/latest/APIReference/API_CreateKeyPair.html

KeyName is required param, a nullable keyName will result in runtime exception.

src/AWS.Deploy.Orchestration/Exceptions.cs
```diff
@@ -9,7 +9,7 @@ namespace AWS.Deploy.Orchestration
     [AWSDeploymentExpectedException]
```
ganeshnj, line 9: All exceptions classes should have message also nullable like `Exception` base class.

src/AWS.Deploy.Orchestration/TemplateEngine.cs
```diff
@@ -65,7 +68,7 @@ public async Task GenerateCDKProjectFromTemplate(Recommendation recommendation,
             {
                 var currentValue = recommendation.GetOptionSettingValue(option);
                 if (currentValue != null)
```
ganeshnj, line 70: `if (currentValue != null)` isn't required.

test/AWS.Deploy.CLI.UnitTests/ApplyPreviousSettingsTests.cs
```diff
@@ -67,8 +77,8 @@ public async Task ApplyApplicationIAMRolePreviousSettings(bool createNew, string
         }
 
         [Theory]
-        [InlineData(true, false, null)]
-        [InlineData(false, true, null)]
+        [InlineData(true, false, "")]
+        [InlineData(false, true, "")]
```
ganeshnj, line 81: You are changing the intent of the test here. We should mark `vpcId` nullable here.

ganeshnj (changes requested):

src/AWS.Deploy.Orchestration/Exceptions.cs
philasmar: Even though the base class has the message as nullable, we do not want to allow the users to pass in a null message to our custom exceptions. This is another way to make sure exceptions are throwing valid messages to the end users as part of our UX improvement.

test/AWS.Deploy.CLI.UnitTests/ApplyPreviousSettingsTests.cs
philasmar: For the purpose of this test, the Assertion that vpcId will be null will always be false since GetOptionSettingValue now always returns a value. It will never be null. If you think the new test is not meaningful, we can remove it.

src/AWS.Deploy.Orchestration/TemplateEngine.cs
philasmar: done

src/AWS.Deploy.Orchestration/Data/AWSResourceQueryer.cs
philasmar: done

src/AWS.Deploy.Orchestration/Exceptions.cs
ganeshnj: I can still pass an empty string. By making it non nullable we aren't gaining anything here, it boils down to PR review.

philasmar: Right, but by having it non-nullable the intention is clear. Just because someone can work around it and pass an empty string shouldn't mean that we shouldn't be as restrictive as possible.

ganeshnj (approved):

Issue #161, opened 2021-03-18
ECS Fargate deployment deletion leaves behind the ECR repository and image


ganeshnj:
**Describe the bug**
On deletion of a ECS Fargate deployment, ECR image used for the deployment stays as it is.

**To Reproduce**
1. `dotnet new webapp -n "MyWebApp"`
2. `cd MyWebApp`
3. `dotnet aws deploy`
4. deploy with default configuration
5. `dotnet aws delete-deployment MyWebApp`
6. Observe ECR image used for MyWebApp

**Logs**
N/A

**Expected behavior**
Corresponding ECR image and repository (if created with deployment) must be deleted on deletion of the deployment.

**Screenshots**
N/A

**Screen recording**
N/A

**Environment (please complete the following information):**
 - CLI Version: [eg. 1.0.0]  0.4.2
 - Include the output of `dotnet --info`
 - OS Info:  All

**Additional context**

---

This is a :bug: bug report

philasmar:
As discussed in the linked PR, this could potentially delete customer data and that is an unwanted behavior.

PR #211, opened 2021-05-13
feat: remove ECR repository on delete-deployment for ECS Fargate based deployments

ganeshnj:
*Issue #161

*Description of changes:*
AWS.Deploy.CLI adds aws-dotnet-deploy tag (with value recipe id as value). At the time of deletion, deployment tool checks whether the tag is present or not and deletes only when tag is present.

### DIFF
--- a/src/AWS.Deploy.CLI/Commands/DeleteDeploymentCommand.cs
+++ b/src/AWS.Deploy.CLI/Commands/DeleteDeploymentCommand.cs
@@ -2,13 +2,17 @@
 // SPDX-License-Identifier: Apache-2.0
 
 using System;
+using System.Collections.Generic;
 using System.Linq;
 using System.Threading.Tasks;
 using Amazon.CloudFormation;
 using Amazon.CloudFormation.Model;
+using Amazon.ECR.Model;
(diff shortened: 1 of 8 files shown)


src/AWS.Deploy.CLI/Commands/DeleteDeploymentCommand.cs
```diff
@@ -65,6 +74,18 @@ public async Task ExecuteAsync(string stackName)
                 // Monitor updates the stdout with current status of the CloudFormation stack
                 var _ = monitor.StartAsync();
 
+                // delete the ECR repository if exists for a container based deployment
+                // repository must be tagged with deploy tool tag
+                var repository = await GetECRRepository(stackName.ToLower());
```
normj, line 79: This is potentially deleting customer data. We don't know if they have reused this repository for other purposes or even deployed the image we created for another deployment.

This should be an opt-in parameter to delete the repository to be safe from unintentionally deleting customer data. If they don't opt in we could put a warning messages to let them know this resource still exists.

normj (changes requested):

PR #214, opened 2021-05-17
feat: Add ability to pass encryption key info through stdin from frontend to server-mode

normj:
*Description of changes:*
When server mode is started up the frontend can choose to pass in the parameter `--encryption-keyinfo-stdin`. If it is set then the cli will read from stdin for base64 encoded json document that contains encryption key info that server mode will use to decrypted any secret information. 

The json key document is versioned with currently only 1.0 version supported. This allows for server mode to safely add new styles of passing key information without breaking existing frontends.


src/AWS.Deploy.CLI/Commands/ServerModeCommand.cs
```diff
@@ -16,26 +19,76 @@ public class ServerModeCommand
...
+                        }
+                        if (keyInfo.IV != null)
+                        {
+                            aes.IV = Convert.FromBase64String(keyInfo.IV);
+                        }
+
+                        encryptionProvider = new AesEncryptionProvider(aes);
+                        break;
+                    case null:
+                        throw new InvalidEncryptionKeyInfoException("Missing require \"Version\" property in encryption key info");
```
philasmar, line 78: nit: change "require" to "required"

philasmar (approved):

src/AWS.Deploy.CLI/Commands/ServerModeCommand.cs
normj: Fixed

src/AWS.Deploy.CLI/Commands/ServerModeCommand.cs
```diff
@@ -16,26 +19,76 @@ public class ServerModeCommand
...
+                    default:
+                        throw new InvalidEncryptionKeyInfoException($"Unsupported encryption key info {keyInfo.Version}");
+                }
+
```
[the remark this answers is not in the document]
normj: Done

test/AWS.Deploy.CLI.IntegrationTests/ServerModeTests.cs
```diff
@@ -89,6 +93,61 @@ public async Task GetRecommendations()
...
+
+            using var httpClient = ServerModeHttpClientFactory.ConstructHttpClient(ResolveCredentials, aes);
+
+            InMemoryInteractiveService interactiveService = new InMemoryInteractiveService();
+            var keyInfo = new EncryptionKeyInfo
+            {
+                Version = EncryptionKeyInfo.VERSION_1_0,
+                Key = Convert.ToBase64String(aes.Key),
+                IV = Convert.ToBase64String(aes.IV)
+            };
```
ganeshnj, line 114: Key and IV is correctly initialized here on creating the instance of `Aes`. However in case of CLI, we are reading from stdin and they can be null.

POCO also have Key and IV nullable
```chsarp
    public class EncryptionKeyInfo
    {
        public const string VERSION_1_0 = "1.0";

        /// <summary>
        /// The version of the key info. If the property is set to a value that server mode has not implemented to support then
        /// a fatal exception is thrown.
        /// </summary>
        public string? Version { get; set; }

        /// <summary>
        /// Encryption key base 64 encoded 
        /// </summary>
        public string? Key { get; set; }

        /// <summary>
        /// Encryption IV base 64 encoded
        /// </summary>
        public string? IV { get; set; }

        public static EncryptionKeyInfo ParseStdInKeyInfo(string input)
        {
            var json = Encoding.UTF8.GetString(Convert.FromBase64String(input));
            var keyInfo = Newtonsoft.Json.JsonConvert.DeserializeObject<EncryptionKeyInfo>(json);

            if(string.IsNullOrEmpty(keyInfo.Key))
            {
                throw newwInvalidEncryptionKeyInfoException("Encryption key info is missing \"Key\" property.");
            }

            return keyInfo;
        }
    }
```
we should have some mechanism of saying these (IV and Key) are minimum properties to be passed in order it to work.

src/AWS.Deploy.CLI/Commands/ServerModeCommand.cs
```diff
@@ -16,26 +19,75 @@ public class ServerModeCommand
...
+         }
+
+        private IEncryptionProvider CreateEncryptionProvider()
+        {
+            IEncryptionProvider encryptionProvider;
+            if (_encryptionKeyInfoStdIn)
+            {
+                _interactiveService.WriteLine("Waiting on encryption key info from stdin");
+                var input = _interactiveService.ReadLine();
+                var keyInfo = EncryptionKeyInfo.ParseStdInKeyInfo(input);
```
ganeshnj, line 59: Should we make `EncryptionKeyInfo` more like ComplexInput type where we can ask customer to individual fields then asking for a complex input?

test/AWS.Deploy.CLI.UnitTests/Utilities/InMemoryInteractiveService.cs
```diff
@@ -0,0 +1,160 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+using System;
+using System.Diagnostics;
+using System.IO;
+
+namespace AWS.Deploy.CLI.UnitTests.Utilities
```
ganeshnj, line 8: can we move it to some shared project may be `AWS.Deploy.Tests.Common`?

ganeshnj (changes requested):

src/AWS.Deploy.CLI/Commands/ServerModeCommand.cs
normj: No, the customer is the IDE and they are not reading stdout to know what properties to set. Having a single input blob makes it easier for the IDE to know it was when stdin write to do.

test/AWS.Deploy.CLI.UnitTests/Utilities/InMemoryInteractiveService.cs
normj: Done

ganeshnj (approved):

PR #218, opened 2021-05-19
feat: monitor parent process and terminate if parent process is no longer running

philasmar:
*Issue #, if available:*
DOTNET-5083

*Description of changes:*
feat: monitor parent process and terminate if parent process is no longer running

### DIFF
--- a/src/AWS.Deploy.CLI/Commands/ServerModeCommand.cs
+++ b/src/AWS.Deploy.CLI/Commands/ServerModeCommand.cs
@@ -2,9 +2,8 @@
 // SPDX-License-Identifier: Apache-2.0
 
 using System;
-using System.Collections.Generic;
 using System.Security.Cryptography;
-using System.Text;
+using System.Diagnostics;
 using System.Threading;
 using System.Threading.Tasks;
 using AWS.Deploy.CLI.ServerMode;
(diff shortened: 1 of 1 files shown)


src/AWS.Deploy.CLI/Commands/ServerModeCommand.cs
```diff
@@ -35,7 +33,58 @@ public async Task ExecuteAsync(CancellationToken cancellationToken = default(Can
...
+                var webTask = host.RunAsync(cancellationToken);
+
+                //This call will wait until one of the tasks completes.
+                //The monitor task will complete if a parent pid is not found.
+                await Task.WhenAny(monitorTask, webTask);
+
+                //The monitor task is expected to complete only when a parent pid is not found.
+                if (monitorTask.IsCompleted && monitorTask.Result?.ProcessNotFound == true)
+                {
+                    throw new ParentProcessNotFoundException("The parent process is no longer running.");
```
normj, line 52: I don't think we have to throw an exception causing a non-zero exit code. I think we should just log that the parent process exited so we are shutting down server mode.

normj (changes requested):

src/AWS.Deploy.CLI/Commands/ServerModeCommand.cs
philasmar: Fixed.

normj (approved):

src/AWS.Deploy.CLI/Commands/ServerModeCommand.cs
```diff
@@ -91,5 +120,29 @@ private IEncryptionProvider CreateEncryptionProvider()
...
+            while (true)
+            {
+                try
+                {
+                    Process.GetProcessById((int)_parentPid);
+                    await Task.Delay(1000, token);
+                }
+                catch
+                {
+                    return new MonitoringExited { ProcessNotFound = true };
```
ganeshnj, line 143: Is there a specific exception that you can catch?

philasmar: There is, but the method throws multiple exceptions. I want to consider any issue retrieving the parent process as the same issue for our context. If anything goes wrong, then the parent process is not available.

ganeshnj: Then I think we don't need

```
        public class MonitoringExited
        {
            public bool ProcessNotFound {get;set;}
        }
```
Setting ProcessNotFound to true gives a false indication. We should simply return a boolean that indicates the whether GetProcessById threw an exception or not.

If needed we can always expand return type but I don't reason for MonitoringExited existence.

philasmar: I removed MonitoringExited and returned a bool.

PR #219, opened 2021-05-20
feat: server mode exits with code -100 if port is in use

philasmar:
*Issue #, if available:*
DOTNET-5084

*Description of changes:*
feat: server mode exits with code -100 if port is in use


normj (changes requested):
Code looks fine but you should be able to create a test case launching 2 instances of the ServerModeCommand with the same port.

philasmar:
> Code looks fine but you should be able to create a test case launching 2 instances of the ServerModeCommand with the same port.

I added a test case.

normj (approved):

ganeshnj (approved):

PR #221, opened 2021-05-21
feat: implement `ServerModeSession` utility class

ganeshnj:
*Issue #, if available:*

*Description of changes:*
ServerModeSession exposes Start method which starts the deployment tools server on a random registered port number for the given process id.

It also offers resiliency for port unavailable error by allow a configurable number of retries.

https://user-images.githubusercontent.com/8882380/120391508-f47b0600-c2e3-11eb-82eb-2e01507927bc.mp4

### DIFF
--- a/src/AWS.Deploy.ServerMode.Client/ServerModeSession.cs
+++ b/src/AWS.Deploy.ServerMode.Client/ServerModeSession.cs
@@ -1,0 +1,243 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.\r
+// SPDX-License-Identifier: Apache-2.0
+
+using System;
+using System.Diagnostics;
+using System.IO;
+using System.Net.Http;
+using System.Runtime.InteropServices;
+using System.Security.Cryptography;
+using System.Text;
+using System.Threading;
+using System.Threading.Tasks;
+using Amazon.Runtime;
+using Newtonsoft.Json;
+
+namespace AWS.Deploy.ServerMode.Client
+{
+    /// <summary>
+    /// Helper class that allows launching deployment tool in server mode.
+    /// It abstracts the server mode setup, CLI command execution and retries when desired port is unavailable.
+    /// </summary>
+    public interface IServerModeSession
+    {
+        /// <summary>
+        /// Starts deployment tool in server mode.
+        /// It creates symmetric key and tries to setup the server mode.
+        /// It also handles the retries when a desired port is unavailable in the provided range.
+        /// </summary>
+        /// <param name="cancellationToken">A cancellation token that can be used by other objects or threads to receive notice of cancellation.</param>
+        /// <exception cref="InternalServerModeException">Throws when deployment tool server failed to start for un unknown reason.</exception>
+        /// <exception cref="PortUnavailableException">Throws when deployment tool server failed to start due to unavailability of free ports.</exception>
+        Task Start(CancellationToken cancellationToken);
+
+        /// <summary>
+        /// Builds <see cref="IRestAPIClient"/> client using the cached base URL.
+        /// If succeeded, <param name="restApiClient"></param> is initialized with current session client.
+        /// </summary>
+        /// <param name="credentialsGenerator">Func to that provides AWS credentials</param>
+        /// <param name="restApiClient"><see cref="IRestAPIClient"/> client to initialize.</param>
+        /// <returns>True, if <param name="restApiClient"/> is initialized successfully.</returns>
+        bool TryGetRestAPIClient(Func<Task<AWSCredentials>> credentialsGenerator, out IRestAPIClient? restApiClient);
+
+        /// <summary>
+        /// Builds <see cref="IDeploymentCommunicationClient"/> client using the cached base URL.
+        /// If succeeded, <param name="deploymentCommunicationClient"></param> is initialized with current session client.
+        /// </summary>
+        /// <param name="deploymentCommunicationClient"><see cref="DeploymentCommunicationClient"/> client to initialize.</param>
+        /// <returns>True, if <param name="deploymentCommunicationClient"/> is initialized successfully.</returns>
+        bool TryGetDeploymentCommunicationClient(out IDeploymentCommunicationClient? deploymentCommunicationClient);
+
+        /// <summary>
+        /// Returns the status of the deployment server by checking /api/v1/health API.
+        /// Returns true, if the deployment server returns a success HTTP code.
+        /// </summary>
+        /// <param name="cancellationToken">A cancellation token that can be used by other objects or threads to receive notice of cancellation.</param>
+        /// <returns>Returns true, if the deployment server returns a success HTTP code.</returns>
+        Task<bool> IsAlive(CancellationToken cancellationToken);
+    }
+
+    public class ServerModeSession : IServerModeSession, IDisposable
+    {
+        private const int TCP_PORT_ERROR = -100;
+
+        private readonly int _startPort;
+        private readonly int _endPort;
+        private readonly CommandLineWrapper _commandLineWrapper;
+        private readonly HttpClientHandler _httpClientHandler;
+        private readonly TimeSpan _serverTimeout;
+
+        private string? _baseUrl;
+        private Aes? _aes;
+
+        private string HealthUrl
+        {
+            get
+            {
+                if (_baseUrl == null)
+                {
+                    throw new InvalidOperationException($"{nameof(_baseUrl)} must not be null.");
+                }
+
+                return $"{_baseUrl}/api/v1/health";
+            }
+        }
+
+        public ServerModeSession(int startPort = 10000, int endPort = 10100, bool diagnosticLoggingEnabled = false)
+            : this(new CommandLineWrapper(diagnosticLoggingEnabled),
+                new HttpClientHandler(),
+                TimeSpan.FromSeconds(60),
+                startPort,
+                endPort)
+        {
+        }
+
+        public ServerModeSession(CommandLineWrapper commandLineWrapper,
+            HttpClientHandler httpClientHandler,
+            TimeSpan serverTimeout,
+            int startPort = 10000,
+            int endPort = 10100)
+        {
+            _startPort = startPort;
+            _endPort = endPort;
+            _commandLineWrapper = commandLineWrapper;
+            _httpClientHandler = httpClientHandler;
+            _serverTimeout = serverTimeout;
+        }
+
+        public async Task Start(CancellationToken cancellationToken)
+        {
+            var currentProcessId = Process.GetCurrentProcess().Id;
+
+            for (var port = _startPort; port <= _endPort; port++)
+            {
+                _aes = Aes.Create();
+                _aes.GenerateKey();
+                _aes.GenerateIV();
+
+                var keyInfo = new EncryptionKeyInfo(
+                    EncryptionKeyInfo.VERSION_1_0,
+                    Convert.ToBase64String(_aes.Key),
+                    Convert.ToBase64String(_aes.IV));
+
+                var keyInfoStdin = Convert.ToBase64String(Encoding.UTF8.GetBytes(JsonConvert.SerializeObject(keyInfo)));
+
+                var command = $"dotnet aws server-mode --port {port} --parent-pid {currentProcessId} --encryption-keyinfo-stdin";
+                var startServerTask = _commandLineWrapper.Run(command, keyInfoStdin);
+
+                _baseUrl = $"http://localhost:{port}";
+                var isServerAvailableTask = IsServerAvailable(cancellationToken);
+
+                if (isServerAvailableTask == await Task.WhenAny(startServerTask, isServerAvailableTask).ConfigureAwait(false))
+                {
+                    // The server timed out, this isn't a transient error, therefore, we throw
+                    if (!isServerAvailableTask.Result)
+                    {
+                        throw new InternalServerModeException($"\"{command}\" failed for unknown reason.");
+                    }
+
+                    // Server has started, it is safe to return
+                    return;
+                }
+
+                // For -100 errors, we want to check all the ports in the configured port range
+                // If the error code other than -100, this is an unexpected exit code.
+                if (startServerTask.Result != TCP_PORT_ERROR)
+                {
+                    throw new InternalServerModeException($"\"{command}\" failed for unknown reason.");
+                }
+            }
+
+            throw new PortUnavailableException($"Free port unavailable in {_startPort}-{_endPort} range.");
+        }
+
+        public bool TryGetRestAPIClient(Func<Task<AWSCredentials>> credentialsGenerator, out IRestAPIClient? restApiClient)
+        {
+            if (_baseUrl == null || _aes == null)
+            {
+                restApiClient = null;
+                return false;
+            }
+
+            var httpClient = ServerModeHttpClientFactory.ConstructHttpClient(credentialsGenerator, _aes);
+            restApiClient = new RestAPIClient(_baseUrl, httpClient);
+            return true;
+        }
+
+        public bool TryGetDeploymentCommunicationClient(out IDeploymentCommunicationClient? deploymentCommunicationClient)
+        {
+            if (_baseUrl == null || _aes == null)
+            {
+                deploymentCommunicationClient = null;
+                return false;
+            }
+
+            deploymentCommunicationClient = new DeploymentCommunicationClient(_baseUrl);
+            return true;
+        }
+
+        public async Task<bool> IsAlive(CancellationToken cancellationToken)
+        {
+            var client = new HttpClient(_httpClientHandler);
+
+            try
+            {
+                var response = await client.GetAsync(HealthUrl, cancellationToken);
+                return response.IsSuccessStatusCode;
+            }
+            catch (Exception)
+            {
+                return false;
+            }
+        }
+
+        #region Private methods
+
+        private Task<bool> IsServerAvailable(CancellationToken cancellationToken)
+        {
+            return WaitUntilHelper.WaitUntilSuccessStatusCode(
+                HealthUrl,
+                _httpClientHandler,
+                TimeSpan.FromMilliseconds(100),
+                _serverTimeout,
+                cancellationToken);
+        }
+
+        #endregion
+
+        #region Disposable
+
+        protected virtual void Dispose(bool disposing)
+        {
+            if (disposing)
+            {
+                _aes?.Dispose();
+                _aes = null;
+            }
+        }
+
+        public void Dispose()
+        {
+            Dispose(true);
+            GC.SuppressFinalize(this);
+        }
+
+        #endregion
+
+        private class EncryptionKeyInfo
+        {
+            public const string VERSION_1_0 = "1.0";
+
+            public string Version { get; set; }
+            public string Key { get; set; }
+            public string IV { get; set; }
+
+            public EncryptionKeyInfo(string version, string key, string iv)
+            {
+                Version = version;
+                Key = key;
+                IV = iv;
+            }
+        }
+    }
+}
--- a/AWS.Deploy.sln
+++ b/AWS.Deploy.sln
@@ -64,6 +64,8 @@ Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "AWS.Deploy.ServerMode.Clien
 EndProject
 Project("{D954291E-2A0B-460D-934E-DC6B0785DB48}") = "AWS.Deploy.Constants", "src\AWS.Deploy.Constants\AWS.Deploy.Constants.shproj", "{F2266C44-C8C5-45AD-AA9B-44F8825BDF63}"
 EndProject
+Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "AWS.Deploy.ServerMode.Client.UnitTests", "test\AWS.Deploy.ServerMode.Client.UnitTests\AWS.Deploy.ServerMode.Client.UnitTests.csproj", "{74444ED0-9044-4DF7-A111-7D9F81ADF0FD}"
+EndProject
 Global
 	GlobalSection(SharedMSBuildProjectFiles) = preSolution
 		src\AWS.Deploy.Constants\AWS.Deploy.Constants.projitems*{f2266c44-c8c5-45ad-aa9b-44f8825bdf63}*SharedItemsImports = 13
@@ -158,6 +160,10 @@ Global
 		{C533AA26-797A-4A41-BA28-2364B66DAA5F}.Debug|Any CPU.Build.0 = Debug|Any CPU
 		{C533AA26-797A-4A41-BA28-2364B66DAA5F}.Release|Any CPU.ActiveCfg = Release|Any CPU
 		{C533AA26-797A-4A41-BA28-2364B66DAA5F}.Release|Any CPU.Build.0 = Release|Any CPU
+		{74444ED0-9044-4DF7-A111-7D9F81ADF0FD}.Debug|Any CPU.ActiveCfg = Debug|Any CPU
+		{74444ED0-9044-4DF7-A111-7D9F81ADF0FD}.Debug|Any CPU.Build.0 = Debug|Any CPU
+		{74444ED0-9044-4DF7-A111-7D9F81ADF0FD}.Release|Any CPU.ActiveCfg = Release|Any CPU
+		{74444ED0-9044-4DF7-A111-7D9F81ADF0FD}.Release|Any CPU.Build.0 = Release|Any CPU
 	EndGlobalSection
 	GlobalSection(SolutionProperties) = preSolution
 		HideSolutionNode = FALSE
@@ -186,6 +192,7 @@ Global
 		{1B171F48-D200-407D-A9F2-179E2A602821} = {11C7056E-93C1-408B-BD87-5270595BBE0E}
 		{C533AA26-797A-4A41-BA28-2364B66DAA5F} = {11C7056E-93C1-408B-BD87-5270595BBE0E}
 		{F2266C44-C8C5-45AD-AA9B-44F8825BDF63} = {11C7056E-93C1-408B-BD87-5270595BBE0E}
+		{74444ED0-9044-4DF7-A111-7D9F81ADF0FD} = {BD466B5C-D8B0-4069-98A9-6DC8F01FA757}
 	EndGlobalSection
 	GlobalSection(ExtensibilityGlobals) = postSolution
 		SolutionGuid = {5A4B2863-1763-4496-B122-651A38A4F5D7}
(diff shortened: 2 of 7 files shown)


src/AWS.Deploy.ServerMode.Client/ServerModeSession.cs
```diff
@@ -0,0 +1,122 @@
...
+    /// Helper class that allows launching deployment tool in server mode.
+    /// </summary>
+    public class ServerModeSession
+    {
+        private const int TCP_PORT_ERROR = -100;
+
+        private readonly int _startPort;
+        private readonly int _endPort;
+
+        public ServerModeSession(int startPort = 5000, int endPort = 5100)
```
normj, line 23: Since new .NET web projects default to 5000 I think we should stay well clear from there. What about 10000 to 10100?

src/AWS.Deploy.ServerMode.Client/ServerModeSession.cs
```diff
@@ -0,0 +1,122 @@
...
+        /// <exception cref="PortUnavailableException">Throws when deployment tool server failed to start due to unavailability of free ports.</exception>
+        public async Task<string> Start()
+        {
+            var currentProcessId = Process.GetCurrentProcess().Id;
+
+            for (var port = _startPort; port <= _endPort; port++)
+            {
+                var baseUrl = $"http://localhost:{port}";
+
+                var command = $"dotnet aws server-mode --port {port} --parent-pid {currentProcessId}";
```
normj, line 43: We need to set the `--encryption-keyinfo-stdin` switch and then generate the encryption key info and pass that into stdin. This test might help you see what needs to be done. https://github.com/aws/aws-dotnet-deploy/blob/main/test/AWS.Deploy.CLI.IntegrationTests/ServerModeTests.cs#L108

We will also need to make sure we help create the RestClient with the key information as well.

src/AWS.Deploy.ServerMode.Client/ServerModeSession.cs
```diff
@@ -0,0 +1,122 @@
...
+
+        private async Task<int> RunCommand(string command)
+        {
+            var arguments = RuntimeInformation.IsOSPlatform(OSPlatform.Windows) ? $"/c {command}" : $"-c \"{command}\"";
+
+            var processStartInfo = new ProcessStartInfo
+            {
+                FileName = GetSystemShell(),
+                Arguments = arguments,
+                UseShellExecute = false, // we don't want to start terminal
```
normj, line 80: We will want this configurable. So if we are in a diagnostic mode we can see the shell window.

src/AWS.Deploy.ServerMode.Client/Exceptions/InternalServerModeException.cs
```diff
@@ -0,0 +1,17 @@
...
+using System;
+
+namespace AWS.Deploy.ServerMode.Client.Exceptions
+{
+    /// <summary>
+    /// Exception is thrown when deployment tool server failed to start for un unknown reason.
+    /// </summary>
+    public class InternalServerModeException : Exception
+    {
+        public InternalServerModeException(string message): base(message)
```
philasmar, line 13: Can you use the same overloads like all our other exceptions? (message and inner exception)
Also why are we creating a new .cs file for every new exception?

src/AWS.Deploy.ServerMode.Client/Exceptions/InternalServerModeException.cs
```diff
@@ -0,0 +1,17 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.\r
+// SPDX-License-Identifier: Apache-2.0
+
+using System;
+
+namespace AWS.Deploy.ServerMode.Client.Exceptions
+{
+    /// <summary>
+    /// Exception is thrown when deployment tool server failed to start for un unknown reason.
```
philasmar, line 9: fix typo "un" to "an"

src/AWS.Deploy.ServerMode.Client/Exceptions/PortUnavailableException.cs
```diff
@@ -0,0 +1,17 @@
...
+// SPDX-License-Identifier: Apache-2.0
+
+using System;
+
+namespace AWS.Deploy.ServerMode.Client.Exceptions
+{
+    /// <summary>
+    /// Exception is thrown when deployment tool server failed to start due to unavailability of free ports.
+    /// </summary>
+    public class PortUnavailableException : Exception
```
philasmar, line 11: Can you use the same overloads like all our other exceptions? (message and inner exception)
Also why are we creating a new .cs file for every new exception?

src/AWS.Deploy.ServerMode.Client/ServerModeSession.cs
```diff
@@ -0,0 +1,122 @@
...
+using AWS.Deploy.ServerMode.Client.Exceptions;
+
+namespace AWS.Deploy.ServerMode.Client
+{
+    /// <summary>
+    /// Helper class that allows launching deployment tool in server mode.
+    /// </summary>
+    public class ServerModeSession
+    {
+        private const int TCP_PORT_ERROR = -100;
```
philasmar, line 18: We are also declaring this constant here https://github.com/aws/aws-dotnet-deploy/blob/main/src/AWS.Deploy.CLI/CommandReturnCodes.cs#L36
Can we unify this an reference 1 place?

src/AWS.Deploy.ServerMode.Client/ServerModeSession.cs
```diff
@@ -0,0 +1,122 @@
...
+                UseShellExecute = false, // we don't want to start terminal
+            };
+
+            var process = Process.Start(processStartInfo);
+            if (process == null)
+            {
+                throw new InvalidOperationException();
+            }
+
+            while (true)
```
normj, line 89: What is this loop doing that just calling process `WaitForExit` doesn't do.

src/AWS.Deploy.ServerMode.Client/ServerModeSession.cs
```diff
@@ -0,0 +1,122 @@
...
+using System.Diagnostics;
+using System.Net.Http;
+using System.Runtime.InteropServices;
+using System.Threading.Tasks;
+using AWS.Deploy.ServerMode.Client.Exceptions;
+
+namespace AWS.Deploy.ServerMode.Client
+{
+    /// <summary>
+    /// Helper class that allows launching deployment tool in server mode.
```
philasmar, line 14: Can you add more details on what this is doing? There's no mention on why we need the start port and end port for constructing this class.

src/AWS.Deploy.ServerMode.Client/Exceptions/PortUnavailableException.cs
```diff
@@ -0,0 +1,17 @@
...
+    {
+        public PortUnavailableException(string message): base(message)
```
philasmar, line 13: A similar exception is available here https://github.com/aws/aws-dotnet-deploy/blob/main/src/AWS.Deploy.CLI/Exceptions.cs#L69 and created for the same reason. We should reuse instead of replicating these things.

src/AWS.Deploy.ServerMode.Client/ServerModeSession.cs
```diff
@@ -0,0 +1,122 @@
...
+            _startPort = startPort;
+            _endPort = endPort;
+        }
+
+        /// <summary>
+        /// Start deployment tool in server mode.
+        /// </summary>
+        /// <returns>Base URL of the  deployment tool local server.</returns>
+        /// <exception cref="InternalServerModeException">Throws when deployment tool server failed to start for un unknown reason.</exception>
+        /// <exception cref="PortUnavailableException">Throws when deployment tool server failed to start due to unavailability of free ports.</exception>
```
philasmar, line 34: Same comment about reusing this exception

src/AWS.Deploy.ServerMode.Client/ServerModeSession.cs
```diff
@@ -0,0 +1,122 @@
...
+            for (var port = _startPort; port <= _endPort; port++)
+            {
+                var baseUrl = $"http://localhost:{port}";
+
+                var command = $"dotnet aws server-mode --port {port} --parent-pid {currentProcessId}";
+                var startServerTask = RunCommand(command);
+
+                var isServerAvailableTask = IsServerAvailable(baseUrl);
+
+                // If the server is server health check succeeded
```
philasmar, line 48: nit: sentence structure is confusing. "If the server is server"

src/AWS.Deploy.ServerMode.Client/ServerModeSession.cs
```diff
@@ -0,0 +1,122 @@
...
+                // return the baseUrl
+                if (isServerAvailableTask == await Task.WhenAny(startServerTask, isServerAvailableTask))
```
normj, line 50: You aren't checking the boolean return by IsServerAvailable to see if the task didn't just finished because of being timed out.

src/AWS.Deploy.ServerMode.Client/ServerModeSession.cs
```diff
@@ -0,0 +1,122 @@
...
+                {
+                    return baseUrl;
+                }
+
+                // For -100 errors, we want to check all the ports in the configured port range
+                // If the error code other than -100, this is an unexpected exit code.
+                if (startServerTask.Result != TCP_PORT_ERROR)
+                {
+                    throw new InternalServerModeException($"\"{command}\" failed for unknown reason.");
```
philasmar, line 59: nit: I'd rather we not specify the command explicitly, just provide a more descriptive message.

normj (approved):

philasmar (changes requested):

normj (changes requested):
Sorry I meant to say "Request Changes" before.

src/AWS.Deploy.ServerMode.Client/ServerModeSession.cs
ganeshnj: that works, somewhat lesser chances of collision.

awschristou: The Toolkit will manage its own installation and location of the cli. Can an optional tool path argument be added so that dotnet calls can be made with the  `--toolpath` arg?

src/AWS.Deploy.ServerMode.Client/ServerModeSession.cs
```diff
@@ -0,0 +1,122 @@
...
+            _endPort = endPort;
+        }
+
+        /// <summary>
+        /// Start deployment tool in server mode.
+        /// </summary>
+        /// <returns>Base URL of the  deployment tool local server.</returns>
+        /// <exception cref="InternalServerModeException">Throws when deployment tool server failed to start for un unknown reason.</exception>
+        /// <exception cref="PortUnavailableException">Throws when deployment tool server failed to start due to unavailability of free ports.</exception>
+        public async Task<string> Start()
```
awschristou, line 35: Please provide an overload that uses a CancellationToken, so that calling clients can back out of the request if needed. (the non-overloaded version would use `CancellationToken.None`)

src/AWS.Deploy.ServerMode.Client/ServerModeSession.cs
ganeshnj: Added a todo, we will add ability to configure settings such as port, diagnostics in a separate PR.

src/AWS.Deploy.ServerMode.Client/ServerModeSession.cs
ganeshnj: Fixed.

src/AWS.Deploy.ServerMode.Client/ServerModeSession.cs
ganeshnj: Added a flag which can be set from outside.

src/AWS.Deploy.ServerMode.Client/ServerModeSession.cs
ganeshnj: We must change in our CLIWrapper also.

src/AWS.Deploy.ServerMode.Client/ServerModeSession.cs
ganeshnj: For timeout, isServerAvailableTask  returns false. Added the nested check to see the result.

src/AWS.Deploy.ServerMode.Client/Exceptions/InternalServerModeException.cs
ganeshnj: We don't have inner exception there we don't need inner exception overload here.

I moved exceptions to Exceptions.cs like we to in CLI.

src/AWS.Deploy.ServerMode.Client/Exceptions/InternalServerModeException.cs
ganeshnj: Fixed

src/AWS.Deploy.ServerMode.Client/Exceptions/PortUnavailableException.cs
ganeshnj: We don't have inner exception there we don't need inner exception overload here.

I moved exceptions to Exceptions.cs like we to in CLI.

src/AWS.Deploy.ServerMode.Client/ServerModeSession.cs
ganeshnj: `AWS.Deploy.ServerMode.Client.csproj` doesn't (can't) reference CLI. It is an independent package which will be used by the VS Toolkit.

src/AWS.Deploy.ServerMode.Client/ServerModeSession.cs
ganeshnj: Fixed.

src/AWS.Deploy.ServerMode.Client/Exceptions/PortUnavailableException.cs
ganeshnj: `AWS.Deploy.ServerMode.Client.csproj` doesn't (can't) reference CLI. It is an independent package which will be used by the VS Toolkit.

src/AWS.Deploy.ServerMode.Client/ServerModeSession.cs
ganeshnj: `AWS.Deploy.ServerMode.Client.csproj` doesn't (can't) reference CLI. It is an independent package which will be used by the VS Toolkit.

src/AWS.Deploy.ServerMode.Client/ServerModeSession.cs
ganeshnj: Fixed.

src/AWS.Deploy.ServerMode.Client/ServerModeSession.cs
ganeshnj: This is for VS Toolkit, I don't see a reason not to specify the command. It will never be shown to the customer. Also, we dont't know in this case, what is the underlying reason, therefore, having command atleast make debugging easier.

src/AWS.Deploy.ServerMode.Client/ServerModeSession.cs
ganeshnj: Added

src/AWS.Deploy.ServerMode.Client/ServerModeSession.cs
philasmar: I am not saying that we should reference the CLI, but we should think about unifying these things maybe by using a shared project. I am sensing a pattern here and I don't want to end up having to maintain 2 places.

philasmar (approved):

src/AWS.Deploy.ServerMode.Client/RestAPIClientFactory.cs
```diff
@@ -0,0 +1,30 @@
...
+    {
+        /// <summary>
+        /// Factory method that runs the deployment tool in server mode
+        /// and creates the <see cref="RestAPIClient"/> client with the server mode start configuration.
+        /// TODO: ability to pass configuration file path
+        /// </summary>
+        /// <param name="credentialsGenerator">Func to that provides AWS credentials</param>
+        /// <param name="cancellationToken">A cancellation token that can be used by other objects or threads to receive notice of cancellation.</param>
+        /// <returns>Returns <see cref="RestAPIClient"/> client for the current session.</returns>
+        public static async Task<IRestAPIClient> BuildAPIClient(Func<Task<AWSCredentials>> credentialsGenerator, CancellationToken cancellationToken)
```
awschristou, line 21: From the Toolkit, I plan to request different clients from the running session. Each client will have its own credentials generator. What do you think about a `ServerModeStartInfo` extension method that performs the construction of the http client and the rest client, and returns the rest client? This would make a nice abstraction that lets a caller get the rest client without worrying about the details.

src/AWS.Deploy.ServerMode.Client/ServerModeSession.cs
```diff
@@ -0,0 +1,152 @@
...
+            }
+
+            foreach (var line in stdIn)
+            {
+                await process.StandardInput.WriteLineAsync(line);
+            }
+
+            process.WaitForExit(-1);
+
+            return await Task.FromResult(process.ExitCode);
```
awschristou, line 127: These two lines may need to get wrapped in a `Task.Run` call. `WaitForExit` appears to lock things up for me (the other task that checks for a successful status code does not proceed to run).

ganeshnj: is  this because you are calling the method from main thread?

awschristou: I am, however the main UI thread doesn't get locked up as things are still responsive. The Tasks that are created in ServerModeSession::Start are what appear to stop running. The call `isServerAvailableTask == await Task.WhenAny(startServerTask, isServerAvailableTask)` never completes.

src/AWS.Deploy.ServerMode.Client/RestAPIClientFactory.cs
awschristou: Sounds great. I love the interface approach as well. In addition, I'd like if it was possible to:
* query if a server is up (this will let the Toolkit know if the server stops that it should update the UI accordingly and discard its RestApiClient)
* stop a running server (this will let the Toolkit attempt to clean up running server on shut down)

awschristou: I'd like to be able to get a rest api client in a non-async manner. Would it be possible to add a version of GetRestAPIClient that doesn't try to start the server but just produces the api client?

ganeshnj: I would prefer having the code of restarting the server at one place. Either the caller, make sure server is running, if not first call Start and then get the client. Or let the Session handle this. Both ways, it works for me.

src/AWS.Deploy.ServerMode.Client/ServerModeSession.cs
```diff
@@ -0,0 +1,152 @@
...
+
+    public class ServerModeSession : IServerModeSession
+    {
+        private const int TCP_PORT_ERROR = -100;
+
+        private readonly int _startPort;
+        private readonly int _endPort;
+        private readonly bool _diagnosticLoggingEnabled;
+
+        public ServerModeSession(int startPort = 10000, int endPort = 10100, bool diagnosticLoggingEnabled = false)
```
awschristou, line 41: Recapping some earlier discussions, the Toolkit will need to explicitly specify the tool path (of the CLI), and optionally provide a path to a JSON that contains other parameters. The JSON would optionally configure the start/end port, and whether diagnostic logging is enabled.  

We want to avoid a case where the Toolkit has to generate a json at runtime, which is why the tool path is excluded from the json file.

Its okay for this change to take place in a follow up PR.

src/AWS.Deploy.ServerMode.Client/RestAPIClientFactory.cs
ganeshnj: > I'd like to be able to get a rest api client in a non-async manner.

This will be tricky, we have to take out the Start server out of the call stack then,

awschristou: Ok, lets have the caller (Toolkit) take responsibility for re/starting the server. This way it has the flexibility to handle the lifecycle as needed.

src/AWS.Deploy.ServerMode.Client/ServerModeSession.cs
ganeshnj: Can you try with latest changes pushed, I tested in an WPF app and it is working expectedly.

src/AWS.Deploy.ServerMode.Client/RestAPIClientFactory.cs
ganeshnj: Updated.

src/AWS.Deploy.ServerMode.Client/ServerModeSession.cs
```diff
@@ -0,0 +1,260 @@
...
+            {
+                return false;
+            }
+        }
+
+        #region Private methods
+
+        private Task<bool> IsServerAvailable(string baseUrl, CancellationToken cancellationToken)
+        {
+            var healthUrl = $"{baseUrl}/api/v1/health";
```
awschristou, line 168: Is there any way to avoid hardcoding and keep this in sync with the endpoint url?

src/AWS.Deploy.ServerMode.Client/ServerModeStartInfo.cs
```diff
@@ -0,0 +1,19 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.\r
+// SPDX-License-Identifier: Apache-2.0
+
+using System.Security.Cryptography;
+
+namespace AWS.Deploy.ServerMode.Client
+{
+    public class ServerModeStartInfo
```
awschristou, line 8: This class does not appear to be used anymore

src/AWS.Deploy.ServerMode.Client/ServerModeSession.cs
```diff
@@ -0,0 +1,277 @@
...
+            {
+                return false;
+            }
+        }
+
+        #region Private methods
+
+        private Task<bool> IsServerAvailable(string baseUrl, CancellationToken cancellationToken)
+        {
+            return WaitUntilHelper.WaitUntilSuccessStatusCode(HealthUrl, TimeSpan.FromMilliseconds(100), TimeSpan.FromMinutes(10), cancellationToken);
```
awschristou, line 186: baseUrl is unused

src/AWS.Deploy.ServerMode.Client/ServerModeSession.cs
```diff
@@ -0,0 +1,277 @@
...
+        private string HealthUrl
+        {
+            get
+            {
+                if (_baseUrl == null)
+                {
+                    throw new InvalidOperationException($"{nameof(_baseUrl)} must not be null.");
+                }
+
+                return $"{_baseUrl}/api/v1/health";
```
awschristou, line 80: What prevents this url from getting out of sync with what is defined in the rest api?

src/AWS.Deploy.ServerMode.Client/ServerModeSession.cs
```diff
@@ -0,0 +1,277 @@
...
+            var processStartInfo = new ProcessStartInfo
+            {
+                FileName = GetSystemShell(),
+                Arguments = arguments,
+                UseShellExecute = false, // UseShellExecute must be false in allow redirection of StdIn.
+                RedirectStandardInput = true,
+                CreateNoWindow = !_diagnosticLoggingEnabled, // It allows displaying stdout and stderr on the screen.
+            };
+
+            var process = Process.Start(processStartInfo);
```
awschristou, line 202: The Toolkit will need `IServerModeSession` to have a "Stop" function, allowing it to terminate the server/process. The CLI makes an attempt to shut down by watching the parent process, but this should be a fallback mechanism. We are already seeing a negative impact from this during Toolkit development.

What would be the downside to storing `process` (or even just the processId), so that a `Stop` function could attempt to kill that process?

ganeshnj: I would keep this feature out of this PR. There are multiple ways to solve this problem
- Abort the process using process id
- Create an API endpoint which allows aborting the server

Let's first get this PR and then we can implement it later.

src/AWS.Deploy.ServerMode.Client/ServerModeSession.cs
ganeshnj: Rest api is generated code and we don't have any way of sharing the code there AFIK. We will need to keep it up to date as we bump the versions of our APIs.

src/AWS.Deploy.ServerMode.Client/ServerModeSession.cs
ganeshnj: fixed.

src/AWS.Deploy.ServerMode.Client/ServerModeStartInfo.cs
ganeshnj: removed, it is not nested in ServerModeSession itself.

src/AWS.Deploy.ServerMode.Client/ServerModeSession.cs
awschristou: Agreed. I've stubbed around this in the Toolkit for now in anticipation of something here in the near future.

awschristou (approved):

awschristou (approved):

src/AWS.Deploy.ServerMode.Client/ServerModeSession.cs
```diff
@@ -0,0 +1,243 @@
...
+        /// <returns>True, if <param name="deploymentCommunicationClient"/> is initialized successfully.</returns>
+        bool TryGetDeploymentCommunicationClient(out IDeploymentCommunicationClient? deploymentCommunicationClient);
+
+        /// <summary>
+        /// Returns the status of the deployment server by checking /api/v1/health API.
+        /// Returns true, if the deployment server returns a success HTTP code.
+        /// </summary>
+        /// <param name="cancellationToken">A cancellation token that can be used by other objects or threads to receive notice of cancellation.</param>
+        /// <returns>Returns true, if the deployment server returns a success HTTP code.</returns>
+        ValueTask<bool> IsAlive(CancellationToken cancellationToken);
```
normj, line 57: ValueTask will be hard for Visual Studio which is .NET Framework 4.6.1 to use. Switch to regular Task.

normj (changes requested):

src/AWS.Deploy.ServerMode.Client/ServerModeSession.cs
ganeshnj: Fixed.

normj (approved):

PR #226, opened 2021-05-25
feat: Generate interfaces for server mode clients

normj:
*Description of changes:*
Create interfaces for server mode clients to allow VS to be able to mock the clients.

The RestAPI interface is generated but setting the `GenerateClientInterfaces` property to true in the generator.

The SignalR DeploymentCommunicationClient interface is created by hand extracting the public methods from DeploymentCommunicationClient.

### DIFF
--- a/src/AWS.Deploy.ServerMode.Client/DeploymentCommunicationClient.cs
+++ b/src/AWS.Deploy.ServerMode.Client/DeploymentCommunicationClient.cs
@@ -10,7 +10,20 @@ using Microsoft.Extensions.Logging;
 
 namespace AWS.Deploy.ServerMode.Client
 {
-    public class DeploymentCommunicationClient : IDisposable
+    public interface IDeploymentCommunicationClient : IDisposable
+    {
+        Action<string>? ReceiveLogDebugLine { get; set; }
+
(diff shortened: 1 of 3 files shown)


ganeshnj (approved):

src/AWS.Deploy.ServerMode.Client/DeploymentCommunicationClient.cs
```diff
@@ -10,7 +10,20 @@
...
+        Action<string>? ReceiveLogErrorMessageLine { get; set; }
+
+        Action<string>? ReceiveLogMessageLineAction { get; set; }
+
+        Action<string>? ReceiveLogAllLogAction { get; set; }
+
+        Task JoinSession(string sessionId);
+    }
+
+    public class DeploymentCommunicationClient : IDeploymentCommunicationClient, IDisposable
```
philasmar, line 26: Why do we need to specify the IDisposable interface on both IDeploymentCommunicationClient and DeploymentCommunicationClient?

normj: Removed IDisposable from DeploymentCommunicationClient.

philasmar (approved):

PR #229, opened 2021-05-27
feat: Add ability to redeploy without any prompts for user input

96malhar:
*Issue #, if available:*
DOTNET-4887

*Description of changes:*
This PR adds a `disable-interactive` flag which can be passed as a CLI argument. This flag allows a redeployment to be performed without any user prompts

### DIFF
--- a/src/AWS.Deploy.CLI/Commands/DeployCommand.cs
+++ b/src/AWS.Deploy.CLI/Commands/DeployCommand.cs
@@ -175,7 +175,8 @@ namespace AWS.Deploy.CLI.Commands
 
             var configurableOptionSettings = selectedRecommendation.Recipe.OptionSettings.Union(deploymentBundleDefinition.Parameters);
 
-            await ConfigureDeployment(selectedRecommendation, configurableOptionSettings, false);
+            if (!_toolInteractiveService.DisableInteractive)
+                await ConfigureDeployment(selectedRecommendation, configurableOptionSettings, false);
 
             var cloudApplication = new CloudApplication(cloudApplicationName, string.Empty);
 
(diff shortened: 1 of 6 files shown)


src/AWS.Deploy.CLI/Commands/DeployCommand.cs
```diff
@@ -102,6 +102,10 @@ public async Task ExecuteAsync(string stackName, bool saveCdkProject)
 
             var deployedApplication = deployedApplications.FirstOrDefault(x => string.Equals(x.Name, cloudApplicationName));
 
+            // Throw an error if we are NOT performing a redeployment and the DisableInteractive flag is set to true.
+            if (_toolInteractiveService.DisableInteractive && deployedApplication == null)
+                throw new InvalidCliArgumentException("The \"--disable-interactive\" CLI argument can only be used during a redeployment. Please remove this argument and try again.");
```
philasmar, line 107: We don't need this. We shouldn't blow up if the user is doing a deployment.

philasmar (changes requested):

src/AWS.Deploy.CLI/Commands/DeployCommand.cs
96malhar: As per my discussion with @normj, this flag should only be used if the user is trying to perform a redeployment. If a user is performing a new deployment with this flag, then an error must be thrown

normj: When this work is merged with Phil work to pass in a user settings file then the flag can be used for initial deployment as well as redeployments.

96malhar: Removed this line so that a user can perform an initial deployment even when this flag is set.

PR #232, opened 2021-06-01
feat: use local config file when deploying

philasmar:
*Issue #, if available:*
DOTNET-4883

*Description of changes:*
feat- use local config file when deploying

### DIFF
--- a/src/AWS.Deploy.CLI/Commands/DeployCommand.cs
+++ b/src/AWS.Deploy.CLI/Commands/DeployCommand.cs
@@ -66,7 +66,7 @@ namespace AWS.Deploy.CLI.Commands
             _cdkManager = cdkManager;
         }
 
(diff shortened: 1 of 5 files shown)


src/AWS.Deploy.CLI/Commands/DeployCommand.cs
```diff
@@ -189,6 +193,67 @@ public async Task ExecuteAsync(string stackName, bool saveCdkProject)
             await orchestrator.DeployRecommendation(cloudApplication, selectedRecommendation);
         }
 
+        private void ConfigureDeployment(Recommendation recommendation, IEnumerable<OptionSettingItem> configurableOptionSettings, UserDeploymentSettings userDeploymentSettings)
```
ganeshnj, line 196: Can you add a test case? Also having a sample input JSON file would give better idea how it is consumed.

ganeshnj (changes requested):

PR #234, opened 2021-06-03
fix: fix telemetry info not being reported

philasmar:
*Issue #, if available:*

*Description of changes:*
fix: fix telemetry info not being reported

### DIFF
--- a/src/AWS.Deploy.CLI/Utilities/CommandLineWrapper.cs
+++ b/src/AWS.Deploy.CLI/Utilities/CommandLineWrapper.cs
@@ -73,6 +73,7 @@ namespace AWS.Deploy.CLI.Utilities
                 processStartInfo.UseShellExecute = _useSeparateWindow;
             }
 
+            UpdateEnvironmentVariables(processStartInfo, environmentVariables);
             _processStartInfoAction?.Invoke(processStartInfo);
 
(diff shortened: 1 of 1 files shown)


ganeshnj (approved):

src/AWS.Deploy.CLI/Utilities/CommandLineWrapper.cs
```diff
@@ -73,6 +73,7 @@ public class CommandLineWrapper : ICommandLineWrapper
                 processStartInfo.UseShellExecute = _useSeparateWindow;
             }
 
+            UpdateEnvironmentVariables(processStartInfo, environmentVariables);
```
ganeshnj, line 76: I wonder how can we make unbreakable as a thought exercise.

ganeshnj (approved):

src/AWS.Deploy.CLI/Utilities/CommandLineWrapper.cs
philasmar: I dont know about unbreakable, but we could abstract the Process class and add the telemetry stuff in there. Anything that spins up a new process would have the telemetry stuff in there.

96malhar (approved):

PR #237, opened 2021-06-10
docs: add Deployment Tool design doc

philasmar:
*Issue #, if available:*
DOTNET-5133

*Description of changes:*
docs: add Deployment Tool design doc


docs/design/deployment-tool.md
```diff
@@ -0,0 +1,173 @@
...
+\
+&nbsp;
+
+## Proposal
+
+The experience that will be created with this tool will allow .NET developers to take their .NET cloud-ready application that they got started in their IDE of choice and easily deploy it to AWS while learning about AWS deployment services and what the appropriate service for their application is.
+
+_We want to onboard developers to AWS quickly, so that they realize the benefits of AWS without spending hours reading documentation._ said Ken Exner, Director, AWS Developer Tools. The new features and improved tools allow developers to make the right choice of compute for their .NET applications, and to rapidly deploy their .NET applications using AWS best practices. They also greatly improve productivity by reducing the amount of non-differentiating code developers need to write and maintain for their applications.
```
normj, line 19: This paragraph is all marketing and doesn't need to be in a tech design document.

normj (changes requested):
I think PR is not what we want to do. I know it is coping large amounts of the content from some of our early design docs but those were written to communicate more to management what we were building. So the documents comes off as a lot more marketing then actual tech design.

docs/design/deployment-tool.md
philasmar: Based on the ticket that irene created https://sim.amazon.com/issues/DOTNET-5133 we wanted the doc to have a problem statement and use cases (which is why the top part looks like a marketing document). How do you propose we move forward?

philasmar: I removed that block.

PR #240, opened 2021-06-15
feat: Use local config file for deployment and enable deployment without any prompts

96malhar:
*Issue #, if available:*
* DOTNET - 4887 (Ability to re-deploy without any prompts)
* DOTNET - 4883 (Use local config file when deploying)

*Description of changes:*
* Added a CLI option flag `--silent` to perform a deployment without any prompts for user input.
* Added a CLI option flag `--apply <config-file-path>`  to perform a deployment using a user provided config file.
* Changed the input to `CommandHandler.Create` to a custom object because that method has a maximum parameter limit of 7.
* Removed the `PreviousDeploymentSettings` class as part of this [ask](https://github.com/aws/aws-dotnet-deploy/pull/240#discussion_r653100627)

An example config file looks as follows:
```
{
    "AWSProfile": "default",
    "AWSRegion": "us-west-2",
    "StackName": "MyAppStack",
    "RecipeId": "AspNetAppEcsFargate",
    "OptionSettingsConfig":{
        "ECSCluster": 
        {
            "CreateNew": true,
            "NewClusterName": "MyNewCluster"
        },
        "ECSServiceName": "MyNewService",
        "DesiredCount": 3,
        "ApplicationIAMRole": 
        {
            "CreateNew": true
        },
        "Vpc": 
        {
            "IsDefault": true
        },
        "AdditionalECSServiceSecurityGroups": "",
        "TaskCpu": 256,
        "TaskMemory": 512,
        "DockerExecutionDirectory": "C:\\codebase"
    }
  }
  
```

Verified that it build correctly by running the integration test suite in the devex gamma account (416945707730)
* CodeBuild Job - aws-dotnet-deploy-test-pr
* Build # - 26


src/AWS.Deploy.CLI/Commands/DeployCommand.cs
```diff
@@ -194,6 +201,129 @@ public async Task ExecuteAsync(string stackName, bool saveCdkProject)
...
+        private string GetCloudApplicationName(string? stackName, UserDeploymentSettings? userDeploymentSettings, List<CloudApplication> deployedApplications)
+        {
+            if (!string.IsNullOrEmpty(stackName))
+                return stackName;
+
+            if (userDeploymentSettings == null || string.IsNullOrEmpty(userDeploymentSettings.StackName))
+            {
+                if (_toolInteractiveService.DisableInteractive)
+                {
+                    var message = "The \"--silent\" CLI argument can only be used if a CDK stack name is provided either via CLI argument or through a deployment-settings file. " +
```
normj, line 294: Be explicit and say which CLI argument to use to set the name. `--stack-name`

src/AWS.Deploy.Common/UserDeploymentSettings.cs
```diff
@@ -0,0 +1,66 @@
...
+                    userDeploymentSettings.TraverseRootToLeaf(userDeploymentSettings.OptionSettingsConfig.Root);
+                return userDeploymentSettings;
+            }
+            catch (Exception ex)
+            {
+                throw new InvalidUserDeploymentSettingsException("An error occured while trying to deserialize the User Deployment Settings file.", ex);
+            }
+        }
+
+        private void TraverseRootToLeaf(JToken node)
```
normj, line 49: This logic fails if the user settings has an empty object. For example if there is an empty `ApplicationIAMRole` object the user gets an error the following error:
```
An error occured while trying to deserialize the User Deployment Settings file.
```

Empty objects for complex shapes should be ignored.

normj: Also the intent of what is being done here should be commented. It took awhile to reverse engineer this to understand we are initially deseralizing to a Newtonsoft JObject and then post processing after deserialization to create a separate settings collection of just the leaf nodes of the JSON object structure.

src/AWS.Deploy.CLI/Commands/DeployCommand.cs
```diff
@@ -194,6 +201,129 @@ public async Task ExecuteAsync(string stackName, bool saveCdkProject)
...
+                    optionSetting = recommendation.GetOptionSetting(optionSettingJsonPath);
+                    isPartOfDeploymentBundle = false;
+                }
+                    
+                if (!recommendation.IsExistingCloudApplication || optionSetting.Updatable)
+                {
+                    object settingValue;
+                    try
+                    {
+                        switch (optionSetting.Type)
```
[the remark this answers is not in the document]
normj: I understand now that when `UserDeploymentSettings` is deserialized it is pruning the option settings to the leaf nodes. Overall I'm not a fan of this approach because it breaks the separation of concerns. Right now `UserDeploymentSettings` has some pretty weird logic of rearranging the data structures for the data in the JSON file just to match what this function needs. It makes the code confusing and hard to follow.

normj (changes requested):
We need some integ tests with deployment happening via settings file. I know profile and region are hard due to integ tests using whatever CodeBuild is configured but we should be able to have some tests for recipe, stack and settings.

src/AWS.Deploy.Common/UserDeploymentSettings.cs
```diff
@@ -0,0 +1,66 @@
...
+    /// </summary>
+    public class UserDeploymentSettings
+    {
+        public string? AWSProfile { get; set; }
+
+        public string? AWSRegion { get; set; }
+
+        public string? StackName { get; set; }
+
+        public string? SelectedRecipeId { get; set; }
```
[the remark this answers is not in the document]
96malhar: Fixed

src/AWS.Deploy.CLI/Commands/DeployCommand.cs
```diff
@@ -194,6 +201,129 @@ public async Task ExecuteAsync(string stackName, bool saveCdkProject)
...
+                    {
+                        throw new InvalidOverrideValueException($"Invalid value {optionSettingValue} for option setting item {optionSettingJsonPath}");
+                    }
+                    
+                    optionSetting.SetValueOverride(settingValue);
+
+                    if (isPartOfDeploymentBundle)
+                        SetDeploymentBundleOptionSetting(recommendation, optionSetting.Id, settingValue);
+
+                    if (optionSetting.Id.Equals("EC2KeyPair"))
```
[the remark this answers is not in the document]
96malhar: Removed the logic to create a _**new**_ EC2 Key Pair using the config file. If a Key Pair name is specified, it will be set as an OptionSettingValue. However, during deployment if found that that key pair does not already exist, then the deployment will fail either at the CDK level or the Cloud Formation level.

src/AWS.Deploy.CLI/Commands/DeployCommand.cs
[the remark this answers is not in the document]
96malhar: The Option Setting type that goes inside the switch case, will always be of a primitive type because it is a leaf Option Setting Item.

src/AWS.Deploy.CLI/Commands/DeployCommand.cs
96malhar: Fixed

src/AWS.Deploy.Common/UserDeploymentSettings.cs
96malhar: Added a comment for the method and handled the case when empty objects are encountered.

src/AWS.Deploy.CLI/Commands/CommandFactory.cs
```diff
@@ -110,26 +113,32 @@ private Command BuildDeployCommand()
...
-            deployCommand.Handler = CommandHandler.Create<string, string, string, string, bool, bool>(async (profile, region, projectPath, stackName, saveCdkProject, diagnostics) =>
+            deployCommand.Handler = CommandHandler.Create(async (DeployCommandHandlerInput input) =>
             {
                 try
                 {
-                    _toolInteractiveService.Diagnostics = diagnostics;
+                    _toolInteractiveService.Diagnostics = input.Diagnostics;
+                    _toolInteractiveService.DisableInteractive = input.Silent;
+                    
+                    var previousSettings = PreviousDeploymentSettings.ReadSettings(input.ProjectPath ?? "", null);
```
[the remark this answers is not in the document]
96malhar: I will handle this as part of a separate PR.

96malhar:
> We need some integ tests with deployment happening via settings file. I know profile and region are hard due to integ tests using whatever CodeBuild is configured but we should be able to have some tests for recipe, stack and settings.

Added integ tests and verified that they run successfully in the Codebuild job inside gamma.

normj (approved):

src/AWS.Deploy.CLI/Commands/DeployCommand.cs
```diff
@@ -194,6 +208,126 @@ public async Task ExecuteAsync(string stackName, bool saveCdkProject)
...
+                var optionSettingJsonPath = entry.Key;
+                var optionSettingValue = entry.Value;
+
+                var isPartOfDeploymentBundle = true;
+                var optionSetting = deploymentBundleDefinition.Parameters.FirstOrDefault(x => x.Id.Equals(optionSettingJsonPath));
+                if (optionSetting == null)
+                {
+                    optionSetting = recommendation.GetOptionSetting(optionSettingJsonPath);
+                    isPartOfDeploymentBundle = false;
+                }
```
philasmar, line 230: I don't like what you are doing here. We don't need to differentiate between the deployment bundle option settings and the rest of the option settings. At the end, they are all option settings with Type Hints. As long as we are executing the type hints, we shouldn't have an issue. You don't need to manually execute the type hint commands based on a switch.

src/AWS.Deploy.CLI/Commands/DeployCommand.cs
```diff
@@ -194,6 +208,126 @@ public async Task ExecuteAsync(string stackName, bool saveCdkProject)
...
+                    {
+                        throw new InvalidOverrideValueException($"Invalid value {optionSettingValue} for option setting item {optionSettingJsonPath}");
+                    }
+                    
+                    optionSetting.SetValueOverride(settingValue);
+
+                    if (isPartOfDeploymentBundle)
+                        SetDeploymentBundleOptionSetting(recommendation, optionSetting.Id, settingValue);
+                }
+            }
```
philasmar, line 262: I don't like what you are doing here. We don't need to differentiate between the deployment bundle option settings and the rest of the option settings. At the end, they are all option settings with Type Hints. As long as we are executing the type hints, we shouldn't have an issue. You don't need to manually execute the type hint commands based on a switch.

src/AWS.Deploy.CLI/Commands/DeployCommand.cs
```diff
@@ -194,6 +208,126 @@ public async Task ExecuteAsync(string stackName, bool saveCdkProject)
...
+                case "DotnetPublishArgs":
+                    new DotnetPublishArgsCommand(_consoleUtilities).OverrideValue(recommendation, settingValue.ToString() ?? "");
+                    break;
+                case "SelfContainedBuild":
+                    new DotnetPublishSelfContainedBuildCommand(_consoleUtilities).OverrideValue(recommendation, (bool)settingValue);
+                    break;
+                default:
+                    throw new OptionSettingItemDoesNotExistException($"The Option Setting Item { optionSettingId } does not exist.");
+            }
+        }
```
philasmar, line 287: I don't like what you are doing here. We don't need to differentiate between the deployment bundle option settings and the rest of the option settings. At the end, they are all option settings with Type Hints. As long as we are executing the type hints, we shouldn't have an issue. You don't need to manually execute the type hint commands based on a switch.

philasmar (changes requested):

src/AWS.Deploy.CLI/Commands/DeployCommand.cs
96malhar: There are 2 issues with trying to invoke TypeHintCommands for **_non-deployment-bundle option settings._**
* The TypeHintCommands are only available for top level option settings (ex. IAMRole, ECSCluster, VPC etc). However, when performing a deployment using a config file, we are only dealing with **_leaf option settings_** (ex. IAMRole.CreateNew, ECSCluster.NewClusterName etc.)
* These TypeHintCommands are only responsible for reading the input from the CLI and create a corresponding TypeHinResponse based on those values. But since we are only dealing **_leaf option settings_**, we can directly set its value by calling in the `SetValueOverride` method. We will lose separation of concern if we amend the existing TypeHintCommands to facilitate the deployment using a config file (and will also complicate things in the process)

For the **_deployment-bundle option settings_**, the TypeHintCommands also perform some validation (since there are no validators specified as part of the recipe file). Due to this dependency I added the `OverrideValue` method that takes in the desired value, performs the **same validations** on it and finally sets it as part of the deployment bundle.

src/AWS.Deploy.CLI/Commands/DeployCommand.cs
96malhar: There are 2 issues with trying to invoke TypeHintCommands for **_non-deployment-bundle option settings._**
* The TypeHintCommands are only available for top level option settings (ex. IAMRole, ECSCluster, VPC etc). However, when performing a deployment using a config file, we are only dealing with **_leaf option settings_** (ex. IAMRole.CreateNew, ECSCluster.NewClusterName etc.)
* These TypeHintCommands are only responsible for reading the input from the CLI and create a corresponding TypeHinResponse based on those values. But since we are only dealing with **_leaf option settings_**, we can directly set its value by calling in the `SetValueOverride` method. We will lose separation of concern if we amend the existing TypeHintCommands to facilitate the deployment using a config file (and will also complicate things in the process)

For the **_deployment-bundle option settings_**, the TypeHintCommands also perform some validation (since there are no validators specified as part of the recipe file). Due to this dependency I added the `OverrideValue` method that takes in the desired value, performs the **same validations** on it and finally sets it as part of the deployment bundle.

src/AWS.Deploy.CLI/Commands/DeployCommand.cs
96malhar: There are 2 issues with trying to invoke TypeHintCommands for **_non-deployment-bundle option settings._**
* The TypeHintCommands are only available for top level option settings (ex. IAMRole, ECSCluster, VPC etc). However, when performing a deployment using a config file, we are only dealing with **_leaf option settings_** (ex. IAMRole.CreateNew, ECSCluster.NewClusterName etc.)
* These TypeHintCommands are only responsible for reading the input from the CLI and create a corresponding TypeHinResponse based on those values. But since we are only dealing with **_leaf option settings_**, we can directly set its value by calling in the `SetValueOverride` method. We will lose separation of concern if we amend the existing TypeHintCommands to facilitate the deployment using a config file (and will also complicate things in the process)

For the **_deployment-bundle option settings_**, the TypeHintCommands also perform some validation (since there are no validators specified as part of the recipe file). Due to this dependency I added the `OverrideValue` method that takes in the desired value, performs the **same validations** on it and finally sets it as part of the deployment bundle.

philasmar (approved):

PR #241, opened 2021-06-16
fix: Only set AWS credentials for processes that require it

96malhar:
*Issue #, if available:*
DOTNET-5136

*Description of changes:*
Currently we are setting the AWS credentials as environment variables for every command line process. This PR aims to mitigate that by only setting the credentials for processes that require it.

The following processes need AWS credentials:
1. Docker Login Command - This command needs the credentials to generate authorization tokens which are passed as `--username` and `--password` arguments. These tokens are generate by calling the appropriate API using the ECR service client - [proof](https://github.com/aws/aws-dotnet-deploy/blob/main/src/AWS.Deploy.Orchestration/Data/AWSResourceQueryer.cs#L206)
2. CDK Bootstrap command - The credentials need to be passed as environment variables.
3. CDK deploy command - The credentials need to be passed as environment variables.

### DIFF
--- a/src/AWS.Deploy.Orchestration/CdkProjectHandler.cs
+++ b/src/AWS.Deploy.Orchestration/CdkProjectHandler.cs
@@ -50,13 +50,15 @@ namespace AWS.Deploy.Orchestration
             _interactiveService.LogMessageLine("Starting deployment of CDK Project");
 
             // Ensure region is bootstrapped
-            await _commandLineWrapper.Run($"npx cdk bootstrap aws://{session.AWSAccountId}/{session.AWSRegion}");
+            await _commandLineWrapper.Run($"npx cdk bootstrap aws://{session.AWSAccountId}/{session.AWSRegion}",
+                needAwsCredentials: true);
 
             // Handover to CDK command line tool
             // Use a CDK Context parameter to specify the settings file that has been serialized.
             await _commandLineWrapper.Run( $"npx cdk deploy --require-approval never -c {Constants.CloudFormationIdentifier.SETTINGS_PATH_CDK_CONTEXT_PARAMETER}=\"{appSettingsFilePath}\"",
                 workingDirectory: cdkProjectPath,
-                environmentVariables: environmentVariables);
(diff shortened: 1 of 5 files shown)


src/AWS.Deploy.Orchestration/CdkProjectHandler.cs
```diff
@@ -50,13 +50,15 @@ public async Task CreateCdkDeployment(OrchestratorSession session, CloudApplicat
...
+            await _commandLineWrapper.Run($"npx cdk bootstrap aws://{session.AWSAccountId}/{session.AWSRegion}",
+                needAwsCredentials: false);
 
             // Handover to CDK command line tool
             // Use a CDK Context parameter to specify the settings file that has been serialized.
             await _commandLineWrapper.Run( $"npx cdk deploy --require-approval never -c {Constants.CloudFormationIdentifier.SETTINGS_PATH_CDK_CONTEXT_PARAMETER}=\"{appSettingsFilePath}\"",
                 workingDirectory: cdkProjectPath,
-                environmentVariables: environmentVariables);
+                environmentVariables: environmentVariables,
+                needAwsCredentials: false);
```
normj, line 61: When doing CDK deploy is when we should pass in AWS credentials. We need to be sure the profile that has been configured is what credentials are using.

I assume this worked for you because you have a default profile which is just being by CDK deploy in your case because no credentials are passed. Please test this feature without having a default profile configured on your system.

normj (changes requested):

src/AWS.Deploy.Orchestration/CdkProjectHandler.cs
96malhar: Fixed this and tested my changes by removing the default profile.

normj (approved):

96malhar:
> Why are you explicitly setting needAwsCredentials to false when that is the default. You don't need to do any modifications to most of the calls. Just the ones that needs credentials.

Fixed the function calls by not passing in an explicit value where only the default value is required.

philasmar (approved):

PR #245, opened 2021-06-18
feat: add GetConfigSettings API to retrieve Option Setting Items in Server Mode

philasmar:
*Issue #, if available:*
IDE-5165

*Description of changes:*
feat: add GetConfigSettings API to retrieve Option Setting Items in Server Mode

### DIFF
--- a/src/AWS.Deploy.CLI/ServerMode/Controllers/DeploymentController.cs
+++ b/src/AWS.Deploy.CLI/ServerMode/Controllers/DeploymentController.cs
@@ -27,6 +27,7 @@ using AWS.Deploy.CLI.Extensions;
 using AWS.Deploy.Orchestration.Utilities;
 using Microsoft.AspNetCore.Authorization;
 using Amazon.Runtime;
+using AWS.Deploy.Common.Recipes;
 
 namespace AWS.Deploy.CLI.ServerMode.Controllers
 {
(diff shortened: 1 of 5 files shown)


src/AWS.Deploy.CLI/ServerMode/Controllers/DeploymentController.cs
```diff
@@ -127,6 +127,62 @@ public async Task<IActionResult> GetRecommendations(string sessionId)
...
+
+            var deploymentBundleDefinition = orchestrator.GetDeploymentBundleDefinition(state.SelectedRecommendation);
+
+            var configurableOptionSettings = state.SelectedRecommendation.Recipe.OptionSettings.Union(deploymentBundleDefinition.Parameters);
+
+            var output = new GetOptionSettingsOutput();
+
+            foreach (var setting in configurableOptionSettings)
+            {
+                if ((state.SelectedRecommendation.IsExistingCloudApplication && !setting.Updatable) || !state.SelectedRecommendation.IsOptionSettingDisplayable(setting))
```
normj, line 160: We should still send non updatable settings with a flag that says the setting is read only. That way the IDE can display all of the current settings for the deployment.

src/AWS.Deploy.CLI/ServerMode/Controllers/DeploymentController.cs
```diff
@@ -127,6 +127,62 @@ public async Task<IActionResult> GetRecommendations(string sessionId)
...
+                    continue;
+
+                var settingSummary = new OptionSettingItemSummary(setting.Id, setting.Name, setting.Description, setting.Type.ToString())
+                {
+                    TypeHint = setting.TypeHint?.ToString(),
+                    Value = state.SelectedRecommendation.GetOptionSettingValue(setting),
+                    Advanced = setting.AdvancedSetting
+                };
+
+                setting.ChildOptionSettings.ForEach(childSetting =>
```
normj, line 170: Does this work if the option settings are more then 2 level deep? Meaning a object setting containing an object setting.

normj (changes requested):

src/AWS.Deploy.CLI/ServerMode/Controllers/DeploymentController.cs
philasmar: We do not support that in general. An Object type can only have children of primitive types.

src/AWS.Deploy.CLI/ServerMode/Controllers/DeploymentController.cs
philasmar: Fixed

src/AWS.Deploy.CLI/ServerMode/Controllers/DeploymentController.cs
ganeshnj: This is not correct, our logic are written in way that it can handle any arbitrary depth. By chance, we only have 2 level settings in recipe.

philasmar: Added recursive logic to handle a depth more than 2

ganeshnj (approved):

normj (approved):

PR #246, opened 2021-06-21
fix: Update EBS platformArn regex validation to make the account ID as optional

96malhar:
*Description of changes:*
* Fixed regex validation to make the account ID as optional in Elastic Beanstalk platformArn.
* Included valid regions in the test-cases.

image

The above image shows all the default settings when performing a deployment with an Elastic Beanstalk recipe. As seen, the platform ARN does not contain the account-ID

The regex has been inferred from this [doc](https://docs.aws.amazon.com/elasticbeanstalk/latest/dg/AWSHowTo.iam.policies.arn.html)


ganeshnj:
Could you add source of regex in PR description?

normj (approved):

ganeshnj (approved):

PR #249, opened 2021-06-22
fix: fix invalid date format exception thrown from VS Toolkit

philasmar:
*Issue #, if available:*
IDE-5162

*Description of changes:*
Fix for the error:
"'DateTimeInvalidLocalFormat' : 'A UTC DateTime is being converted to text in a format that is only correct for local times. This can happen when calling DateTime.ToString using the 'z' format specifier, which will include a local time zone offset in the output. In that case, either use the 'Z' format specifier, which designates a UTC time, or use the 'o' format string, which is the recommended way to persist a DateTime in text. This can also occur when passing a DateTime to be serialized by XmlConvert or DataSet. If using XmlConvert.ToString, pass in XmlDateTimeSerializationMode.RoundtripKind to serialize correctly. If using DataSet, set the DateTimeMode on the DataColumn object to DataSetDateTime.Utc. '"

### DIFF
--- a/src/AWS.Deploy.ServerMode.Client/ServerModeHttpClient.cs
+++ b/src/AWS.Deploy.ServerMode.Client/ServerModeHttpClient.cs
@@ -61,7 +61,7 @@ namespace AWS.Deploy.ServerMode.Client
                 {"awsAccessKeyId", credentials.AccessKey },
                 {"awsSecretKey", credentials.SecretKey },
                 {"requestId", Guid.NewGuid().ToString() },
-                {"issueDate", DateTime.UtcNow.ToString("yyyy-MM-dd'T'HH:mm:ss.fffzzz", DateTimeFormatInfo.InvariantInfo) }
+                {"issueDate", DateTime.UtcNow.ToString("yyyy-MM-dd'T'HH:mm:ss.fffZ", DateTimeFormatInfo.InvariantInfo) }
             };
 
             if(!string.IsNullOrEmpty(credentials.Token))


src/AWS.Deploy.ServerMode.Client/ServerModeHttpClient.cs
```diff
@@ -61,7 +61,7 @@ public static void AddAuthorizationHeader(HttpRequestMessage request, ImmutableC
                 {"awsAccessKeyId", credentials.AccessKey },
                 {"awsSecretKey", credentials.SecretKey },
                 {"requestId", Guid.NewGuid().ToString() },
-                {"issueDate", DateTime.UtcNow.ToString("yyyy-MM-dd'T'HH:mm:ss.fffzzz", DateTimeFormatInfo.InvariantInfo) }
```
ganeshnj, line 64: I never liked these formats, can we use simple good epoch time? 

https://www.epochconverter.com/

philasmar: Most APIs usually use the a more human readable form like the one we use. I am not a fan of epoch time.

ganeshnj: I can disagree and commit here.

Just like we never store the age, we save date of birth. Time should be agnostic of all factor.

Epoch time 
- doesn't have timezone
- one universal format works on all platforms
- no issues with daylight saving time
- apps can convert it local time based on their needs

Yes it comes with readability issue but there are tones of tools to convert it.

philasmar: The format we are using "ISO 8601" should have the benefits of epoch time. So we should be fine.

normj (approved):

src/AWS.Deploy.ServerMode.Client/ServerModeHttpClient.cs
normj: Epoch dates are rarely used in pure .NET solutions.

ganeshnj (approved):

PR #251, opened 2021-06-25
feat: add SetConfigSettings API to set configuration settings

philasmar:
*Issue #, if available:*

*Description of changes:*
Added SetConfigSettings that takes a list of updated settings and applies them. If there is an error setting any of them, a list of errors is returned.

### DIFF
--- a/src/AWS.Deploy.CLI/ServerMode/Controllers/DeploymentController.cs
+++ b/src/AWS.Deploy.CLI/ServerMode/Controllers/DeploymentController.cs
@@ -182,6 +182,46 @@ namespace AWS.Deploy.CLI.ServerMode.Controllers
         }
 
         /// <summary>
+        /// Applies a value for a list of option setting items on the selected recommendation.
+        /// Option setting updates are provided as Key Value pairs with the Key being the JSON path to the leaf node.
+        /// Only primitive data types are supported for Value updates. The Value is a string value which will be parsed as its corresponding data type.
+        /// </summary>
+        [HttpPut("session/<sessionId>/settings")]
+        [SwaggerOperation(OperationId = "ApplyConfigSettings")]
+        [SwaggerResponse(200, type: typeof(ApplyConfigSettingsOutput))]
+        [Authorize]
+        public IActionResult ApplyConfigSettings(string sessionId, [FromBody] ApplyConfigSettingsInput input)
+        {
+            var state = _stateServer.Get(sessionId);
+            if (state == null)
+            {
+                return NotFound($"Session ID {sessionId} not found.");
+            }
+
+            if (state.SelectedRecommendation == null)
+            {
+                return NotFound($"A deployment target is not set for Session ID {sessionId}.");
+            }
+
+            var output = new ApplyConfigSettingsOutput();
+
+            foreach (var updatedSetting in input.UpdatedSettings)
+            {
+                try
+                {
+                    var setting = state.SelectedRecommendation.GetOptionSetting(updatedSetting.Key);
+                    setting.SetValueOverride(updatedSetting.Value);
+                }
+                catch (Exception ex)
+                {
+                    output.FailedConfigUpdates.Add(updatedSetting.Key, ex.Message);
+                }
+            }
+
+            return Ok(output);
+        }
+        
+        /// <summary>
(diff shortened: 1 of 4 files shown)


src/AWS.Deploy.CLI/ServerMode/Models/SetConfigSettingsInput.cs
```diff
@@ -0,0 +1,12 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.\r
+// SPDX-License-Identifier: Apache-2.0
+
+using System.Collections.Generic;
+
+namespace AWS.Deploy.CLI.ServerMode.Models
+{
+    public class SetConfigSettingsInput
+    {
+        public Dictionary<string, string> UpdatedSettings { get; set; } = new Dictionary<string, string>();
```
ganeshnj, line 10: I expect dictionary value to be `object`, how is it going to work with nested option settings.

If the key is path to the leaf setting, then it is worth calling `JsonPath` in docs to make it clear for the readers.

Can you also add some documentation how is it working?

ganeshnj (changes requested):

src/AWS.Deploy.CLI/ServerMode/Models/SetConfigSettingsInput.cs
philasmar: I added more documentation on the operation.

ganeshnj (approved):

src/AWS.Deploy.CLI/ServerMode/Controllers/DeploymentController.cs
```diff
@@ -127,6 +127,46 @@ public async Task<IActionResult> GetRecommendations(string sessionId)
...
+                    var setting = state.SelectedRecommendation.GetOptionSetting(updatedSetting.Key);
+                    setting.SetValueOverride(updatedSetting.Value);
+                }
+                catch (Exception ex)
+                {
+                    output.FailedConfigUpdates.Add(updatedSetting.Key, ex.Message);
+                }
+            }
+
+            return Ok(output);
```
normj, line 167: Should the response contain the current state of the settings? Some settings default values could have changed and the state of updatablity can change as well. We can either return them here along with the errors or the UI has to make a separate call get the settings again.

src/AWS.Deploy.CLI/ServerMode/Controllers/DeploymentController.cs
```diff
@@ -127,6 +127,46 @@ public async Task<IActionResult> GetRecommendations(string sessionId)
...
+        /// <summary>
+        /// Sets a value for a list of option setting items on the selected recommendation.
+        /// Option setting updates are provided as Key Value pairs with the Key being the JSON path to the leaf node.
+        /// Only primitive data types are supported for Value updates. The Value is a string value which will be parsed as its corresponding data type.
+        /// </summary>
+        [HttpPut("session/<sessionId>/settings")]
+        [SwaggerOperation(OperationId = "SetConfigSettings")]
+        [SwaggerResponse(200, type: typeof(SetConfigSettingsOutput))]
+        [Authorize]
+        public IActionResult SetConfigSettings(string sessionId, [FromBody] SetConfigSettingsInput input)
```
normj, line 139: nit: I kind of prefer the term `ApplyConfigSettings` instead of `SetConfigSettings` because it gives more indication there will be side effects when setting a specific setting.

normj (changes requested):

src/AWS.Deploy.CLI/ServerMode/Controllers/DeploymentController.cs
philasmar: I thought about this after reading the document and did a small research and the convention is just to do the update for PUT requests. To get the new state, you should do another GET. That is what I am doing from the toolkit. After a PUT, I do another GET to check the items.

src/AWS.Deploy.CLI/ServerMode/Controllers/DeploymentController.cs
philasmar: I changed the name.

normj (approved):

PR #253, opened 2021-06-29
feat: add GetRecipe API to describe Recipes returned by other APIs

philasmar:
*Issue #, if available:*

*Description of changes:*
feat: add GetRecipe API to describe Recipes returned by other APIs

### DIFF
--- a/src/AWS.Deploy.CLI/ServerMode/Controllers/RecipeController.cs
+++ b/src/AWS.Deploy.CLI/ServerMode/Controllers/RecipeController.cs
@@ -1,0 +1,51 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.\r
+// SPDX-License-Identifier: Apache-2.0
+
+using System.Linq;
+using AWS.Deploy.CLI.ServerMode.Models;
+using AWS.Deploy.Orchestration;
+using Microsoft.AspNetCore.Mvc;
+using Swashbuckle.AspNetCore.Annotations;
+
+namespace AWS.Deploy.CLI.ServerMode.Controllers
+{
+    [Produces("application/json")]
+    [ApiController]
+    [Route("api/v1/[controller]")]
+    public class RecipeController : ControllerBase
+    {
+        /// <summary>
+        /// Gets a summary of the specified Recipe.
+        /// </summary>
+        [HttpGet("{recipeId}")]
+        [SwaggerOperation(OperationId = "GetRecipe")]
+        [SwaggerResponse(200, type: typeof(RecipeSummary))]
+        public IActionResult GetRecipe(string recipeId, [FromQuery] string? projectPath = null)
+        {
+            if (string.IsNullOrEmpty(recipeId))
+            {
+                return BadRequest($"A Recipe ID was not provided.");
+            }
+
+            var recipeDefinitions = RecipeHandler.GetRecipeDefinitions();
+            var selectedRecipeDefinition = recipeDefinitions.FirstOrDefault(x => x.Id.Equals(recipeId));
+
+            if (selectedRecipeDefinition == null)
+            {
+                return BadRequest($"Recipe ID {recipeId} not found.");
+            }
+
+            var output = new RecipeSummary(
+                selectedRecipeDefinition.Id,
+                selectedRecipeDefinition.Version,
+                selectedRecipeDefinition.Name,
+                selectedRecipeDefinition.Description,
+                selectedRecipeDefinition.TargetService,
+                selectedRecipeDefinition.DeploymentType.ToString(),
+                selectedRecipeDefinition.DeploymentBundle.ToString()
+            );
+
+            return Ok(output);
+        }
+    }
+}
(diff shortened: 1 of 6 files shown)


src/AWS.Deploy.CLI/ServerMode/Controllers/DeploymentController.cs
```diff
@@ -128,6 +128,46 @@ public async Task<IActionResult> GetRecommendations(string sessionId)
...
+            }
+
+            var output = new RecipeSummary(
+                recommendation.Recipe.Id,
+                recommendation.Recipe.Version,
+                recommendation.Recipe.Name,
+                recommendation.Recipe.Description,
+                recommendation.Recipe.TargetService,
+                recommendation.Recipe.DeploymentType.ToString(),
+                recommendation.Recipe.DeploymentBundle.ToString()
```
ganeshnj, line 165: Any reason behind converting the enum to string? enumerations are supported in JSON.

src/AWS.Deploy.CLI/ServerMode/Controllers/DeploymentController.cs
```diff
@@ -128,6 +128,46 @@ public async Task<IActionResult> GetRecommendations(string sessionId)
...
         }
 
+        /// <summary>
+        /// Gets a summary of the Recipe for the session's project.
+        /// </summary>
+        [HttpGet("session/<sessionId>/recipe/<recipeId>")]
+        [SwaggerOperation(OperationId = "GetRecipe")]
+        [SwaggerResponse(200, type: typeof(RecipeSummary))]
+        [Authorize]
+        public IActionResult GetRecipe(string sessionId, string recipeId)
```
normj, line 138: Seems like getting recipe metadata data has nothing to do with a deployment or an active session. I think we should add a separate ASP.NET Core controller for recipe metadata and the root path can me `recipe` instead of `session`. We don't need to make the methods require authentication or a session.

normj (changes requested):

src/AWS.Deploy.CLI/ServerMode/Controllers/DeploymentController.cs
philasmar: Leaving them as enums blows up from the toolkit's side when it's trying to deserialize them.

src/AWS.Deploy.CLI/ServerMode/Controllers/DeploymentController.cs
philasmar: Fixed

ganeshnj (approved):

src/AWS.Deploy.CLI/ServerMode/Controllers/RecipeController.cs
```diff
@@ -0,0 +1,51 @@
...
+        /// Gets a summary of the specified Recipe.
+        /// </summary>
+        [HttpGet("{recipeId}")]
+        [SwaggerOperation(OperationId = "GetRecipe")]
+        [SwaggerResponse(200, type: typeof(RecipeSummary))]
+        public IActionResult GetRecipe(string recipeId)
+        {
+            if (string.IsNullOrEmpty(recipeId))
+            {
+                return NotFound($"A Recipe ID was not provided.");
```
normj, line 27: Instead of `NotFound` I think `BadRequest` makes more sense.

normj (changes requested):
We need tests for the new API

src/AWS.Deploy.CLI/ServerMode/Controllers/RecipeController.cs
philasmar: Fixed

philasmar:
> We need tests for the new API

I added some tests.

src/AWS.Deploy.CLI/ServerMode/Controllers/RecipeController.cs
```diff
@@ -0,0 +1,51 @@
...
+    [Route("api/v1/[controller]")]
+    public class RecipeController : ControllerBase
+    {
+        /// <summary>
+        /// Gets a summary of the specified Recipe.
+        /// </summary>
+        [HttpGet("{recipeId}")]
+        [SwaggerOperation(OperationId = "GetRecipe")]
+        [SwaggerResponse(200, type: typeof(RecipeSummary))]
+        public IActionResult GetRecipe(string recipeId)
```
normj, line 23: I think we should take the project path as an optional parameter coming in as a header. That way if the project path is available we can search for the recipes based on the context of a specific project. That way when we save CDK projects we can lookup recipes the upcoming json mapping file.  We don't need to do anything with the value till we have the save CDK project feature implemented but I think we should get the API define fully from the start.

normj (changes requested):

src/AWS.Deploy.CLI/ServerMode/Controllers/RecipeController.cs
philasmar: I have added the optional query parameter.

normj (approved):

PR #254, opened 2021-06-29
fix: Fix misleading language in user input prompt to deploy to a new or an existing Cloudformation stack

96malhar:
*Issue #, if available:*
DOTNET-5193

*Description of changes:*
This PR fixes the misleading language in user input prompt to deploy to a new or an existing Cloudformation stack.

image
image

### DIFF
--- a/src/AWS.Deploy.CLI/ConsoleUtilities.cs
+++ b/src/AWS.Deploy.CLI/ConsoleUtilities.cs
@@ -19,13 +19,13 @@ namespace AWS.Deploy.CLI
     public interface IConsoleUtilities
     {
         Recommendation AskToChooseRecommendation(IList<Recommendation> recommendations);
-        string AskUserToChoose(IList<string> values, string title, string? defaultValue);
-        T AskUserToChoose<T>(IList<T> options, string title, T defaultValue)
+        string AskUserToChoose(IList<string> values, string title, string? defaultValue, string? defaultChoosePrompt = null);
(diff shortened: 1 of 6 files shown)


src/AWS.Deploy.CLI/ConsoleUtilities.cs
```diff
@@ -246,14 +258,17 @@ public UserResponse<T> AskUserToChooseOrCreateNew<T>(IEnumerable<T> options, str
             };
         }
 
-        public string AskUserForValue(string message, string defaultValue, bool allowEmpty, string resetValue = "", params Func<string, string>[] validators)
+        public string AskUserForValue(string message, string defaultValue, bool allowEmpty, string resetValue = "", bool selectOrCreateNewStackName = false, params Func<string, string>[] validators)
```
normj, line 261: I don't want to have to add another bool every time we come up with another question we need a special label. We should just allow the method to take an option "Enter" label. If one exists this function uses otherwise uses the default.

normj (changes requested):

src/AWS.Deploy.CLI/ConsoleUtilities.cs
96malhar: Replaced the bool with flexible default prompts and default labels.

normj (approved):

ganeshnj (approved):

PR #257, opened 2021-07-07
feat: Create command to save the CDK deployment project

96malhar:
*Issue #, if available:*
1. `DOTNET-5245`
2. `DOTNET-5246`

*Description of changes:*
1. Add a new command that generate the CDK deployment project and saves it at a user-specified directory. 
2. Generate a deployment recipe snapshot that is taken at the time of the CDK deployment project generation.

### DIFF
--- a/src/AWS.Deploy.CLI/Commands/CommandFactory.cs
+++ b/src/AWS.Deploy.CLI/Commands/CommandFactory.cs
@@ -15,6 +15,7 @@ using AWS.Deploy.Orchestration.CDK;
 using AWS.Deploy.Orchestration.Data;
 using AWS.Deploy.Orchestration.Utilities;
 using AWS.Deploy.CLI.Commands.CommandHandlerInput;
+using AWS.Deploy.Common.IO;
 
 namespace AWS.Deploy.CLI.Commands
 {
@@ -32,6 +33,8 @@ namespace AWS.Deploy.CLI.Commands
         private static readonly Option<bool> _optionDiagnosticLogging = new(new []{"-d", "--diagnostics"}, "Enable diagnostic output.");
         private static readonly Option<string> _optionApply = new("--apply", "Path to the deployment settings file to be applied.");
         private static readonly Option<bool> _optionDisableInteractive = new(new []{"-s", "--silent" }, "Disable interactivity to deploy without any prompts for user input.");
+        private static readonly Option<string> _optionOutputDirectory = new(new[]{"-o", "--output"}, "Directory path in which the CDK deployment project will be saved.");
+        private static readonly Option<string> _optionProjectDisplayName = new(new[] { "--project-display-name" }, "The name of the deployment project that will be displayed in the list of available deployment options.");
 
         private readonly IToolInteractiveService _toolInteractiveService;
         private readonly IOrchestratorInteractiveService _orchestratorInteractiveService;
@@ -98,6 +101,7 @@ namespace AWS.Deploy.CLI.Commands
             rootCommand.Add(BuildDeployCommand());
             rootCommand.Add(BuildListCommand());
             rootCommand.Add(BuildDeleteCommand());
+            rootCommand.Add(BuildDeploymentProjectCommand());
             rootCommand.Add(BuildServerModeCommand());
 
             return rootCommand;
@@ -305,6 +309,81 @@ namespace AWS.Deploy.CLI.Commands
                 }
             });
             return listCommand;
+        }
+
+        /// <summary>
+        /// Builds the top level command called "deployment-project" which supports the creation and saving of the
+        /// CDK deployment project.
+        /// </summary>
+        /// <returns>An instance of the <see cref="Command"/> class</returns>
+        private Command BuildDeploymentProjectCommand()
+        {
+            var deploymentProjectCommand = new Command("deployment-project",
+                "Save the deployment project inside a user provided directory path.");
+
+            var generateDeploymentProjectCommand = new Command("generate",
+                "Save the deployment project inside a user provided directory path without proceeding with a deployment")
+            {
+                _optionOutputDirectory,
+                _optionDiagnosticLogging,
+                _optionProjectPath,
+                _optionProjectDisplayName
+            };
+
+            generateDeploymentProjectCommand.Handler = CommandHandler.Create(async (GenerateDeploymentProjectCommandHandlerInput input) =>
+            {
+                try
+                {
+                    _toolInteractiveService.Diagnostics = input.Diagnostics;
+                    var projectDefinition = await _projectParserUtility.Parse(input.ProjectPath ?? "");
+
+                    var saveDirectory = input.Output ?? "";
+                    var projectDisplayName = input.ProjectDisplayName ?? "";
+
+                    OrchestratorSession session = new OrchestratorSession(projectDefinition);
+
+                    var targetApplicationFullPath = new DirectoryManager().GetDirectoryInfo(projectDefinition.ProjectPath).FullName;
+
+                    var generateDeploymentProject = new GenerateDeploymentProjectCommand(
+                        _toolInteractiveService,
+                        _consoleUtilities,
+                        _cdkProjectHandler,
+                        _commandLineWrapper,
+                        new DirectoryManager(),
+                        new FileManager(),
+                        session,
+                        targetApplicationFullPath);
+
+                    await generateDeploymentProject.ExecuteAsync(saveDirectory, projectDisplayName);
+
+                    return CommandReturnCodes.SUCCESS;
+                }
+                catch (Exception e) when (e.IsAWSDeploymentExpectedException())
+                {
+                    _toolInteractiveService.WriteErrorLine("Failed to generate deployment project.");
+                    if (input.Diagnostics)
+                        _toolInteractiveService.WriteErrorLine(e.PrettyPrint());
+                    else
+                    {
+                        _toolInteractiveService.WriteErrorLine(string.Empty);
+                        _toolInteractiveService.WriteErrorLine(e.Message);
+                    }
+                    // bail out with an non-zero return code.
+                    return CommandReturnCodes.USER_ERROR;
+                }
+                catch (Exception e)
+                {
+                    // This is a bug
+                    _toolInteractiveService.WriteErrorLine(
+                        "Unhandled exception.  This is a bug.  Please copy the stack trace below and file a bug at https://github.com/aws/aws-dotnet-deploy. " +
+                        e.PrettyPrint());
+
+                    return CommandReturnCodes.UNHANDLED_EXCEPTION;
+                }
+            });
+
+            deploymentProjectCommand.Add(generateDeploymentProjectCommand);
+            return deploymentProjectCommand;
         }
 
         private Command BuildServerModeCommand()
--- a/src/AWS.Deploy.CLI/Commands/CommandHandlerInput/GenerateDeploymentProjectCommandHandlerInput.cs
+++ b/src/AWS.Deploy.CLI/Commands/CommandHandlerInput/GenerateDeploymentProjectCommandHandlerInput.cs
@@ -1,0 +1,16 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+namespace AWS.Deploy.CLI.Commands.CommandHandlerInput
+{
+    /// <summary>
+    /// This class maps the command line options for the "deployment-project generate" command to the appropriate C# properties.
+    /// </summary>
+    public class GenerateDeploymentProjectCommandHandlerInput
+    {
+        public string? ProjectPath { get; set; }
+        public bool Diagnostics { get; set; }
+        public string? Output { get; set; }
+        public string? ProjectDisplayName { get; set; }
+    }
+}
--- a/src/AWS.Deploy.CLI/Commands/GenerateDeploymentProjectCommand.cs
+++ b/src/AWS.Deploy.CLI/Commands/GenerateDeploymentProjectCommand.cs
@@ -1,0 +1,231 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+using System;
+using System.Collections.Generic;
+using System.IO;
+using System.Threading.Tasks;
+using AWS.Deploy.Common;
+using AWS.Deploy.Common.IO;
+using AWS.Deploy.Common.Recipes;
+using AWS.Deploy.Orchestration;
+using AWS.Deploy.Orchestration.Utilities;
+using AWS.Deploy.Recipes;
+using Newtonsoft.Json;
+
+namespace AWS.Deploy.CLI.Commands
+{
+    /// <summary>
+    /// The class supports the functionality to create a new CDK project and save it at a customer
+    /// specified directory location.
+    /// </summary>
+    public class GenerateDeploymentProjectCommand
+    {
+        private readonly IToolInteractiveService _toolInteractiveService;
+        private readonly IConsoleUtilities _consoleUtilities;
+        private readonly ICdkProjectHandler _cdkProjectHandler;
+        private readonly ICommandLineWrapper _commandLineWrapper;
+        private readonly IDirectoryManager _directoryManager;
+        private readonly IFileManager _fileManager;
+        private readonly OrchestratorSession _session;
+        private readonly string _targetApplicationFullPath;
+        
+        public GenerateDeploymentProjectCommand(
+            IToolInteractiveService toolInteractiveService,
+            IConsoleUtilities consoleUtilities,
+            ICdkProjectHandler cdkProjectHandler,
+            ICommandLineWrapper commandLineWrapper,
+            IDirectoryManager directoryManager,
+            IFileManager fileManager,
+            OrchestratorSession session,
+            string targetApplicationFullPath)
+        {
+            _toolInteractiveService = toolInteractiveService;
+            _consoleUtilities = consoleUtilities;
+            _cdkProjectHandler = cdkProjectHandler;
+            _commandLineWrapper = commandLineWrapper;
+            _directoryManager = directoryManager;
+            _fileManager = fileManager;
+            _session = session;
+            _targetApplicationFullPath = targetApplicationFullPath;
+        }
+
+        /// <summary>
+        /// This method takes a user specified directory path and generates the CDK deployment project at this location.
+        /// If the provided directory path is an empty string, then a default directory is created to save the CDK deployment project.
+        /// </summary>
+        /// <param name="saveCdkDirectoryPath">An absolute or a relative path provided by the user.</param>
+        /// <param name="projectDisplayName">The name of the deployment project that will be displayed in the list of available deployment options.</param>
+        /// <returns></returns>
+        public async Task ExecuteAsync(string saveCdkDirectoryPath, string projectDisplayName)
+        {
+            var orchestrator = new Orchestrator(_session, new[] { RecipeLocator.FindRecipeDefinitionsPath() });
+            var recommendations = await GenerateDeploymentRecommendations(orchestrator);
+            var selectedRecommendation = _consoleUtilities.AskToChooseRecommendation(recommendations);
+
+            if (string.IsNullOrEmpty(saveCdkDirectoryPath))
+                saveCdkDirectoryPath = GenerateDefaultSaveDirectoryPath();
+
+            var newDirectoryCreated = CreateSaveCdkDirectory(saveCdkDirectoryPath);
+
+            var (isValid, errorMessage) = ValidateSaveCdkDirectory(saveCdkDirectoryPath);
+            if (!isValid)
+            {
+                if (newDirectoryCreated)
+                    _directoryManager.Delete(saveCdkDirectoryPath);
+                throw new InvalidSaveDirectoryForCdkProject(errorMessage);
+            }
+            
+            var directoryUnderSourceControl = await IsDirectoryUnderSourceControl(saveCdkDirectoryPath);
+            if (!directoryUnderSourceControl)
+            {
+                var userPrompt = "Warning: The target directory is not being tracked by source control. If the saved deployment " +
+                    "project is used for deployment it is important that the deployment project is retained to allow " +
+                    "future redeployments to previously deployed applications. " + Environment.NewLine + Environment.NewLine +
+                    "Do you still want to continue?";
+
+                _toolInteractiveService.WriteLine();
+                var yesNoResult = _consoleUtilities.AskYesNoQuestion(userPrompt, YesNo.Yes);
+
+                if (yesNoResult == YesNo.No)
+                {
+                    if (newDirectoryCreated)
+                        _directoryManager.Delete(saveCdkDirectoryPath);
+                    return;
+                }
+            }
+
+            await _cdkProjectHandler.CreateCdkProjectForDeployment(selectedRecommendation, _session, saveCdkDirectoryPath);
+            await GenerateDeploymentRecipeSnapShot(selectedRecommendation, saveCdkDirectoryPath, projectDisplayName);
+
+            var saveCdkDirectoryInfo = _directoryManager.GetDirectoryInfo(saveCdkDirectoryPath);
+            _toolInteractiveService.WriteLine();
+            _toolInteractiveService.WriteLine($"The CDK deployment project is saved at: {saveCdkDirectoryInfo.FullName}");
+        }
+
+        /// <summary>
+        /// This method generates the appropriate recommendations for the target deployment project.
+        /// </summary>
+        /// <param name="orchestrator"><see cref="Orchestrator"/></param>
+        /// <returns>A list of <see cref="Recommendation"/></returns>
+        private async Task<List<Recommendation>> GenerateDeploymentRecommendations(Orchestrator orchestrator)
+        {
+            var recommendations = await orchestrator.GenerateDeploymentRecommendations(forDeployment: false);
+            if (recommendations.Count == 0)
+            {
+                throw new FailedToGenerateAnyRecommendations("The project you are trying to deploy is currently not supported.");
+            }
+            return recommendations;
+        }
+
+        /// <summary>
+        /// This method takes the path to the target deployment project and creates
+        /// a default save directory inside the parent folder of the current directory.
+        /// For example:
+        /// Target project directory - C:\Codebase\MyWebApp
+        /// Generated default save directory - C:\Codebase\MyWebAppCDK If the save directory already exists, then a suffix number is attached.
+        /// </summary>
+        /// <returns>The defaukt save directory path.</returns>
+        private string GenerateDefaultSaveDirectoryPath()
+        {
+            var applicatonDirectoryFullPath = _directoryManager.GetDirectoryInfo(_targetApplicationFullPath).Parent.FullName;
+            var saveCdkDirectoryFullPath = applicatonDirectoryFullPath + "CDK";
+
+            var suffixNumber = 0;
+            while (_directoryManager.Exists(saveCdkDirectoryFullPath))
+                saveCdkDirectoryFullPath = applicatonDirectoryFullPath + $"CDK{++suffixNumber}";
+            
+            return saveCdkDirectoryFullPath;
+        }
+
+        /// <summary>
+        /// This method takes a path and creates a new directory at this path if one does not already exist.
+        /// </summary>
+        /// <param name="saveCdkDirectoryPath">Relative or absolute path of the directory at which the CDK deployment project will be saved.</param>
+        /// <returns>A boolean to indicate if a new directory was created.</returns>
+        private bool CreateSaveCdkDirectory(string saveCdkDirectoryPath)
+        {
+            var newDirectoryCreated = false;
+            if (!_directoryManager.Exists(saveCdkDirectoryPath))
+            {
+                _directoryManager.CreateDirectory(saveCdkDirectoryPath);
+                newDirectoryCreated = true;
+            }
+            return newDirectoryCreated;   
+        }
+
+        /// <summary>
+        /// This method takes the path to the intended location of the CDK deployment project and performs validations on it.
+        /// </summary>
+        /// <param name="saveCdkDirectoryPath">Relative or absolute path of the directory at which the CDK deployment project will be saved.</param>
+        /// <returns>A tuple containaing a boolean that indicates if the directory is valid and a corresponding string error message.</returns>
+        private Tuple<bool, string> ValidateSaveCdkDirectory(string saveCdkDirectoryPath)
+        {
+            var errorMessage = string.Empty;
+            var isValid = true;
+            var targetApplicationDirectoryFullPath = _directoryManager.GetDirectoryInfo(_targetApplicationFullPath).Parent.FullName;
+           
+            if (!_directoryManager.IsEmpty(saveCdkDirectoryPath))
+            {
+                errorMessage += "The directory specified for saving the CDK project is non-empty. " +
+                    "Please provide an empty directory path and try again." + Environment.NewLine;
+
+                isValid = false;
+            }
+            if (_directoryManager.ExistsInsideDirectory(targetApplicationDirectoryFullPath, saveCdkDirectoryPath))
+            {
+                errorMessage += "The directory used to save the CDK deployment project is contained inside of " +
+                    "the target application project directory. Please specify a different directory and try again.";
+
+                isValid = false;
+            }
+
+            return new Tuple<bool, string>(isValid, errorMessage.Trim());
+        }
+
+        /// <summary>
+        /// Checks if the location of the saved CDK deployment project is monitored by a source control system.
+        /// </summary>
+        /// <param name="saveCdkDirectoryPath">Relative or absolute path of the directory at which the CDK deployment project will be saved.</param>
+        /// <returns></returns>
+        private async Task<bool> IsDirectoryUnderSourceControl(string saveCdkDirectoryPath)
+        {
+            var gitStatusResult = await _commandLineWrapper.TryRunWithResult("git status", saveCdkDirectoryPath);
+            var svnStatusResult = await _commandLineWrapper.TryRunWithResult("svn status", saveCdkDirectoryPath);
+            return gitStatusResult.Success || svnStatusResult.Success;
+        }
+
+        /// <summary>
+        /// Generates a snapshot of the deployment recipe inside the location at which the CDK deployment project is saved.
+        /// </summary>
+        /// <param name="recommendation"><see cref="Recommendation"/></param>
+        /// <param name="saveCdkDirectoryPath">Relative or absolute path of the directory at which the CDK deployment project will be saved.</param>
+        /// <param name="projectDisplayName">The name of the deployment project that will be displayed in the list of available deployment options.</param>
+        private async Task GenerateDeploymentRecipeSnapShot(Recommendation recommendation, string saveCdkDirectoryPath, string projectDisplayName)
+        {
+            var targetApplicationDirectoryName = _directoryManager.GetDirectoryInfo(_targetApplicationFullPath).Parent.Name;
+            var recipeSnapshotFileName = _directoryManager.GetDirectoryInfo(saveCdkDirectoryPath).Name + ".recipe";
+            var recipeSnapshotFilePath = Path.Combine(saveCdkDirectoryPath, recipeSnapshotFileName);
+            var recipePath = recommendation.Recipe.RecipePath;
+
+            if (string.IsNullOrEmpty(recipePath))
+                throw new InvalidOperationException("The recipe path cannot be null or empty as part " +
+                    $"of the {nameof(recommendation.Recipe)} object");
+
+            var recipeBody = await _fileManager.ReadAllTextAsync(recipePath);
+            var recipe = JsonConvert.DeserializeObject<RecipeDefinition>(recipeBody);
+            
+            var recipeName = string.IsNullOrEmpty(projectDisplayName) ?
+                $"Deployment project for {targetApplicationDirectoryName} to {recommendation.Recipe.TargetService}"
+                : projectDisplayName;
+
+            recipe.Id = Guid.NewGuid().ToString();
+            recipe.Name = recipeName;
+            recipe.CdkProjectTemplateId = null;
+            recipe.CdkProjectTemplate = null;
+
+            var recipeSnapshotBody = JsonConvert.SerializeObject(recipe, Formatting.Indented);
+            await _fileManager.WriteAllTextAsync(recipeSnapshotFilePath, recipeSnapshotBody);
+        }
+    }
+}
--- a/src/AWS.Deploy.Common/IO/DirectoryManager.cs
+++ b/src/AWS.Deploy.Common/IO/DirectoryManager.cs
@@ -1,6 +1,7 @@
 // Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
 // SPDX-License-Identifier: Apache-2.0
 
+using System;
 using System.IO;
 
 namespace AWS.Deploy.Common.IO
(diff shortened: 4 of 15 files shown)


src/AWS.Deploy.CLI/Commands/CommandFactory.cs
```diff
@@ -307,6 +309,72 @@ private Command BuildListCommand()
...
+            generateDeploymentProjectCommand.Handler = CommandHandler.Create(async (GenerateDeploymentProjectCommandHandlerInput input) =>
+            {
+                try
+                {
+                    _toolInteractiveService.Diagnostics = input.Diagnostics;
+                    var projectDefinition = await _projectParserUtility.Parse(input.ProjectPath ?? "");
+
+                    var saveDirectory = input.Output ?? "";
+
+                    OrchestratorSession session = new OrchestratorSession(projectDefinition, "", "");
```
ganeshnj, line 334: We should make Region and AccountId nullable rather passing empty string.

src/AWS.Deploy.CLI/Commands/CommandFactory.cs
```diff
@@ -307,6 +309,72 @@ private Command BuildListCommand()
...
+
+                    var generateDeploymentProject = new GenerateDeploymentProjectCommand(
+                        _toolInteractiveService,
+                        _consoleUtilities,
+                        _orchestratorInteractiveService,
+                        _cdkProjectHandler,
+                        _cdkManager,
+                        _awsResourceQueryer,
```
ganeshnj, line 342: We should break the Orchestrator. `IAWSResourceQueryer`, `ICDKManager`, `IDeploymentBundleHandler` are not required to generate the deployment project.

src/AWS.Deploy.CLI/Commands/GenerateDeploymentProjectCommand.cs
```diff
@@ -0,0 +1,185 @@
...
+            if (string.IsNullOrEmpty(saveDirectory))
+                saveDirectory = GenerateDefaultSaveDirectory(_session.ProjectDefinition);
+
+            var directoryCreatedByDeployTool = ValidateSaveDirectory(saveDirectory);
+
+            var result = await CheckDirectoryUnderSourceControl(saveDirectory);
+            if (!result.Success)
+            {
+                _toolInteractiveService.WriteLine();
+                var YesNoResult = _consoleUtilities.AskYesNoQuestion("The target directory for saving the CDK " +
```
ganeshnj, line 67: nit: local variable should start lowercase.

src/AWS.Deploy.CLI/Commands/GenerateDeploymentProjectCommand.cs
```diff
@@ -0,0 +1,185 @@
...
+            _orchestratorInteractiveService = orchestratorInteractiveService;
+            _cdkProjectHandler = cdkProjectHandler;
+            _cdkManager = cdkManager;
+            _awsResourceQueryer = awsResourceQueryer;
+            _deploymentBundleHandler = deploymentBundleHandler;
+            _commandLineWrapper = commandLineWrapper;
+            _session = session;
+        }
+
+        public async Task ExecuteAsync(string saveDirectory)
```
ganeshnj, line 52: Is this full path or just the directory name? Some documentation might help.

ganeshnj: How are we going to test this method? specially file IO given there is quite a bit of logic involved. You might want to use `IFileManager` and `IDirectoryManager` to make this command mockable.

src/AWS.Deploy.CLI/Commands/GenerateDeploymentProjectCommand.cs
```diff
@@ -0,0 +1,185 @@
...
+            }
+        }
+
+        private void CheckSaveDirectoryInsideProjectDirectory(string saveDirectory,
+            ProjectDefinition projectDefinition, bool directoryCreatedByDeployTool)
+        {
+            var saveDirectoryFullPath = new DirectoryInfo(saveDirectory).FullName;
+            var projectDirectoryFullPath = new DirectoryInfo(projectDefinition.ProjectPath).Parent.FullName;
+
+            if (saveDirectoryFullPath.Contains(projectDirectoryFullPath + "\\", StringComparison.InvariantCultureIgnoreCase))
```
ganeshnj, line 168: This is cross platform code, we should use `Path.Join` or `Path.DirectorySeparatorChar`.

src/AWS.Deploy.CLI/Commands/GenerateDeploymentProjectCommand.cs
```diff
@@ -0,0 +1,185 @@
...
+                if (directoryCreatedByDeployTool)
+                    Directory.Delete(saveDirectory);
+
+                var errorMessage = "The directory used to save the CDK deployment project is contained inside of " +
+                    "the target deployment project directory. Please specify a different directory and try again.";
+                throw new InvalidSaveDirectoryForCdkProject(errorMessage);
+            }
+        }
+
+        private async Task<TryRunResult> CheckDirectoryUnderSourceControl(string saveDirectory)
```
ganeshnj, line 179: nit: you aren't using result in the form it is returned. `IsDirectoryUnderSourceControl` fits better which returns a boolean.

How are we going to handle if customer is using a different source control system like SVN?

src/AWS.Deploy.CLI/Commands/GenerateDeploymentProjectCommand.cs
```diff
@@ -0,0 +1,185 @@
...
+using AWS.Deploy.Common;
+using AWS.Deploy.Orchestration;
+using AWS.Deploy.Orchestration.CDK;
+using AWS.Deploy.Orchestration.Data;
+using AWS.Deploy.Orchestration.Utilities;
+using AWS.Deploy.Recipes;
+
+namespace AWS.Deploy.CLI.Commands
+{
+    public class GenerateDeploymentProjectCommand
```
ganeshnj, line 18: How did you test this? we should write some test cases for this class.

ganeshnj (changes requested):

src/AWS.Deploy.CLI/Commands/GenerateDeploymentProjectCommand.cs
```diff
@@ -0,0 +1,185 @@
...
+            if (recommendations.Count == 0)
+            {
+                throw new FailedToGenerateAnyRecommendations("The project you are trying to deploy is currently not supported.");
+            }
+            return recommendations;
+        }
+
+        private string GenerateDefaultSaveDirectory(ProjectDefinition projectDefinition)
+        {
+            var projectDirectoryFullPath = new DirectoryInfo(projectDefinition.ProjectPath).Parent.FullName;
```
ganeshnj, line 109: `DirectoryInfo` constrictor can throw exception. We should guard against it.

src/AWS.Deploy.CLI/Commands/GenerateDeploymentProjectCommand.cs
```diff
@@ -0,0 +1,185 @@
...
+            {
+                var errorMessage = "Failed to create a directory at the specified path as it contains one or more invalid characters";
+                throw new InvalidSaveDirectoryForCdkProject(errorMessage, e);
+            }
+        }
+
+        private void CheckSaveDirectoryInsideProjectDirectory(string saveDirectory,
+            ProjectDefinition projectDefinition, bool directoryCreatedByDeployTool)
+        {
+            var saveDirectoryFullPath = new DirectoryInfo(saveDirectory).FullName;
```
ganeshnj, line 165: Same comment as L109

src/AWS.Deploy.CLI/Commands/CommandFactory.cs
96malhar: Fixed

src/AWS.Deploy.CLI/Commands/GenerateDeploymentProjectCommand.cs
96malhar: Changed the method name to `IsDirectoryUnderSourceControl` that returns a `bool`.

src/AWS.Deploy.CLI/Commands/GenerateDeploymentProjectCommand.cs
96malhar: Fixed

src/AWS.Deploy.CLI/Commands/CommandFactory.cs
96malhar: Added an `Orchestrator` constructor that only takes the `OrchestratorSession` and `RecipeDefinitionPaths`

src/AWS.Deploy.CLI/Commands/GenerateDeploymentProjectCommand.cs
96malhar: Fixed

src/AWS.Deploy.CLI/Commands/GenerateDeploymentProjectCommand.cs
96malhar: Added a wrapper method called `GetDirectoryInfo` that handles any exceptions thrown by the `DirectoryInfo` constructor.

src/AWS.Deploy.CLI/Commands/GenerateDeploymentProjectCommand.cs
96malhar: Added a wrapper method called GetDirectoryInfo that handles any exceptions thrown by the DirectoryInfo constructor.

src/AWS.Deploy.CLI/Commands/GenerateDeploymentProjectCommand.cs
96malhar: I have added integration tests for this command that performs and end-to-end test of all the file IO.

src/AWS.Deploy.CLI/Commands/GenerateDeploymentProjectCommand.cs
96malhar: Added Integration tests and verified that they run successfully on Gamma.

src/AWS.Deploy.CLI/Commands/GenerateDeploymentProjectCommand.cs
```diff
@@ -0,0 +1,237 @@
...
+                if (newDirectoryCreated)
+                    Directory.Delete(saveCdkDirectoryPath);
+                throw new InvalidSaveDirectoryForCdkProject(errorMessage);
+            }
+
+            var directoryUnderSourceControl = await IsDirectoryUnderSourceControl(saveCdkDirectoryPath);
+            if (!directoryUnderSourceControl)
+            {
+                _toolInteractiveService.WriteLine();
+                var yesNoResult = _consoleUtilities.AskYesNoQuestion("The target directory for saving the CDK " +
```
normj, line 70: If we are going to have this warning then we should explain why it is necessary. For example we want to explain that it is important that project is preserved if it is used for deployment to support redeployments. We should double check with Irene this experience.

src/AWS.Deploy.CLI/Commands/GenerateDeploymentProjectCommand.cs
```diff
@@ -0,0 +1,237 @@
...
+            }
+            return recommendations;
+        }
+
+        /// <summary>
+        /// This method takes the path to the target deployment project and creates
+        /// a default save directory inside the parent folder of the current directory.
+        /// For example:
+        /// Target project directory - C:\Codebase\MyWebApp\xyz.csproj
+        /// Generated default save directory - C:\Codebase\MyWebAppDeploymentProject If the save directory already exists, then a suffix number is attached.
```
normj, line 117: Lets use the default name format of `<app-project>.Deployment.csproj`. Periods help the name look more readable and a Project suffix in the name is redundant.

src/AWS.Deploy.CLI/Commands/GenerateDeploymentProjectCommand.cs
```diff
@@ -0,0 +1,237 @@
...
+        private string CheckSaveDirectoryInsideProjectDirectory(string saveCdkDirectoryPath, string projectPath)
+        {
+            var errorMessage = string.Empty;
+
+            var saveDirectoryFullPath = GetDirectoryInfo(saveCdkDirectoryPath).FullName;
+            var projectDirectoryFullPath = GetDirectoryInfo(projectPath).Parent.FullName;
+            if (saveDirectoryFullPath.Contains(projectDirectoryFullPath + Path.DirectorySeparatorChar, StringComparison.InvariantCultureIgnoreCase))
+            {
+                errorMessage = "The directory used to save the CDK deployment project is contained inside of " +
+                    "the target deployment project directory. Please specify a different directory and try again.";
```
normj, line 201: "target deployment project directory" -> "target application project directory"

src/AWS.Deploy.CLI/Commands/GenerateDeploymentProjectCommand.cs
```diff
@@ -0,0 +1,237 @@
...
+
+        /// <summary>
+        /// Checks if the location of the saved CDK deployment project is monitored by a source control system.
+        /// </summary>
+        /// <param name="saveCdkDirectoryPath">Relative or absolute path of the directory at which the CDK deployment project will be saved.</param>
+        /// <returns></returns>
+        private async Task<bool> IsDirectoryUnderSourceControl(string saveCdkDirectoryPath)
+        {
+            var gitStatusResult = await _commandLineWrapper.TryRunWithResult("git status", saveCdkDirectoryPath);
+            var svnStatusResult = await _commandLineWrapper.TryRunWithResult("svn status", saveCdkDirectoryPath);
```
normj, line 215: Wow, impressive with the svn support :)

src/AWS.Deploy.Orchestration/CdkProjectHandler.cs
```diff
@@ -61,18 +63,47 @@ public async Task CreateCdkDeployment(OrchestratorSession session, CloudApplicat
...
+        /// <summary>
+        /// Generates a snapshot of the deployment recipe inside the location at which the CDK deployment project is saved.
+        /// </summary>
+        /// <param name="recommendation"><see cref="Recommendation"/></param>
+        /// <param name="saveCdkDirectoryPath"></param>
+        private void GenerateDeploymentRecipeSnapShot(Recommendation recommendation, string saveCdkDirectoryPath)
+        {
+            var sourceFileName = recommendation.Recipe.RecipePath;
+            var recipeId = recommendation.Recipe.Id;
+            var destFileName = Path.Combine(saveCdkDirectoryPath, recipeId + "-" + Guid.NewGuid().ToString() + ".recipe");
```
normj, line 105: There is no reason to put a Guid in the file name and it makes it unreadable. I suggest use the same name as the project name except using `csproj` as the extension use `recipe`.

src/AWS.Deploy.Orchestration/CdkProjectHandler.cs
```diff
@@ -61,18 +63,47 @@ public async Task CreateCdkDeployment(OrchestratorSession session, CloudApplicat
...
+            File.Copy(sourceFileName, destFileName);
```
normj, line 106: In the recipe definition we need to replace the Id which is where the Guid can come from. We also need to add new fields indicating what recipe id and version this project was saved from.

normj (changes requested):

src/AWS.Deploy.CLI/Commands/GenerateDeploymentProjectCommand.cs
96malhar: Fixed

src/AWS.Deploy.CLI/Commands/GenerateDeploymentProjectCommand.cs
96malhar: Fixed

src/AWS.Deploy.Orchestration/CdkProjectHandler.cs
96malhar: Fixed

src/AWS.Deploy.Orchestration/CdkProjectHandler.cs
96malhar: Replaced the RecipeId with a Guid. Also added the ability for customers to provide their own recipe name (and also use a default name if one is not provided).

src/AWS.Deploy.CLI/Commands/GenerateDeploymentProjectCommand.cs
96malhar: Changed the warning to reflect the above.

src/AWS.Deploy.CLI/Commands/CommandFactory.cs
```diff
@@ -32,6 +33,8 @@ public class CommandFactory : ICommandFactory
         private static readonly Option<bool> _optionDiagnosticLogging = new(new []{"-d", "--diagnostics"}, "Enable diagnostic output.");
         private static readonly Option<string> _optionApply = new("--apply", "Path to the deployment settings file to be applied.");
         private static readonly Option<bool> _optionDisableInteractive = new(new []{"-s", "--silent" }, "Disable interactivity to deploy without any prompts for user input.");
+        private static readonly Option<string> _optionOutputDirectory = new(new[]{"-o", "--output"}, "Directory path in which the CDK deployment project will be saved.");
+        private static readonly Option<string> _optionRecipeName = new(new[] { "--recipe-name" }, "The custom name of the recipe that is saved as part of the recipe definition");
```
normj, line 37: Since the CLI command is deployment project and users don't know what recipes are from the context of the CLI command lets call this `--project-display-name`. The description can be: `The name of the deployment project that will be displayed in the list of available deployment options.`

src/AWS.Deploy.CLI/Commands/CommandFactory.cs
```diff
@@ -307,6 +311,80 @@ private Command BuildListCommand()
...
 
+        /// <summary>
+        /// Builds the top level command called "deployment-project" which supports the creation and saving on the
+        /// CDK deployment project.
+        /// </summary>
+        /// <returns>An instance of the <see cref="Command"/> class</returns>
+        private Command BuildDeploymentProjectCommand()
+        {
+            var deploymentProjectCommand = new Command("deployment-project",
+                "Save the CDK deployment project inside a user provided directory path.");
```
normj, line 322: Because this is more then just a CDK project lets change "CDK deployment project" to just "deployment project". I don't want users to think they can just run the cdk CLI on this project.

src/AWS.Deploy.CLI/Commands/CommandFactory.cs
```diff
@@ -307,6 +311,80 @@ private Command BuildListCommand()
...
+
+            var generateDeploymentProjectCommand = new Command("generate",
+                "Save the CDK deployment project inside a user provided directory path without proceeding with a deployment")
```
normj, line 325: Same comment as before about CDK.

src/AWS.Deploy.CLI/Commands/GenerateDeploymentProjectCommand.cs
```diff
@@ -0,0 +1,211 @@
...
+            {
+                if (newDirectoryCreated)
+                    _directoryManager.Delete(saveCdkDirectoryPath);
+                throw new InvalidSaveDirectoryForCdkProject(errorMessage);
+            }
+            
+            var directoryUnderSourceControl = await IsDirectoryUnderSourceControl(saveCdkDirectoryPath);
+            if (!directoryUnderSourceControl)
+            {
+                var userPrompt = "The target directory for saving the CDK " +
```
normj, line 81: My suggest:
`Warning: The target directory is not being tracked by source control. If the saved saved deployment project is used for deployment it is important the deployment project is retained to allow future redeployments to deployed applications.`

src/AWS.Deploy.CLI/Commands/GenerateDeploymentProjectCommand.cs
```diff
@@ -0,0 +1,211 @@
...
+        /// <param name="saveCdkDirectoryPath"></param>
+        /// <param name="customRecipeName">The custom name of the recipe that is saved as part of the recipe definition</param>
+        private async Task GenerateDeploymentRecipeSnapShot(Recommendation recommendation, string saveCdkDirectoryPath, string customRecipeName)
+        {
+            var targetApplicationDirectoryName = _directoryManager.GetDirectoryInfo(_targetApplicationFullPath).Parent.Name;
+            var recipeSnapshotFileName = _directoryManager.GetDirectoryInfo(saveCdkDirectoryPath).Name + ".recipe";
+            var recipeSnapshotFilePath = Path.Combine(saveCdkDirectoryPath, recipeSnapshotFileName);
+
+            var recipeName = string.IsNullOrEmpty(customRecipeName) ?
+                $"Deployment recipe for - {targetApplicationDirectoryName}" : customRecipeName;
```
normj, line 202: My suggestion for default name:
`$"Deployment project for {targetApplicationDirectoryName} to {recommendation.Recipe.TargetService}"`

src/AWS.Deploy.CLI/Commands/GenerateDeploymentProjectCommand.cs
```diff
@@ -0,0 +1,211 @@
...
+
+            recommendation.Recipe.Id = Guid.NewGuid().ToString();
+            recommendation.Recipe.Name = recipeName;
```
normj, line 205: We should remove CdkProjectTemplate and CdkProjectTemplateId from the recipe file since we will no longer be running the project template again.

src/AWS.Deploy.CLI/Commands/GenerateDeploymentProjectCommand.cs
```diff
@@ -0,0 +1,211 @@
...
+
+            var recipeBody = JsonConvert.SerializeObject(recommendation.Recipe, Formatting.Indented);
```
normj, line 207: I saw `RecipePath` is being serialized out. That should not be written out to disk. I think we should avoid serializing and deserializing the `recommendation.Recipe` property. Instead we should use the RecipePath we get from here and then load the recipe JSON file directly, manipulate the JSON file and then save it back to new location. There is no guarantees that the recommendation hasn't changed the state of the recipe like adding RecipePath.

normj (changes requested):

src/AWS.Deploy.CLI/Commands/CommandFactory.cs
96malhar: Fixed

src/AWS.Deploy.CLI/Commands/CommandFactory.cs
96malhar: Fixed

src/AWS.Deploy.CLI/Commands/CommandFactory.cs
96malhar: Fixed

src/AWS.Deploy.CLI/Commands/GenerateDeploymentProjectCommand.cs
96malhar: Fixed

src/AWS.Deploy.CLI/Commands/GenerateDeploymentProjectCommand.cs
96malhar: Fixed

src/AWS.Deploy.CLI/Commands/GenerateDeploymentProjectCommand.cs
96malhar: Fixed

src/AWS.Deploy.CLI/Commands/GenerateDeploymentProjectCommand.cs
96malhar: Fixed

src/AWS.Deploy.CLI/Commands/CommandHandlerInput/GenerateDeploymentProjectCommandHandlerInput.cs
```diff
@@ -0,0 +1,16 @@
...
+{
+    /// <summary>
+    /// This class maps the command line options for the "deployment-project generate" command to the appropriate C# properties.
+    /// </summary>
+    public class GenerateDeploymentProjectCommandHandlerInput
+    {
+        public string? ProjectPath { get; set; }
+        public bool Diagnostics { get; set; }
+        public string? Output { get; set; }
+        public string? ProjectDisplayName { get; set; }
```
ganeshnj, line 14: why do we need display name as param?

src/AWS.Deploy.CLI/Commands/GenerateDeploymentProjectCommand.cs
```diff
@@ -0,0 +1,218 @@
...
+        /// </summary>
+        /// <returns>The defaukt save directory path.</returns>
+        private string GenerateDefaultSaveDirectoryPath()
+        {
+            var applicatonDirectoryFullPath = _directoryManager.GetDirectoryInfo(_targetApplicationFullPath).Parent.FullName;
+            var saveCdkDirectoryFullPath = applicatonDirectoryFullPath + ".Deployment";
+
+            var suffixNumber = 0;
+            while (_directoryManager.Exists(saveCdkDirectoryFullPath))
+                saveCdkDirectoryFullPath = applicatonDirectoryFullPath + $".Deployment{++suffixNumber}";
```
ganeshnj, line 136: Can you confirm with @ik130 the naming?

src/AWS.Deploy.CLI/Commands/GenerateDeploymentProjectCommand.cs
```diff
@@ -0,0 +1,218 @@
...
+            }
+            return newDirectoryCreated;   
+        }
+
+        /// <summary>
+        /// This method takes the path to the intended location of the CDK deployment project and performs validations on it.
+        /// </summary>
+        /// <param name="saveCdkDirectoryPath">Relative or absolute path of the directory at which the CDK deployment project will be saved.</param>
+        /// <returns>A validation error message or an empty string if there are no errors.</returns>
+        private string ValidateSaveCdkDirectory(string saveCdkDirectoryPath)
```
ganeshnj, line 162: We shouldn't rely on the error message string for the state. I expect this method to return a boolean.

If you need message, you can also return a boolean and string tuple.

src/AWS.Deploy.CLI/Commands/GenerateDeploymentProjectCommand.cs
```diff
@@ -0,0 +1,218 @@
...
+        /// <param name="recommendation"><see cref="Recommendation"/></param>
+        /// <param name="saveCdkDirectoryPath">Relative or absolute path of the directory at which the CDK deployment project will be saved.</param>
+        /// <param name="projectDisplayName">The name of the deployment project that will be displayed in the list of available deployment options.</param>
+        private async Task GenerateDeploymentRecipeSnapShot(Recommendation recommendation, string saveCdkDirectoryPath, string projectDisplayName)
+        {
+            var targetApplicationDirectoryName = _directoryManager.GetDirectoryInfo(_targetApplicationFullPath).Parent.Name;
+            var recipeSnapshotFileName = _directoryManager.GetDirectoryInfo(saveCdkDirectoryPath).Name + ".recipe";
+            var recipeSnapshotFilePath = Path.Combine(saveCdkDirectoryPath, recipeSnapshotFileName);
+
+            var recipeBody = await _fileManager.ReadAllTextAsync(recommendation.Recipe.RecipePath ?? "");
```
ganeshnj, line 202: we must assert on `RecipePath`, if it is null then there is a bigger problem than passing an empty string path.

src/AWS.Deploy.CLI/Commands/GenerateDeploymentProjectCommand.cs
```diff
@@ -0,0 +1,218 @@
...
+            var recipe = JsonConvert.DeserializeObject<RecipeDefinition>(recipeBody);
+            
+            var recipeName = string.IsNullOrEmpty(projectDisplayName) ?
+                $"Deployment project for {targetApplicationDirectoryName} to {recommendation.Recipe.TargetService}"
+                : projectDisplayName;
+
+            recipe.Id = Guid.NewGuid().ToString();
+            recipe.Name = recipeName;
```
ganeshnj, line 210: Not related to this PR but we need to figure out how to handle the unique recipe id along with keeping the original recipe id.

src/AWS.Deploy.CLI/Commands/GenerateDeploymentProjectCommand.cs
```diff
@@ -0,0 +1,218 @@
...
+            recipe.CdkProjectTemplateId = string.Empty;
+            recipe.CdkProjectTemplate = string.Empty;
```
ganeshnj, line 212: should we consider making these nullable?

src/AWS.Deploy.Common/IO/DirectoryManager.cs
```diff
@@ -1,23 +1,94 @@
...
+        {
+            try
+            {
+                return Directory.CreateDirectory(path);
+            }
+            catch (Exception ex)
+            {
+                throw new FailedToCreateDirectoryException($"Failed to create a directory at path - {path}", ex);
+            } 
+        }
```
ganeshnj, line 33: File IO wrappers are for mocking only. We must not add any logic over the static methods.

We should capture the right exception in the caller.

src/AWS.Deploy.Common/IO/DirectoryManager.cs
```diff
@@ -1,23 +1,94 @@
...
+            }
+        }
+
+        public bool IsEmpty(string path) => GetFiles(path).Length == 0 && GetDirectories(path).Length == 0;
+
+        public bool ExistsInsideDirectory(string parentDirectoryPath, string childPath)
+        {
+            var parentDirectoryFullPath = GetDirectoryInfo(parentDirectoryPath).FullName;
+            var childFullPath = GetDirectoryInfo(childPath).FullName;
+            return childFullPath.Contains(parentDirectoryFullPath + Path.DirectorySeparatorChar, StringComparison.InvariantCulture);
```
ganeshnj, line 79: Can you add some unit tests for this?

src/AWS.Deploy.Orchestration/CdkAppSettingsSerializer.cs
```diff
@@ -23,8 +23,8 @@ public string Build(CloudApplication cloudApplication, Recommendation recommenda
                 projectPath,
                 recommendation.Recipe.Id,
                 recommendation.Recipe.Version,
-                session.AWSAccountId,
-                session.AWSRegion,
+                session.AWSAccountId ?? "",
+                session.AWSRegion ?? "",
```
ganeshnj, line 27: We should keep them nullable. Specially for region, logic can be different. We might want to fallback to default region depending on the requirement.

src/AWS.Deploy.Orchestration/Orchestrator.cs
```diff
@@ -73,6 +91,15 @@ public async Task<List<Recommendation>> GenerateDeploymentRecommendations()
...
         public async Task DeployRecommendation(CloudApplication cloudApplication, Recommendation recommendation)
         {
+            if (_interactiveService == null)
+                throw new InvalidOrchestratorException($"{nameof(_interactiveService)} is null as part of the orchestartor object");
+            if (_cdkManager == null)
+                throw new InvalidOrchestratorException($"{nameof(_cdkManager)} is null as part of the orchestartor object");
+            if (_cdkProjectHandler == null)
+                throw new InvalidOrchestratorException($"{nameof(_cdkProjectHandler)} is null as part of the orchestartor object");
+            if (_session == null)
+                throw new InvalidOrchestratorException($"{nameof(_session)} is null as part of the orchestartor object");
```
ganeshnj, line 101: Throw `InvalidOperationException`, we don't need custom exception. `InvalidOperationException` is purposefully created for invalid state.

https://docs.microsoft.com/en-us/dotnet/api/system.invalidoperationexception?view=net-5.0

src/AWS.Deploy.Orchestration/Orchestrator.cs
```diff
@@ -124,6 +151,13 @@ public DeploymentBundleDefinition GetDeploymentBundleDefinition(Recommendation r
 
         public async Task<bool> CreateContainerDeploymentBundle(CloudApplication cloudApplication, Recommendation recommendation)
         {
+            if (_interactiveService == null)
+                throw new InvalidOrchestratorException($"{nameof(_recipeDefinitionPaths)} is null as part of the orchestartor object");
+            if (_dockerEngine == null)
+                throw new InvalidOrchestratorException($"{nameof(_dockerEngine)} is null as part of the orchestartor object");
+            if (_deploymentBundleHandler == null)
+                throw new InvalidOrchestratorException($"{nameof(_deploymentBundleHandler)} is null as part of the orchestartor object");
```
ganeshnj, line 159: Same as above L101.

src/AWS.Deploy.Orchestration/TemplateEngine.cs
```diff
@@ -56,15 +56,17 @@ public async Task GenerateCDKProjectFromTemplate(Recommendation recommendation,
...
                 // CDK Template projects can parameterize the version number of the AWS.Deploy.Recipes.CDK.Common package. This avoid
                 // projects having to be modified every time the package version is bumped.
                 { "AWSDeployRecipesCDKCommonVersion", FileVersionInfo.GetVersionInfo(typeof(Constants.CloudFormationIdentifier).Assembly.Location).ProductVersion
                                                       ?? throw new InvalidAWSDeployRecipesCDKCommonVersionException("The version number of the AWS.Deploy.Recipes.CDK.Common package is invalid.") }
             };
 
+            if (!string.IsNullOrEmpty(session.AWSAccountId))
+                templateParameters["AWSAccountID"] = session.AWSAccountId;
+            if (!string.IsNullOrEmpty(session.AWSRegion))
+                templateParameters["AWSRegion"] = session.AWSRegion;
```
ganeshnj, line 68: do we use `AWSAccountID` and `AWSRegion` template param? I thought we removed them and they are part of configs now.

test/AWS.Deploy.CLI.IntegrationTests/SaveCdkDeploymentProject/SaveCdkDeploymentProjectTests.cs
```diff
@@ -0,0 +1,167 @@
...
+using AWS.Deploy.CLI.Extensions;
+using AWS.Deploy.CLI.IntegrationTests.Extensions;
+using AWS.Deploy.CLI.IntegrationTests.Services;
+using Microsoft.Extensions.DependencyInjection;
+using Xunit;
+using Task = System.Threading.Tasks.Task;
+
+namespace AWS.Deploy.CLI.IntegrationTests.SaveCdkDeploymentProject
+{
+    public class SaveCdkDeploymentProjectTests : IDisposable
```
ganeshnj, line 16: We are doing FileIO in tests which is discouraged. Can you use `IDirectoryManager` and `IFileManager` mocks? Our test cases must not rely on the FileIO. We already have test implementations for both. You can extend then based on your needs.

src/AWS.Deploy.CLI/Commands/CommandFactory.cs
```diff
@@ -32,6 +33,8 @@ public class CommandFactory : ICommandFactory
         private static readonly Option<bool> _optionDiagnosticLogging = new(new []{"-d", "--diagnostics"}, "Enable diagnostic output.");
         private static readonly Option<string> _optionApply = new("--apply", "Path to the deployment settings file to be applied.");
         private static readonly Option<bool> _optionDisableInteractive = new(new []{"-s", "--silent" }, "Disable interactivity to deploy without any prompts for user input.");
+        private static readonly Option<string> _optionOutputDirectory = new(new[]{"-o", "--output"}, "Directory path in which the CDK deployment project will be saved.");
+        private static readonly Option<string> _optionProjectDisplayName = new(new[] { "--project-display-name" }, "The name of the deployment project that will be displayed in the list of available deployment options.");
```
ganeshnj, line 37: Why do we need this?

ganeshnj (changes requested):

src/AWS.Deploy.CLI/Commands/CommandHandlerInput/GenerateDeploymentProjectCommandHandlerInput.cs
normj: I suggest we need a display name so when we save the CDK project we having something to put in the Name field of the recipe file. I want to avoid after saving a deployment project that we don't have 2 recommendations that look the same when we go back and deploy.

src/AWS.Deploy.CLI/Commands/GenerateDeploymentProjectCommand.cs
normj: I told Malhar we worry about keeping track of the original when we get to the stories about updating saved CDK projects. There is still the possibility we won't support updating and if that is the case having the original recipe id and version might not be needed.

Is there something wrong with using a Guid for the recipe ID of the new saved deployment project?

src/AWS.Deploy.CLI/Commands/CommandHandlerInput/GenerateDeploymentProjectCommandHandlerInput.cs
96malhar: I have included a CLI argument called `--project-display-name` which the customers can use to give a name to their deployment project. This name will be displayed when listing out different deployment recommendations

src/AWS.Deploy.CLI/Commands/GenerateDeploymentProjectCommand.cs
96malhar: Resolving this conversation as this comment is pertinent to a future story and not this one.

src/AWS.Deploy.CLI/Commands/CommandFactory.cs
96malhar: Refer to this comment - https://github.com/aws/aws-dotnet-deploy/pull/257#discussion_r671465771

src/AWS.Deploy.CLI/Commands/GenerateDeploymentProjectCommand.cs
96malhar: Fixed

src/AWS.Deploy.CLI/Commands/GenerateDeploymentProjectCommand.cs
96malhar: Fixed

src/AWS.Deploy.Orchestration/Orchestrator.cs
96malhar: Fixed

src/AWS.Deploy.Orchestration/TemplateEngine.cs
96malhar: That is correct. It was handled as part of this PR - https://github.com/aws/aws-dotnet-deploy/pull/250

src/AWS.Deploy.CLI/Commands/GenerateDeploymentProjectCommand.cs
96malhar: Fixed

src/AWS.Deploy.CLI/Commands/GenerateDeploymentProjectCommand.cs
96malhar: Changed the suffix to `CDK` instead of `.Deployment`. This change came from @ik130.

test/AWS.Deploy.CLI.IntegrationTests/SaveCdkDeploymentProject/SaveCdkDeploymentProjectTests.cs
96malhar: The `IDirectoryManager` and `IFileManager` mocks are used to **bypass** any File I/O that is encountered while testing a larger scheme of things. In this case File I/O itself is the core component that we need to test. Trying to write unit tests which fakes the behavior of File I/O does not achieve our purpose and that is why I have included integration tests to perform end-to-end testing. I also spoke about this with @normj and he concurs.

src/AWS.Deploy.Common/IO/DirectoryManager.cs
96malhar: There is different File I/O wrapper called `TestDirectoryManager` and `TestFileManager` used for mocking. Adding the exception in the wrapper class saves the effort into wrapping **every** call around a try-catch block.

src/AWS.Deploy.Orchestration/CdkAppSettingsSerializer.cs
96malhar: Made them nullable.

src/AWS.Deploy.Common/IO/DirectoryManager.cs
ganeshnj: > File IO wrappers are for mocking only.

I wasn't clear enough, for mocking I meant, converting static methods to instance methods by wrapping them in `*Manager` class. 

Depending on the use case, we should use better messaging than just "Failed to do something". This information is already supplied by the exceptions thrown in FileIO methods, [one example](https://docs.microsoft.com/en-us/dotnet/api/system.io.directory.createdirectory?view=net-5.0).

I expect, if generate command fails, we must write something

```
Failed to generate deployment project

UnAuthorizedAccessException: Unable to access file. The file is read-only. // This is one example
```

src/AWS.Deploy.CLI/Commands/GenerateDeploymentProjectCommand.cs
96malhar: Made them nullable

src/AWS.Deploy.Common/IO/DirectoryManager.cs
96malhar: The invocation of the generate command is wrapped around a try-catch block inside `CommandFactory.cs`. If the diagnostics mode is turned on then the entire stack trace is printed. But without it only the top-level exception is printed.

The custom exceptions are also catching any exception thrown by System.IO and storing them as inner exceptions.

src/AWS.Deploy.Common/IO/DirectoryManager.cs
96malhar: Refer to https://github.com/aws/aws-dotnet-deploy/pull/257#discussion_r671526419

normj (approved):

ganeshnj (approved):

PR #258, opened 2021-07-09
feat: Add logging messages to show what docker build command and from what directory the build will happen

normj:
*Description of changes:*
While recently setting up a CodeBuild jobs I was running into a lot of Docker build issues. It would have greatly helped my debugging effort if I knew for sure what exactly was the Docker execution directory was in CodeBuild. Added logging messages to better help debugging efforts in the future.

### DIFF
--- a/src/AWS.Deploy.Orchestration/DeploymentBundleHandler.cs
+++ b/src/AWS.Deploy.Orchestration/DeploymentBundleHandler.cs
@@ -56,6 +56,9 @@ namespace AWS.Deploy.Orchestration
             var buildArgs = GetDockerBuildArgs(recommendation);
 
             var dockerBuildCommand = $"docker build -t {imageTag} -f \"{dockerFile}\"{buildArgs} .";
+            _interactiveService.LogMessageLine($"Docker Execution Directory: {Path.GetFullPath(dockerExecutionDirectory)}");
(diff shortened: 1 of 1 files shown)


src/AWS.Deploy.Orchestration/DeploymentBundleHandler.cs
```diff
@@ -56,6 +56,9 @@ public async Task<string> BuildDockerImage(CloudApplication cloudApplication, Re
             var buildArgs = GetDockerBuildArgs(recommendation);
 
             var dockerBuildCommand = $"docker build -t {imageTag} -f \"{dockerFile}\"{buildArgs} .";
+            _interactiveService.LogMessageLine($"Docker Execution Directory: {Path.GetFullPath(dockerExecutionDirectory)}");
+            _interactiveService.LogMessageLine($"Docker Build Command: {dockerBuildCommand}");
```
philasmar, line 60: Can we use LogDebugLine instead? I'd rather we only show this info when we are in diagnostics mode.

normj: I would rather keep it as a regular logging message. I have found this information in the output really helpful when debugging issues in our other CLIs. Users who have problems often put this output as part of the issue.

philasmar (approved):

ganeshnj (approved):

PR #259, opened 2021-07-12
fix: server mode not shutting down when parent process exits

philasmar:
*Issue #, if available:*
IDE-4723

*Description of changes:*
fix: server mode not shutting down when parent process exits

### DIFF
--- a/src/AWS.Deploy.CLI/Commands/ServerModeCommand.cs
+++ b/src/AWS.Deploy.CLI/Commands/ServerModeCommand.cs
@@ -59,31 +59,27 @@ namespace AWS.Deploy.CLI.Commands
             }
             else
             {
-                var monitorTask = WaitOnParentPid(cancellationToken);
-                var webTask = host.RunAsync(cancellationToken);
-
-                //This call will wait until one of the tasks completes.
-                //The monitor task will complete if a parent pid is not found.
-                await Task.WhenAny(monitorTask, webTask);
-
-                //The monitor task is expected to complete only when a parent pid is not found.
-                if (monitorTask.IsCompleted && monitorTask.Result)
+                try
                 {
-                    _interactiveService.WriteLine(string.Empty);
-                    _interactiveService.WriteLine("The parent process is no longer running.");
-                    _interactiveService.WriteLine("Server mode is shutting down...");
-                    await host.StopAsync(cancellationToken);
+                    var process = Process.GetProcessById((int)_parentPid);
+                    process.EnableRaisingEvents = true;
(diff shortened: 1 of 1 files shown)


src/AWS.Deploy.CLI/Commands/ServerModeCommand.cs
```diff
@@ -136,7 +136,9 @@ private async Task<bool> WaitOnParentPid(CancellationToken token)
             {
                 try
                 {
-                    Process.GetProcessById((int)_parentPid);
+                    var process = Process.GetProcessById((int)_parentPid);
+                    if (process.HasExited)
+                        return true;
                     await Task.Delay(1000, token);
                 }
                 catch
```
ganeshnj, line 144: Can you remove the infinite while and use `WaitForExit`?

https://docs.microsoft.com/en-us/dotnet/api/system.diagnostics.process.waitforexit?view=net-5.0#system-diagnostics-process-waitforexit(system-int32)

ganeshnj (changes requested):

src/AWS.Deploy.CLI/Commands/ServerModeCommand.cs
philasmar: WaitForExit is a blocking call. I reworked it to use events to remove the while loop.

src/AWS.Deploy.CLI/Commands/ServerModeCommand.cs
```diff
@@ -59,33 +59,29 @@ public async Task ExecuteAsync(CancellationToken cancellationToken = default(Can
...
-                if (webTask.IsCompleted && webTask.IsFaulted)
+                catch (Exception)
                 {
-                    var innerException = webTask.Exception?.InnerException;
-                    if (innerException != null)
-                        throw innerException;
+                    return;
                 }
+
+                await host.RunAsync(cancellationToken);
```
ganeshnj, line 73: It appears like a regression, previously when the host task (web app) exited, we were throwing an exception which is no longer true now.

I think `ServerModeSession` will kick in and re-start the server but we are losing the diagnostics here IMO.

philasmar: Before we were using Task.WhenAny which causes an AggregateException to be thrown. We were catching that and throwing back the inner exception. Now we don't have that anymore and await host.RunAsync(cancellationToken) will just throw an exception same as before. The same behavior was retained.

ganeshnj (approved):

96malhar (approved):

normj (approved):

PR #264, opened 2021-07-15
feat: Create a deployment-manifest file

96malhar:
*Issue #, if available:*
DOTNET-5236

*Description of changes:*
This PR creates a deployment-manifest file to keep track of the saved CDK deployment projects.

### DIFF
--- a/src/AWS.Deploy.CLI/Commands/CommandFactory.cs
+++ b/src/AWS.Deploy.CLI/Commands/CommandFactory.cs
@@ -16,6 +16,7 @@ using AWS.Deploy.Orchestration.Data;
 using AWS.Deploy.Orchestration.Utilities;
 using AWS.Deploy.CLI.Commands.CommandHandlerInput;
 using AWS.Deploy.Common.IO;
+using AWS.Deploy.Common.DeploymentManifest;
 
 namespace AWS.Deploy.CLI.Commands
 {
@@ -52,6 +53,7 @@ namespace AWS.Deploy.CLI.Commands
         private readonly IDeployedApplicationQueryer _deployedApplicationQueryer;
         private readonly ITypeHintCommandFactory _typeHintCommandFactory;
         private readonly IConsoleUtilities _consoleUtilities;
+        private readonly IDeploymentManifestEngine _deploymentManifestEngine;
 
         public CommandFactory(
             IToolInteractiveService toolInteractiveService,
@@ -69,7 +71,8 @@ namespace AWS.Deploy.CLI.Commands
             ITemplateMetadataReader templateMetadataReader,
             IDeployedApplicationQueryer deployedApplicationQueryer,
             ITypeHintCommandFactory typeHintCommandFactory,
-            IConsoleUtilities consoleUtilities)
+            IConsoleUtilities consoleUtilities,
+            IDeploymentManifestEngine deploymentManifestEngine)
         {
             _toolInteractiveService = toolInteractiveService;
             _orchestratorInteractiveService = orchestratorInteractiveService;
@@ -87,6 +90,7 @@ namespace AWS.Deploy.CLI.Commands
             _deployedApplicationQueryer = deployedApplicationQueryer;
             _typeHintCommandFactory = typeHintCommandFactory;
             _consoleUtilities = consoleUtilities;
+            _deploymentManifestEngine = deploymentManifestEngine;
         }
 
         public Command BuildRootCommand()
@@ -352,6 +356,7 @@ namespace AWS.Deploy.CLI.Commands
                         new DirectoryManager(),
                         new FileManager(),
                         session,
+                        _deploymentManifestEngine,
                         targetApplicationFullPath);
 
                     await generateDeploymentProject.ExecuteAsync(saveDirectory, projectDisplayName);
@@ -360,7 +365,6 @@ namespace AWS.Deploy.CLI.Commands
                 }
                 catch (Exception e) when (e.IsAWSDeploymentExpectedException())
                 {
-                    _toolInteractiveService.WriteErrorLine("Failed to generate deployment project.");
                     if (input.Diagnostics)
                         _toolInteractiveService.WriteErrorLine(e.PrettyPrint());
                     else
--- a/src/AWS.Deploy.CLI/Commands/GenerateDeploymentProjectCommand.cs
+++ b/src/AWS.Deploy.CLI/Commands/GenerateDeploymentProjectCommand.cs
@@ -6,6 +6,7 @@ using System.Collections.Generic;
 using System.IO;
 using System.Threading.Tasks;
 using AWS.Deploy.Common;
+using AWS.Deploy.Common.DeploymentManifest;
 using AWS.Deploy.Common.IO;
 using AWS.Deploy.Common.Recipes;
 using AWS.Deploy.Orchestration;
@@ -28,6 +29,7 @@ namespace AWS.Deploy.CLI.Commands
         private readonly IDirectoryManager _directoryManager;
         private readonly IFileManager _fileManager;
         private readonly OrchestratorSession _session;
+        private readonly IDeploymentManifestEngine _deploymentManifestEngine;
         private readonly string _targetApplicationFullPath;
         
         public GenerateDeploymentProjectCommand(
@@ -38,6 +40,7 @@ namespace AWS.Deploy.CLI.Commands
             IDirectoryManager directoryManager,
             IFileManager fileManager,
             OrchestratorSession session,
+            IDeploymentManifestEngine deploymentManifestEngine,
             string targetApplicationFullPath)
         {
             _toolInteractiveService = toolInteractiveService;
@@ -47,6 +50,7 @@ namespace AWS.Deploy.CLI.Commands
             _directoryManager = directoryManager;
             _fileManager = fileManager;
             _session = session;
+            _deploymentManifestEngine = deploymentManifestEngine;
             _targetApplicationFullPath = targetApplicationFullPath;
         }
 
@@ -73,7 +77,8 @@ namespace AWS.Deploy.CLI.Commands
             {
                 if (newDirectoryCreated)
                     _directoryManager.Delete(saveCdkDirectoryPath);
-                throw new InvalidSaveDirectoryForCdkProject(errorMessage);
+                errorMessage = $"Failed to generate deployment project.{Environment.NewLine}{errorMessage}";
+                throw new InvalidSaveDirectoryForCdkProject(errorMessage.Trim());
             }
             
             var directoryUnderSourceControl = await IsDirectoryUnderSourceControl(saveCdkDirectoryPath);
@@ -98,9 +103,11 @@ namespace AWS.Deploy.CLI.Commands
             await _cdkProjectHandler.CreateCdkProjectForDeployment(selectedRecommendation, _session, saveCdkDirectoryPath);
             await GenerateDeploymentRecipeSnapShot(selectedRecommendation, saveCdkDirectoryPath, projectDisplayName);
 
-            var saveCdkDirectoryInfo = _directoryManager.GetDirectoryInfo(saveCdkDirectoryPath);
+            var saveCdkDirectoryFullPath = _directoryManager.GetDirectoryInfo(saveCdkDirectoryPath).FullName;
             _toolInteractiveService.WriteLine();
-            _toolInteractiveService.WriteLine($"The CDK deployment project is saved at: {saveCdkDirectoryInfo.FullName}");
+            _toolInteractiveService.WriteLine($"The CDK deployment project is saved at: {saveCdkDirectoryFullPath}");
+
+            await _deploymentManifestEngine.UpdateDeploymentManifestFile(saveCdkDirectoryFullPath, _targetApplicationFullPath);
         }
 
         /// <summary>
--- a/src/AWS.Deploy.Common/DeploymentManifest/DeploymentManifestEngine.cs
+++ b/src/AWS.Deploy.Common/DeploymentManifest/DeploymentManifestEngine.cs
@@ -1,0 +1,107 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+using System;
+using System.Collections.Generic;
+using System.IO;
+using System.Threading.Tasks;
+using AWS.Deploy.Common.IO;
+using Newtonsoft.Json;
+
+namespace AWS.Deploy.Common.DeploymentManifest
+{
+    public interface IDeploymentManifestEngine
+    {
+        Task UpdateDeploymentManifestFile(string saveCdkDirectoryFullPath, string targetApplicationFullPath);
+    }
+
+    /// <summary>
+    /// This class contains the helper methods to update the deployment manifest file
+    /// that keeps track of the save CDK deployment projects.
+    /// </summary>
+    public class DeploymentManifestEngine : IDeploymentManifestEngine
+    {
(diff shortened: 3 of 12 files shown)


src/AWS.Deploy.CLI/Commands/GenerateDeploymentProjectCommand.cs
```diff
@@ -98,9 +100,11 @@ public async Task ExecuteAsync(string saveCdkDirectoryPath, string projectDispla
             await _cdkProjectHandler.CreateCdkProjectForDeployment(selectedRecommendation, _session, saveCdkDirectoryPath);
             await GenerateDeploymentRecipeSnapShot(selectedRecommendation, saveCdkDirectoryPath, projectDisplayName);
 
-            var saveCdkDirectoryInfo = _directoryManager.GetDirectoryInfo(saveCdkDirectoryPath);
+            var saveCdkDirectoryFullPath = _directoryManager.GetDirectoryInfo(saveCdkDirectoryPath).FullName;
```
ganeshnj, line 103: this seems not required, you can keep the DirectoryInfo object and pass the required property.

src/AWS.Deploy.CLI/Commands/GenerateDeploymentProjectCommand.cs
```diff
@@ -227,5 +231,24 @@ private async Task GenerateDeploymentRecipeSnapShot(Recommendation recommendatio
...
             await _fileManager.WriteAllTextAsync(recipeSnapshotFilePath, recipeSnapshotBody);
         }
+
+        /// <summary>
+        /// This method takes the full path of the directory at which the CDK deployment project is saved and adds a
+        /// corresponding entry in the deployment-manifest file.
+        /// </summary>
+        /// <param name="saveCdkDirectoryFullPath">The full path of the directory at which the CDK deployment project is saved</param>
+        /// <returns></returns>
+        private async Task UpdateDeploymentManifestFile(string saveCdkDirectoryFullPath)
```
ganeshnj, line 241: nit: I would prefer using AbosolutePath and RelativePath terms.

src/AWS.Deploy.Common/DeploymentManifest/DeploymentManifestEngine.cs
```diff
@@ -0,0 +1,91 @@
...
+    {
+        private readonly IDirectoryManager _directoryManager;
+        private readonly IFileManager _fileManager;
+        private readonly string _targetApplicationFullPath;
+        private readonly string _saveCdkDirectoryFullPath;
+
+        public DeploymentManifestEngine(string targetApplicationFullPath, string saveCdkDirectoryFullPath)
+        {
+            _directoryManager = new DirectoryManager();
+            _fileManager = new FileManager();
```
ganeshnj, line 27: Don't initialize in the constructor, use them as constructor dependencies. Both `IFileManager` and `IDirectoryManager` are available in DI 

https://github.com/aws/aws-dotnet-deploy/blob/main/src/AWS.Deploy.CLI/Extensions/CustomServiceCollectionExtension.cs#L40

src/AWS.Deploy.Common/DeploymentManifest/DeploymentManifestEngine.cs
```diff
@@ -0,0 +1,91 @@
...
+    /// that keeps track of the save CDK deployment projects.
+    /// </summary>
+    public class DeploymentManifestEngine
+    {
+        private readonly IDirectoryManager _directoryManager;
+        private readonly IFileManager _fileManager;
+        private readonly string _targetApplicationFullPath;
+        private readonly string _saveCdkDirectoryFullPath;
+
+        public DeploymentManifestEngine(string targetApplicationFullPath, string saveCdkDirectoryFullPath)
```
ganeshnj, line 24: I expect both `targetApplicationFullPath` and `saveCdkDirectoryFullPath` as parameters to public methods. Also add DeploymentManifestEngine to dependencies [here](https://github.com/aws/aws-dotnet-deploy/blob/main/src/AWS.Deploy.CLI/Extensions/CustomServiceCollectionExtension.cs#L40) and use constructor DI to obtain an instance of `IDeploymentManifestEngine`.

ganeshnj (changes requested):

src/AWS.Deploy.CLI/Commands/GenerateDeploymentProjectCommand.cs
96malhar: I only need the full directory path and not the full `DirectoryInfo` object. Hence only capturing the necessary information.

src/AWS.Deploy.CLI/Commands/GenerateDeploymentProjectCommand.cs
96malhar: I have used `xyzFullPath` for **all** variables that refer to the absolute path. I can put a a separate PR to fix this.

src/AWS.Deploy.Common/DeploymentManifest/DeploymentManifestEngine.cs
96malhar: Fixed

src/AWS.Deploy.Common/DeploymentManifest/DeploymentManifestEngine.cs
96malhar: Fixed

src/AWS.Deploy.Common/DeploymentManifest/DeploymentManifestEngine.cs
```diff
@@ -0,0 +1,96 @@
...
+    /// <summary>
+    /// This class contains the helper methods to update the deployment manifest file
+    /// that keeps track of the save CDK deployment projects.
+    /// </summary>
+    public class DeploymentManifestEngine : IDeploymentManifestEngine
+    {
+        private readonly IDirectoryManager _directoryManager;
+        private readonly IFileManager _fileManager;
+
+        public string TargetApplicationFullPath { get; set; } = string.Empty;
```
ganeshnj, line 29: I imagine you are assigning an empty string to get rid of error for initialization during construction. This should be a nullable string and must be asserted on null before it is being used in the methods. It allows avoiding developer error in case one forgets to set to a correct value.

96malhar: Fixed

src/AWS.Deploy.Common/DeploymentManifest/DeploymentManifestEngine.cs
```diff
@@ -0,0 +1,105 @@
...
+    /// <summary>
+    /// This class contains the helper methods to update the deployment manifest file
+    /// that keeps track of the save CDK deployment projects.
+    /// </summary>
+    public class DeploymentManifestEngine : IDeploymentManifestEngine
+    {
+        private readonly IDirectoryManager _directoryManager;
+        private readonly IFileManager _fileManager;
+
+        public string? TargetApplicationFullPath { get; set; }
```
philasmar, line 29: I don't see why this is a public variable. This would be better passed as a parameter to UpdateDeploymentManifestFile. That way, you don't need to define it as a nullable string and set it after initializing the deployment manifest engine.

src/AWS.Deploy.CLI/Commands/CommandFactory.cs
```diff
@@ -343,6 +347,7 @@ private Command BuildDeploymentProjectCommand()
                     OrchestratorSession session = new OrchestratorSession(projectDefinition);
 
                     var targetApplicationFullPath = new DirectoryManager().GetDirectoryInfo(projectDefinition.ProjectPath).FullName;
+                    _deploymentManifestEngine.TargetApplicationFullPath = targetApplicationFullPath;
```
philasmar, line 350: I don't like that we are setting this here. It should either be part of the constructor or passed to the function that needs it. (Following up on a later comment)

src/AWS.Deploy.CLI/Commands/GenerateDeploymentProjectCommand.cs
```diff
@@ -73,7 +77,8 @@ public async Task ExecuteAsync(string saveCdkDirectoryPath, string projectDispla
             {
                 if (newDirectoryCreated)
                     _directoryManager.Delete(saveCdkDirectoryPath);
-                throw new InvalidSaveDirectoryForCdkProject(errorMessage);
+                errorMessage = "Failed to generate deployment project." + Environment.NewLine + errorMessage;
```
philasmar, line 80: Use string interpolation instead of concatenating strings.

src/AWS.Deploy.Common/DeploymentManifest/DeploymentManifestEngine.cs
```diff
@@ -0,0 +1,105 @@
...
+using System.IO;
+using System.Threading.Tasks;
+using AWS.Deploy.Common.IO;
+using Newtonsoft.Json;
+
+namespace AWS.Deploy.Common.DeploymentManifest
+{
+    public interface IDeploymentManifestEngine
+    {
+        string? TargetApplicationFullPath { get; set; }
```
philasmar, line 15: This shouldn't be here (Same comment as below)

src/AWS.Deploy.Common/DeploymentManifest/DeploymentManifestEngine.cs
```diff
@@ -0,0 +1,105 @@
...
+                deploymentManifestModel.DeploymentManifestEntries.Add(new DeploymentManifestEntry(saveCdkDirectoryRelativePath));
+            }
+            else
+            {
+                var deploymentManifestEntries = new List<DeploymentManifestEntry> { new DeploymentManifestEntry(saveCdkDirectoryRelativePath) };
+                deploymentManifestModel = new DeploymentManifestModel(deploymentManifestEntries);
+            }
+
+            var manifestFileJsonString = SerializeManifestModel(deploymentManifestModel);
+            await _fileManager.WriteAllTextAsync(deploymentManifestFilePath, manifestFileJsonString);
```
philasmar, line 66: There's a lot of file IO happening here. We should capture exceptions thrown here and throw back a meaningful expected exception.

src/AWS.Deploy.Common/DeploymentManifest/DeploymentManifestEngine.cs
```diff
@@ -0,0 +1,105 @@
...
+        /// </summary>
+        /// <returns>The path to the deployment-manifest file.</returns>
+        private string GetDeploymentManifestFilePath()
+        {
+            if (string.IsNullOrEmpty(TargetApplicationFullPath))
+                throw new InvalidOperationException($"{nameof(TargetApplicationFullPath)} is null as part of the " +
+                    $"{nameof(DeploymentManifestEngine)} object");
+
+            var projectDirectoryFullPath = _directoryManager.GetDirectoryInfo(TargetApplicationFullPath).Parent.FullName;
+            var deploymentManifestFileFullPath = Path.Combine(projectDirectoryFullPath, "aws-deployments.json");
```
philasmar, line 101: define the file name as a private constant and refer to it here.

src/AWS.Deploy.Common/Exceptions.cs
```diff
@@ -159,6 +159,15 @@ public class InvalidUserDeploymentSettingsException : Exception
         public InvalidUserDeploymentSettingsException(string message, Exception? innerException = null) : base(message, innerException) { }
     }
 
+    /// <summary>
+    /// Exception throw if a failure occured while trying to update the deployment manifest file.
```
philasmar, line 163: typy here. change "throw" to "thrown"

philasmar (changes requested):
Can you please add some tests for the DeploymentManifestEngine?

src/AWS.Deploy.Common/DeploymentManifest/DeploymentManifestEngine.cs
96malhar: Fixed. Passing it as parameter to methods which need this.

src/AWS.Deploy.CLI/Commands/CommandFactory.cs
96malhar: Fixed. Passing it as parameter to methods which need this.

src/AWS.Deploy.CLI/Commands/GenerateDeploymentProjectCommand.cs
96malhar: Fixed

src/AWS.Deploy.Common/DeploymentManifest/DeploymentManifestEngine.cs
96malhar: Fixed

src/AWS.Deploy.Common/DeploymentManifest/DeploymentManifestEngine.cs
96malhar: Fixed

src/AWS.Deploy.Common/Exceptions.cs
96malhar: Fixed

96malhar:
> Can you please add some tests for the DeploymentManifestEngine?

Added integration tests.

src/AWS.Deploy.Common/DeploymentManifest/DeploymentManifestEngine.cs
96malhar: The `DeploymentManifestEngine` is just a wrapper class that supports the serialization and deserialization of the manifest file. Any exception thrown by this class should be caught inside the caller which calls the public methods of this class.

philasmar: The DeploymentManifestEngine has the sole responsibility of updating the manifest file. If it cannot do that job, it needs to throw a specific exception that can be caught by the caller. Either way, you are not catching any exceptions in the caller and any unexpected exception will bubble up to the user. I still think the DeploymentManifestEngine should be throwing an expected exception and that should be documented so that the caller knows that this method throws an exception. RIght now, the caller does not know whether the DeploymentManifestEngine throws exceptions. We need to separate responsibilities and we should not expect the caller to know more than it should.

philasmar (approved):

test/AWS.Deploy.CLI.Common.UnitTests/DeploymentManifestFile/DeploymentManifestFileTests.cs
```diff
@@ -0,0 +1,117 @@
...
+        private bool _isDisposed;
+
+        public DeploymentManifestFileTests()
+        {
+            _fileManager = new FileManager();
+            _directoryManager = new DirectoryManager();
+            var targetApplicationPath = Path.Combine("testapps", "WebAppWithDockerFile", "WebAppWithDockerFile.csproj");
+            _targetApplicationFullPath = _directoryManager.GetDirectoryInfo(targetApplicationPath).FullName;
+            _targetApplicationDirectoryFullPath = _directoryManager.GetDirectoryInfo(targetApplicationPath).Parent.FullName;
+            _deploymentManifestEngine = new DeploymentManifestEngine(new DirectoryManager(), new FileManager());
```
ganeshnj, line 33:
```
_deploymentManifestEngine = new DeploymentManifestEngine(_directoryManager, _fileManager);
```

96malhar: Fixed

ganeshnj (approved):

PR #265, opened 2021-07-20
fix: previous settings are applied to global SelectedRecommendation in state object

philasmar:
*Issue #, if available:*

*Description of changes:*
fix: previous settings are applied to global SelectedRecommendation in state object

### DIFF
--- a/src/AWS.Deploy.CLI/Commands/DeployCommand.cs
+++ b/src/AWS.Deploy.CLI/Commands/DeployCommand.cs
@@ -215,7 +215,7 @@ namespace AWS.Deploy.CLI.Commands
                         " does not match the original recipe used to deploy the application stack.");
             }
 
-            selectedRecommendation.ApplyPreviousSettings(existingCloudApplicationMetadata.Settings);
+            selectedRecommendation = selectedRecommendation.ApplyPreviousSettings(existingCloudApplicationMetadata.Settings);
 
             var header = $"Loading {deployedApplication.Name} settings:";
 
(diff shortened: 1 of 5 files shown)


src/AWS.Deploy.CLI/Commands/GenerateDeploymentProjectCommand.cs
```diff
@@ -221,8 +221,8 @@ private async Task GenerateDeploymentRecipeSnapShot(Recommendation recommendatio
 
             recipe.Id = Guid.NewGuid().ToString();
             recipe.Name = recipeName;
-            recipe.CdkProjectTemplateId = null;
-            recipe.CdkProjectTemplate = null;
+            recipe.CdkProjectTemplateId = string.Empty;
+            recipe.CdkProjectTemplate = string.Empty;
```
ganeshnj, line 225: why do we need empty string? This will have a an affect on JSON serialization and deserialization.

philasmar: They were being explicitly set to null before. In order to preserve the same functionality, I set them to an empty string since I need them to be non nullable. The reason for that is Newtonsoft throws an exception when deserializing the Recipe object when they were set to nullable.

ganeshnj: We should use `NullValueHandling = NullValueHandling.Ignore` option.

https://www.newtonsoft.com/json/help/html/NullValueHandlingIgnore.htm

philasmar: For some reason, i retested and I didn't run into the error so I reverted the changes related to nullability.

ganeshnj (approved):

96malhar (approved):

PR #266, opened 2021-07-22
fix: Correctly resolve target project path

96malhar:
*Description of changes:*
Currently, in the deploy tool, if a **relative path** to the target application project directory is provided it results in a `ProjectFileNotFoundException`.

Example scenario:
* Current working directory - `C:\codebase\aws-dotnet-deploy`
* CLI command - `dotnet aws deploy --project-path ".\testapps\WebAppWithDockerFile"`
* Result - `ProjectFileNotFoundException`

This PR attempts to correct this behavior.

### DIFF
--- a/src/AWS.Deploy.Common/ProjectDefinitionParser.cs
+++ b/src/AWS.Deploy.Common/ProjectDefinitionParser.cs
@@ -33,10 +33,18 @@ namespace AWS.Deploy.Common
             _directoryManager = directoryManager;
         }
 
+        /// <summary>
+        /// This method parses the target application project and sets the
+        /// appropriate metadata as part of the <see cref="ProjectDefinition"/>
+        /// </summary>
+        /// <param name="projectPath">The project path can be an absolute or a relative path to the
+        /// target application project directory or the application project file.</param>
+        /// <returns><see cref="ProjectDefinition"/></returns>
         public async Task<ProjectDefinition> Parse(string projectPath)
         {
             if (_directoryManager.Exists(projectPath))
             {
+                projectPath = _directoryManager.GetDirectoryInfo(projectPath).FullName;
                 var files = _directoryManager.GetFiles(projectPath, "*.csproj");
                 if (files.Length == 1)
                 {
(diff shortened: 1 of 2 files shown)


philasmar (approved):

src/AWS.Deploy.Common/ProjectDefinitionParser.cs
```diff
@@ -37,6 +37,7 @@ public async Task<ProjectDefinition> Parse(string projectPath)
         {
```
ganeshnj, line 37: As you are updating this code, can you add documentation on the behavior of `projectPath` param?

src/AWS.Deploy.Common/ProjectDefinitionParser.cs
```diff
@@ -37,6 +37,7 @@ public async Task<ProjectDefinition> Parse(string projectPath)
...
             if (_directoryManager.Exists(projectPath))
             {
+                projectPath = _directoryManager.GetDirectoryInfo(projectPath).FullName;
```
ganeshnj, line 40: - What testing you have performed?
- How can we ensure it will not break in future?
- Can this be covered via unit tests?

ganeshnj (changes requested):

src/AWS.Deploy.Common/ProjectDefinitionParser.cs
96malhar: Added

src/AWS.Deploy.Common/ProjectDefinitionParser.cs
96malhar: Added unit tests for this. The tests will fail if the change is removed

test/AWS.Deploy.CLI.UnitTests/ProjectDefinitionParserTest.cs
```diff
@@ -0,0 +1,98 @@
...
+            // Act
+            try
+            {
+                projectDefinition = await new ProjectDefinitionParser(new FileManager(), new DirectoryManager())
+                    .Parse(relativeProjectDirectoryPath);
+            }
+            catch (ProjectFileNotFoundException ex)
+            {
+                exception = ex;
+            }
```
ganeshnj, line 51: This is unnecessary, you can remove try-catch. If the method throws, we will know that something is broken.

In test we should only check for throws.

96malhar: Fixed

ganeshnj (approved):

PR #287, opened 2021-08-05
feat: Make ASP.NET Core ECS recipe customizable

normj:
*Description of changes:*
Rework the ASP.NET Core ECS recipe to support persisted deployment projects being able to customize the original recipe's constructs while allowing the deploy tool future opportunities to update the original generated files.

Started working on crafting a README.md for the deployment project. Don't worry too much about the content right now because I'm sure it will change as we get more docs put in place. For now just double check for grammar issues in the README.md.

Also added the following new settings:
* Ability to configure autoscaling for ECS services using either Cpu, Memory or Request metrics
* Ability to adjust the deregistering delay for terminating ECS tasks during redeployment
* Ability to configure to use existing load balancer (with typehint)
* Ability to configure health checks
* Ability to configure Listener Condition rule path patterns. In the future we could other rule types like query strings or headers.
* Made Application IAM role updatable. This is safe because it is just an update the task definition used for the new tasks spun up. It doesn't cause the endpoint to change like Beanstalk does.


src/AWS.Deploy.CLI/Commands/DeployCommand.cs
```diff
@@ -728,11 +731,12 @@ private async Task ConfigureDeploymentFromCli(Recommendation recommendation, Opt
                     settingValue = await typeHintCommand.Execute(recommendation, setting);
                 }
                 else
-                {
+                {
```
philasmar, line 734: nit: can you remove the whitespace?

src/AWS.Deploy.CLI/ConsoleUtilities.cs
```diff
@@ -410,9 +410,9 @@ public void DisplayValues(Dictionary<string, object> objectValues, string indent
                         _interactiveService.WriteLine($"{indent}{key}: {stringValue}");
                     }
                 }
-                else if (value is bool boolValue)
+                else if(value != null)
                 {
-                    _interactiveService.WriteLine($"{indent}{key}: {boolValue}");
+                    _interactiveService.WriteLine($"{indent}{key}: {value}");
                 }
```
philasmar, line 416: do we want to open this up to accept other data types if they are not supported?

src/AWS.Deploy.Recipes.CDK.Common/CDKRecipeCustomizer.cs
```diff
@@ -0,0 +1,47 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+using System;
+using System.Collections.Generic;
+using System.Text;
+
+namespace AWS.Deploy.Recipes.CDK.Common
+{
+    public delegate void CustomizePropsDelegate<GeneratedConstruct>(CustomizePropsEventArgs<GeneratedConstruct> evnt);
```
philasmar, line 10: nit: might as well use "event" instead of "evnt"

src/AWS.Deploy.Recipes/CdkTemplates/AspNetAppEcsFargate/AppStack.cs
```diff
@@ -1,16 +1,14 @@
...
 using Amazon.CDK.AWS.ECS;
-using Amazon.CDK.AWS.ECS.Patterns;
-using Amazon.CDK.AWS.IAM;
 using AWS.Deploy.Recipes.CDK.Common;
+
 using AspNetAppEcsFargate.Configurations;
-using Protocol = Amazon.CDK.AWS.ECS.Protocol;
-using Amazon.CDK.AWS.ECR;
-using System.Collections.Generic;
+
```
philasmar, line 11: nit: can you remove this new line?

src/AWS.Deploy.Recipes/CdkTemplates/AspNetAppEcsFargate/AppStack.cs
```diff
@@ -19,121 +17,42 @@ public class AppStack : Stack
...
-                TaskDefinition = taskDefinition,
-                DesiredCount = settings.DesiredCount,
-                ServiceName = settings.ECSServiceName,
-                AssignPublicIp = settings.Vpc.IsDefault,
-                SecurityGroups = ecsServiceSecurityGroups.ToArray()
-            });
+        /// <summary>
+        /// This method can be used to customize the properties for CDK constructs before creating the constructs.
+        ///
+        /// The pattern used in this method is to check to evnt.ResourceLogicalName to see if the CDK construct about to be created is one
```
philasmar, line 37: typo: "check the event.ResourceLogicalName" instead of "check to evnt.ResourceLogicalName"

src/AWS.Deploy.Recipes/RecipeDefinitions/ASP.NETAppECSFargate.recipe
```diff
@@ -406,6 +406,383 @@
...
+                            "Id": "AutoScaling.ScalingType",
+                            "Value": "Cpu"
+                        }
+                    ]
+                },
+                {
+                    "Id": "MemoryTypeTargetUtilizationPercent",
+                    "Name": "Memory Target Utilization",
+                    "Description": "The target memory utilization that triggers a scaling change.",
+                    "Type": "Double",
```
philasmar, line 675: Why is this a double and not an Int?

src/AWS.Deploy.Recipes/CdkTemplates/AspNetAppEcsFargate/README.md
```diff
@@ -1,14 +1,55 @@
...
+CDK CLI passing in all of the settings gathered in the AWS deploy tool.
+
+## Can I modify the deployment project?
+
+When a deployment projects is saved the project can be customized by adding more CDK constructs or customizing the existing
+CDK constructs.
+
+The default folder structure puts the CDK constructs originally defined by the deployment recipe into a folder called
+"Generated". It is recommended to not directly modify these files and instead customize the settings via the
+AppStack.CustomizeCDKProps() method. This allows the AWS deploy tool to be able easily updated the generated code
```
philasmar, line 40: grammar: "This allows the AWS deploy tool to easily update the generated"

src/AWS.Deploy.Recipes/CdkTemplates/AspNetAppEcsFargate/Generated/RecipeConstruct.cs
```diff
@@ -0,0 +1,342 @@
...
+        public AwsLogDriver? AppLogging { get; private set; }
+
+        public ScalableTaskCount? AutoScalingConfiguration { get; private set; }
+
+        public string AutoScaleTaskCount { get; } = "AutoScaleTaskCount";
+        public string AutoScaleTypeCpuType { get; } = "CpuScaling";
+        public string AutoScaleTypeRequestType { get; } = "RequestScaling";
+        public string AutoScaleTypeMemoryType { get; } = "MemoryScaling";
+
+        public string AddTargetGroup { get; } = "AddTargetGroup";
```
ganeshnj, line 47: Can these be constants? 

Also, may be worth putting inside a (inner) class.

src/AWS.Deploy.Recipes/CdkTemplates/AspNetAppEcsFargate/Generated/RecipeConstruct.cs
```diff
@@ -0,0 +1,342 @@
...
+
+        public RecipeConstruct(Construct scope, RecipeConfiguration<Configuration> recipeConfiguration, IStackProps? props = null)
+            : base(scope, "GeneratedRecipeConstruct")
```
ganeshnj, line 50: This should be an id supplied during the instantiation of the `RecipeConstruct` like any other construct.

src/AWS.Deploy.Recipes/CdkTemplates/AspNetAppEcsFargate/Generated/RecipeConstruct.cs
```diff
@@ -0,0 +1,342 @@
...
+        public ScalableTaskCount? AutoScalingConfiguration { get; private set; }
+
+        public string AutoScaleTaskCount { get; } = "AutoScaleTaskCount";
+        public string AutoScaleTypeCpuType { get; } = "CpuScaling";
+        public string AutoScaleTypeRequestType { get; } = "RequestScaling";
+        public string AutoScaleTypeMemoryType { get; } = "MemoryScaling";
+
+        public string AddTargetGroup { get; } = "AddTargetGroup";
+
+        public RecipeConstruct(Construct scope, RecipeConfiguration<Configuration> recipeConfiguration, IStackProps? props = null)
```
ganeshnj, line 49: A recipe construct must not accept `IStackProps` in its constructor.

There should be a separate Props class like `RecipeConstructProps` which also takes care of `RecipeConfiguration<Configuration>`.

```suggestion
        public RecipeConstruct(Construct scope, string id, IRecipeConstuctProps? props = null)
```

Also having a suffix Construct doesn't go with CDK construct naming, possibly just `Recipe`?

src/AWS.Deploy.Recipes/CdkTemplates/AspNetAppEcsFargate/Generated/Configurations/AutoScalingConfiguration.cs
```diff
@@ -0,0 +1,46 @@
...
+    {
+        const int defaultCooldown = 300;
+
+        public bool Enabled { get; set; }
+
+        public int MinCapacity { get; set; } = 3;
+
+        public int MaxCapacity { get; set; } = 6;
+
+        public string ScalingType { get; set; } = "Cpu";
```
ganeshnj, line 20: Can this be enum?

src/AWS.Deploy.Recipes/CdkTemplates/AspNetAppEcsFargate/Generated/RecipeConstruct.cs
```diff
@@ -0,0 +1,342 @@
...
+            });
+
+            WebAccessSecurityGroup = new SecurityGroup(this, nameof(WebAccessSecurityGroup), InvokeCustomizeCDKPropsEvent(nameof(WebAccessSecurityGroup), this, new SecurityGroupProps
+            {
+                Vpc = AppVpc,
+                SecurityGroupName = $"{recipeConfiguration.StackName}-ECSService"
+            }));
+
+            var ecsServiceSecurityGroups = new List<ISecurityGroup>();
+            ecsServiceSecurityGroups.Add(WebAccessSecurityGroup);
```
ganeshnj, line 167: is it suppose to reference the instance version of `EcsServiceSecurityGroups`?

```suggestion
            EcsServiceSecurityGroups.Add(WebAccessSecurityGroup);
```

src/AWS.Deploy.Recipes/RecipeDefinitions/ASP.NETAppECSFargate.recipe
ganeshnj: https://docs.aws.amazon.com/cdk/api/latest/docs/@aws-cdk_aws-ecs.CpuUtilizationScalingProps.html#targetutilizationpercent

underlying data structure is double

ganeshnj (changes requested):

src/AWS.Deploy.CLI/Commands/DeployCommand.cs
normj: Fixed

src/AWS.Deploy.CLI/ConsoleUtilities.cs
normj: I don't understand why. The code is the same in each block and it triggers bugs when new types are added and we didn't know to update all of the places it matters. For example this code was already broken not showing ints because who ever added ints didn't know to update this block of code.

src/AWS.Deploy.Recipes.CDK.Common/CDKRecipeCustomizer.cs
normj: `evnt` is the common name to use since `event` is a reserved word in .NET.

src/AWS.Deploy.Recipes/CdkTemplates/AspNetAppEcsFargate/AppStack.cs
normj: Done

src/AWS.Deploy.Recipes/CdkTemplates/AspNetAppEcsFargate/AppStack.cs
normj: On purpose because `event` doesn't work as a variable name since it is a reserved word in .NET.

src/AWS.Deploy.Recipes/RecipeDefinitions/ASP.NETAppECSFargate.recipe
normj: Correct the service/cloudformation value is a double

src/AWS.Deploy.Recipes/CdkTemplates/AspNetAppEcsFargate/README.md
normj: Removed the "be able". I had it in because we haven't actually implemented the update features and I was being cautious in case we don't get around to that for awhile.

src/AWS.Deploy.CLI/Commands/DeployCommand.cs
```diff
@@ -360,6 +360,9 @@ private void ConfigureDeploymentFromConfigFile(Recommendation recommendation, Us
                             case OptionSettingValueType.Bool:
                                 settingValue = bool.Parse(optionSettingValue);
                                 break;
+                            case OptionSettingValueType.Double:
+                                settingValue = Double.Parse(optionSettingValue);
```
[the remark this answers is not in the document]
normj: Fixed

src/AWS.Deploy.Recipes.CDK.Common/CDKRecipeCustomizer.cs
```diff
@@ -0,0 +1,47 @@
...
+
+    public class CustomizePropsEventArgs<GeneratedConstruct>
+    {
+        public object Props { get; set; }
+        public string ResourceLogicalName { get; set; }
+        public GeneratedConstruct Construct { get; set; }
```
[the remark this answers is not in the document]
normj: Made the setters private since they are all set during the constructor.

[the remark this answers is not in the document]
normj: `GeneratedConstruct` is the containing construct for the construct about to be created with the `ResourceLogicalName` logical.

I added missing documentation to hopefully make it more clear.

src/AWS.Deploy.Recipes.CDK.Common/CDKRecipeCustomizer.cs
```diff
@@ -0,0 +1,47 @@
...
+        }
+    }
+
+    public class CDKCustomizationException : Exception
+    {
+        public CDKCustomizationException(string message) : base(message)
+        {
+
+        }
+    }
```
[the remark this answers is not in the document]
normj: Removed

src/AWS.Deploy.Recipes/CdkTemplates/AspNetAppEcsFargate/Generated/Configurations/AutoScalingConfiguration.cs
```diff
@@ -0,0 +1,46 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+using System;
+using System.Collections.Generic;
+using System.Text;
+
+namespace AspNetAppEcsFargate.Configurations
```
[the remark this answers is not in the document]
normj: We had always said we wanted to give user's the code directly so they could see how things are created and if they want easily disconnect themselves from our template and customize the deployment however they want. Switching to NuGet packages would be a major shift. It would make story potentially easier but lose a lot in the teaching/learning side.

src/AWS.Deploy.Recipes/CdkTemplates/AspNetAppEcsFargate/Generated/Configurations/Configuration.cs
```diff
@@ -0,0 +1,83 @@
...
+
+        /// <summary>
+        /// Comma-delimited list of security groups assigned to the ECS service.
+        /// </summary>
+        public string AdditionalECSServiceSecurityGroups { get; set; }
+
+        /// <inheritdoc cref="FargateTaskDefinitionProps.Cpu"/>
+        public double? TaskCpu { get; set; }
+
+        /// <inheritdoc cref="FargateTaskDefinitionProps.MemoryLimitMiB"/>
```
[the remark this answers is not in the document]
normj: Replaced with inline comments. The `inheritdoc` doesn't make sense there.

src/AWS.Deploy.Recipes/CdkTemplates/AspNetAppEcsFargate/Generated/Configurations/LoadBalancerConfiguration.cs
```diff
@@ -0,0 +1,84 @@
...
+        /// </summary>
+        public int DeregistrationDelayInSeconds { get; set; } = 60;
+
+        /// <summary>
+        /// The ping path destination where Elastic Load Balancing sends health check requests.
+        /// </summary>
+        public string? HealthCheckPath { get; set; }
+
+        /// <summary>
+        /// The approximate number of seconds ebetween health checks.
```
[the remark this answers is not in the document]
normj: Fixed

src/AWS.Deploy.Recipes/CdkTemplates/AspNetAppEcsFargate/Generated/Configurations/LoadBalancerConfiguration.cs
```diff
@@ -0,0 +1,84 @@
...
+        /// </summary>
+        public int? HealthCheckInternval { get; set; }
```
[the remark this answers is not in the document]
normj: That just makes the serialization from settings file and deploy tool to the recipe harder and not necessary.

src/AWS.Deploy.Recipes/CdkTemplates/AspNetAppEcsFargate/Generated/Configurations/LoadBalancerConfiguration.cs
```diff
@@ -0,0 +1,84 @@
...
+
+        /// <summary>
+        /// The resource path pattern to use with ListenerConditionType is set to "Path"
+        /// </summary>
+        public string? ListenerConditionPathPattern { get; set; }
+
+        /// <summary>
+        /// The priority of the listener condition rule.
+        /// </summary>
+        public int ListenerConditionPriority { get; set; } = 100;
```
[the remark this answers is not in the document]
normj: Fixed

src/AWS.Deploy.Recipes/CdkTemplates/AspNetAppEcsFargate/Generated/RecipeConstruct.cs
```diff
@@ -0,0 +1,342 @@
...
+    public class RecipeConstruct : Construct
+    {
+        public IVpc? AppVpc { get; private set; }
+        public ICluster? EcsCluster { get; private set; }
+        public IRole? AppIAMTaskRole { get; private set; }
+        public TaskDefinition? AppTaskDefinition { get; private set; }
+        public IRepository? EcrRepository { get; private set; }
+        public ContainerDefinition? AppContainerDefinition { get; private set; }
+        public SecurityGroup? WebAccessSecurityGroup { get; private set; }
+        public IList<ISecurityGroup>? EcsServiceSecurityGroups { get; private set; }
```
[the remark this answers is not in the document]
normj: Fixed incorrect use of a local variable to use the `EcsServiceSecurityGroups` property.

src/AWS.Deploy.Recipes/CdkTemplates/AspNetAppEcsFargate/Generated/RecipeConstruct.cs
normj: We want the name to be constant otherwise it changes all of the logical names specified for display resources.

src/AWS.Deploy.Recipes/CdkTemplates/AspNetAppEcsFargate/Generated/RecipeConstruct.cs
normj: Fixed

src/AWS.Deploy.Recipes/CdkTemplates/AspNetAppEcsFargate/Generated/RecipeConstruct.cs
normj: Normally when customizing the props for a CDK construct the user would write code that looks like 
```
if (string.Equals(evnt.ResourceLogicalName, nameof(evnt.Construct.AppContainerDefinition)))
{
   // Customize props
}
```

So to avoid having magical strings in the user's `CustomizeCDKProps` method I suggest the user does a `nameof` operation on the `evnt.Construct.<CDK-Constuct-Property>`.

These CDK properties are used in methods that don't return an object so I can't use the exact same pattern. But by having these as instance properties the end user experience is exact same of calling `evnt.Construct.AutoScaleTypeCpuType`. If I make these constants then the user has to have different behavior for these CDK properties and access them as static from the type name. I don't want users to have to know about the difference.

I did delete `AutoScaleTaskCount` as that was not needed since there was the `AutoScalingConfiguration` property. I also made sure the values for each of the properties match the property name to make sure the `nameof` operator worked correctly.

src/AWS.Deploy.Recipes/CdkTemplates/AspNetAppEcsFargate/Generated/RecipeConstruct.cs
normj: Removed passing in `IStackProps` but I think creating a separate props for the construct with just `RecipeConfiguration<Configuration>` which is basically props is overkill.

src/AWS.Deploy.Recipes/CdkTemplates/AspNetAppEcsFargate/Generated/Configurations/LoadBalancerConfiguration.cs
```diff
@@ -0,0 +1,84 @@
...
+
+        /// <summary>
+        /// The number of consecutive health check successes required before considering an unhealthy target unhealthy.
+        /// </summary>
+        public int? UnhealthyThresholdCount { get; set; }
+
+        /// <summary>
+        /// The type of listener condition to create. Current valid values are "None" and "Path"
+        /// </summary>
+        public string? ListenerConditionType { get; set; }
```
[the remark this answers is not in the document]
normj: Done

src/AWS.Deploy.Recipes/CdkTemplates/AspNetAppEcsFargate/Generated/Configurations/AutoScalingConfiguration.cs
normj: Done

philasmar (approved):

ganeshnj (approved):

PR #288, opened 2021-08-07
feat: Save last used stack and order existing deployments by MRU stack

philasmar:
*Issue #, if available:*

*Description of changes:*
feat: Save last used stack and order existing deployments by MRU stack

### DIFF
--- a/src/AWS.Deploy.CLI/Commands/CommandFactory.cs
+++ b/src/AWS.Deploy.CLI/Commands/CommandFactory.cs
@@ -21,6 +21,7 @@ using AWS.Deploy.CLI.Commands.CommandHandlerInput;
 using AWS.Deploy.Common.IO;
 using AWS.Deploy.Common.DeploymentManifest;
 using AWS.Deploy.Orchestration.DisplayedResources;
+using AWS.Deploy.Orchestration.LocalUserSettings;
 
 namespace AWS.Deploy.CLI.Commands
 {
@@ -61,8 +62,11 @@ namespace AWS.Deploy.CLI.Commands
         private readonly ITypeHintCommandFactory _typeHintCommandFactory;
         private readonly IDisplayedResourcesHandler _displayedResourceHandler;
         private readonly IConsoleUtilities _consoleUtilities;
+        private readonly IDirectoryManager _directoryManager;
+        private readonly IFileManager _fileManager;
         private readonly IDeploymentManifestEngine _deploymentManifestEngine;
         private readonly ICustomRecipeLocator _customRecipeLocator;
+        private readonly ILocalUserSettingsEngine _localUserSettingsEngine;
 
         public CommandFactory(
             IToolInteractiveService toolInteractiveService,
@@ -82,8 +86,11 @@ namespace AWS.Deploy.CLI.Commands
             ITypeHintCommandFactory typeHintCommandFactory,
             IDisplayedResourcesHandler displayedResourceHandler,
             IConsoleUtilities consoleUtilities,
+            IDirectoryManager directoryManager,
+            IFileManager fileManager,
             IDeploymentManifestEngine deploymentManifestEngine,
-            ICustomRecipeLocator customRecipeLocator)
+            ICustomRecipeLocator customRecipeLocator,
+            ILocalUserSettingsEngine localUserSettingsEngine)
         {
             _toolInteractiveService = toolInteractiveService;
             _orchestratorInteractiveService = orchestratorInteractiveService;
@@ -102,8 +109,11 @@ namespace AWS.Deploy.CLI.Commands
             _typeHintCommandFactory = typeHintCommandFactory;
             _displayedResourceHandler = displayedResourceHandler;
             _consoleUtilities = consoleUtilities;
+            _directoryManager = directoryManager;
+            _fileManager = fileManager;
             _deploymentManifestEngine = deploymentManifestEngine;
             _customRecipeLocator = customRecipeLocator;
+            _localUserSettingsEngine = localUserSettingsEngine;
         }
 
         public Command BuildRootCommand()
@@ -194,6 +204,7 @@ namespace AWS.Deploy.CLI.Commands
                         _typeHintCommandFactory,
                         _displayedResourceHandler,
                         _cloudApplicationNameGenerator,
+                        _localUserSettingsEngine,
                         _consoleUtilities,
                         _customRecipeLocator,
                         session);
@@ -269,7 +280,28 @@ namespace AWS.Deploy.CLI.Commands
                         return CommandReturnCodes.USER_ERROR;
                     }
 
-                    await new DeleteDeploymentCommand(_awsClientFactory, _toolInteractiveService, _consoleUtilities).ExecuteAsync(input.DeploymentName);
+                    OrchestratorSession? session = null;
+
+                    try
+                    {
+                        var projectDefinition = await _projectParserUtility.Parse(input.ProjectPath ?? string.Empty);
+
+                        var callerIdentity = await _awsResourceQueryer.GetCallerIdentity();
+
+                        session = new OrchestratorSession(
+                            projectDefinition,
+                            awsCredentials,
+                            awsRegion,
+                            callerIdentity.Account);
+                    }
+                    catch (FailedToFindDeployableTargetException) { }
+
+                    await new DeleteDeploymentCommand(
+                        _awsClientFactory,
+                        _toolInteractiveService,
+                        _consoleUtilities,
+                        _localUserSettingsEngine,
+                        session).ExecuteAsync(input.DeploymentName);
 
                     return CommandReturnCodes.SUCCESS;
                 }
@@ -396,8 +428,8 @@ namespace AWS.Deploy.CLI.Commands
                         _consoleUtilities,
                         _cdkProjectHandler,
                         _commandLineWrapper,
-                        new DirectoryManager(),
-                        new FileManager(),
+                        _directoryManager,
+                        _fileManager,
                         session,
                         _deploymentManifestEngine,
                         targetApplicationFullPath);
--- a/src/AWS.Deploy.CLI/Commands/DeleteDeploymentCommand.cs
+++ b/src/AWS.Deploy.CLI/Commands/DeleteDeploymentCommand.cs
@@ -8,7 +8,9 @@ using Amazon.CloudFormation;
 using Amazon.CloudFormation.Model;
 using AWS.Deploy.CLI.CloudFormation;
 using AWS.Deploy.Common;
-using AWS.Deploy.Recipes.CDK.Common;
+using AWS.Deploy.Common.IO;
+using AWS.Deploy.Orchestration;
+using AWS.Deploy.Orchestration.LocalUserSettings;
 
 namespace AWS.Deploy.CLI.Commands
 {
@@ -23,14 +25,23 @@ namespace AWS.Deploy.CLI.Commands
         private readonly IToolInteractiveService _interactiveService;
         private readonly IAmazonCloudFormation _cloudFormationClient;
         private readonly IConsoleUtilities _consoleUtilities;
+        private readonly ILocalUserSettingsEngine _localUserSettingsEngine;
+        private readonly OrchestratorSession? _session;
         private const int MAX_RETRIES = 4;
 
-        public DeleteDeploymentCommand(IAWSClientFactory awsClientFactory, IToolInteractiveService interactiveService, IConsoleUtilities consoleUtilities)
+        public DeleteDeploymentCommand(
+            IAWSClientFactory awsClientFactory,
+            IToolInteractiveService interactiveService,
+            IConsoleUtilities consoleUtilities,
+            ILocalUserSettingsEngine localUserSettingsEngine,
+            OrchestratorSession? session)
         {
             _awsClientFactory = awsClientFactory;
             _interactiveService = interactiveService;
             _consoleUtilities = consoleUtilities;
             _cloudFormationClient = _awsClientFactory.GetAWSClient<IAmazonCloudFormation>();
+            _localUserSettingsEngine = localUserSettingsEngine;
+            _session = session;
         }
 
         /// <summary>
@@ -67,6 +78,12 @@ namespace AWS.Deploy.CLI.Commands
                 var _ = monitor.StartAsync();
 
                 await WaitForStackDelete(stackName);
+
+                if (_session != null)
+                {
+                    await _localUserSettingsEngine.DeleteLastDeployedStack(stackName, _session.ProjectDefinition.ProjectName, _session.AWSAccountId, _session.AWSRegion);
+                }
+
                 _interactiveService.WriteLine($"{stackName}: deleted");
             }
             finally
--- a/src/AWS.Deploy.CLI/Commands/DeployCommand.cs
+++ b/src/AWS.Deploy.CLI/Commands/DeployCommand.cs
@@ -18,6 +18,8 @@ using AWS.Deploy.Recipes;
 using AWS.Deploy.Orchestration.Data;
 using AWS.Deploy.Orchestration.Utilities;
 using AWS.Deploy.Orchestration.DisplayedResources;
+using AWS.Deploy.Common.IO;
+using AWS.Deploy.Orchestration.LocalUserSettings;
 
 namespace AWS.Deploy.CLI.Commands
 {
@@ -35,6 +37,7 @@ namespace AWS.Deploy.CLI.Commands
         private readonly ITypeHintCommandFactory _typeHintCommandFactory;
         private readonly IDisplayedResourcesHandler _displayedResourcesHandler;
         private readonly ICloudApplicationNameGenerator _cloudApplicationNameGenerator;
+        private readonly ILocalUserSettingsEngine _localUserSettingsEngine;
         private readonly IConsoleUtilities _consoleUtilities;
         private readonly ICustomRecipeLocator _customRecipeLocator;
 
@@ -53,6 +56,7 @@ namespace AWS.Deploy.CLI.Commands
             ITypeHintCommandFactory typeHintCommandFactory,
             IDisplayedResourcesHandler displayedResourcesHandler,
             ICloudApplicationNameGenerator cloudApplicationNameGenerator,
+            ILocalUserSettingsEngine localUserSettingsEngine,
             IConsoleUtilities consoleUtilities,
             ICustomRecipeLocator customRecipeLocator,
             OrchestratorSession session)
@@ -68,6 +72,7 @@ namespace AWS.Deploy.CLI.Commands
             _typeHintCommandFactory = typeHintCommandFactory;
             _displayedResourcesHandler = displayedResourcesHandler;
             _cloudApplicationNameGenerator = cloudApplicationNameGenerator;
+            _localUserSettingsEngine = localUserSettingsEngine;
             _consoleUtilities = consoleUtilities;
             _session = session;
             _cdkManager = cdkManager;
@@ -131,6 +136,7 @@ namespace AWS.Deploy.CLI.Commands
                     _cdkManager,
                     _awsResourceQueryer,
                     _deploymentBundleHandler,
+                    _localUserSettingsEngine,
                     _dockerEngine,
                     _customRecipeLocator,
                     new List<string> { RecipeLocator.FindRecipeDefinitionsPath() });
@@ -142,7 +148,7 @@ namespace AWS.Deploy.CLI.Commands
             var allDeployedApplications = await _deployedApplicationQueryer.GetExistingDeployedApplications();
 
             // Filter compatible applications that can be re-deployed  using the current set of recommendations.
-            var compatibleApplications = GetCompatibleApplications(allDeployedApplications, recommendations);
+            var compatibleApplications = await _deployedApplicationQueryer.GetCompatibleApplications(recommendations, allDeployedApplications, _session);
 
             // Get Cloudformation stack name.
             var cloudApplicationName = GetCloudApplicationName(stackName, userDeploymentSettings, compatibleApplications);
@@ -178,23 +184,6 @@ namespace AWS.Deploy.CLI.Commands
             var cloudApplication = new CloudApplication(cloudApplicationName, selectedRecommendation.Recipe.Id);
 
             return (orchestrator, selectedRecommendation, cloudApplication);
-        }
-
-        /// <summary>
-        /// Filters the applications that can be re-deployed using the current set of available recommendations.
-        /// </summary>
-        /// <param name="allDeployedApplications"></param>
-        /// <param name="recommendations"></param>
-        /// <returns>A list of <see cref="CloudApplication"/> that are compatible for a re-deployment</returns>
-        private List<CloudApplication> GetCompatibleApplications(List<CloudApplication> allDeployedApplications, List<Recommendation> recommendations)
-        {
-            var compatibleApplications = new List<CloudApplication>();
-            foreach (var app in allDeployedApplications)
-            {
-                if (recommendations.Any(rec => string.Equals(rec.Recipe.Id, app.RecipeId, StringComparison.Ordinal)))
-                    compatibleApplications.Add(app);
-            }
-            return compatibleApplications;
         }
 
         /// <summary>
@@ -574,7 +563,7 @@ namespace AWS.Deploy.CLI.Commands
                     }
 
                     _toolInteractiveService.WriteLine(string.Empty);
-                    var answer = _consoleUtilities.AskYesNoQuestion("Do you want to go back and modify the current configuration?", "true");
+                    var answer = _consoleUtilities.AskYesNoQuestion("Do you want to go back and modify the current configuration?", "false");
                     if (answer == YesNo.Yes)
                     {
                         var dockerExecutionDirectory =
--- a/src/AWS.Deploy.CLI/Extensions/CustomServiceCollectionExtension.cs
+++ b/src/AWS.Deploy.CLI/Extensions/CustomServiceCollectionExtension.cs
@@ -12,6 +12,7 @@ using AWS.Deploy.Orchestration;
 using AWS.Deploy.Orchestration.CDK;
 using AWS.Deploy.Orchestration.Data;
 using AWS.Deploy.Orchestration.DisplayedResources;
+using AWS.Deploy.Orchestration.LocalUserSettings;
 using AWS.Deploy.Orchestration.Utilities;
 using Microsoft.Extensions.DependencyInjection;
 using Microsoft.Extensions.DependencyInjection.Extensions;
@@ -54,6 +55,7 @@ namespace AWS.Deploy.CLI.Extensions
             serviceCollection.TryAdd(new ServiceDescriptor(typeof(IZipFileManager), typeof(ZipFileManager), lifetime));
             serviceCollection.TryAdd(new ServiceDescriptor(typeof(IDeploymentManifestEngine), typeof(DeploymentManifestEngine), lifetime));
             serviceCollection.TryAdd(new ServiceDescriptor(typeof(ICustomRecipeLocator), typeof(CustomRecipeLocator), lifetime));
+            serviceCollection.TryAdd(new ServiceDescriptor(typeof(ILocalUserSettingsEngine), typeof(LocalUserSettingsEngine), lifetime));
             serviceCollection.TryAdd(new ServiceDescriptor(typeof(ICommandFactory), typeof(CommandFactory), lifetime));
 
             var packageJsonTemplate = typeof(PackageJsonGenerator).Assembly.ReadEmbeddedFile(PackageJsonGenerator.TemplateIdentifier);
--- a/src/AWS.Deploy.CLI/ServerMode/Controllers/DeploymentController.cs
+++ b/src/AWS.Deploy.CLI/ServerMode/Controllers/DeploymentController.cs
@@ -26,6 +26,8 @@ using Microsoft.AspNetCore.Authorization;
 using Amazon.Runtime;
 using AWS.Deploy.Common.Recipes;
 using AWS.Deploy.Orchestration.DisplayedResources;
+using AWS.Deploy.Common.IO;
+using AWS.Deploy.Orchestration.LocalUserSettings;
 
 namespace AWS.Deploy.CLI.ServerMode.Controllers
 {
@@ -242,13 +244,16 @@ namespace AWS.Deploy.CLI.ServerMode.Controllers
             var output = new GetExistingDeploymentsOutput();
 
             var deployedApplicationQueryer = serviceProvider.GetRequiredService<IDeployedApplicationQueryer>();
-            state.ExistingDeployments = await deployedApplicationQueryer.GetExistingDeployedApplications(state.NewRecommendations);
+            var session = CreateOrchestratorSession(state);
+            state.ExistingDeployments = await deployedApplicationQueryer.GetCompatibleApplications(state.NewRecommendations.ToList(), session: session);
 
             foreach(var deployment in state.ExistingDeployments)
             {
                 output.ExistingDeployments.Add(new ExistingDeploymentSummary(
                     deployment.Name,
-                    deployment.RecipeId));
+                    deployment.RecipeId,
+                    deployment.LastUpdatedTime,
+                    deployment.UpdatedByCurrentUser));
             }
 
             return Ok(output);
@@ -434,6 +439,16 @@ namespace AWS.Deploy.CLI.ServerMode.Controllers
             return serviceProvider;
         }
 
+        private OrchestratorSession CreateOrchestratorSession(SessionState state, AWSCredentials? awsCredentials = null)
+        {
+            return new OrchestratorSession(
+                state.ProjectDefinition,
+                awsCredentials ?? HttpContext.User.ToAWSCredentials() ??
+                    throw new FailedToRetrieveAWSCredentialsException("The tool was not able to retrieve the AWS Credentials."),
+                state.AWSRegion,
+                state.AWSAccountId);
+        }
+
         private Orchestrator CreateOrchestrator(SessionState state, IServiceProvider? serviceProvider = null, AWSCredentials? awsCredentials = null)
         {
             if(serviceProvider == null)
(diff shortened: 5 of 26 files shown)


src/AWS.Deploy.CLI/Commands/DeleteDeploymentCommand.cs
```diff
@@ -66,6 +69,7 @@ public async Task ExecuteAsync(string stackName)
                 var _ = monitor.StartAsync();
 
                 await WaitForStackDelete(stackName);
+
```
normj, line 72: Should we be removing the stack from `aws-deployments.json` after we delete it?

src/AWS.Deploy.Common/DeploymentManifest/DeploymentManifestModel.cs
```diff
@@ -11,11 +11,14 @@ namespace AWS.Deploy.Common.DeploymentManifest
     /// </summary>
     public class DeploymentManifestModel
     {
+        public List<LastDeployedStack> LastDeployedStacks { get; set; }
+
         [JsonProperty("deployment-projects")]
         public List<DeploymentManifestEntry> DeploymentManifestEntries { get; set; }
```
normj, line 17: I'm seeing when I use this that `deployment-projects` is getting written to the JSON file as an empty string. I'm assuming `LastDeployedStacks` would also be serialized as an empty array as well if there were no values. Malhar did some work in his previous PR to make sure we weren't writing out empty lists when serializing the recipe. Check with him to see what he did.

src/AWS.Deploy.Common/DeploymentManifest/DeploymentManifestModel.cs
```diff
@@ -11,11 +11,14 @@ namespace AWS.Deploy.Common.DeploymentManifest
     /// </summary>
     public class DeploymentManifestModel
     {
+        public List<LastDeployedStack> LastDeployedStacks { get; set; }
```
normj, line 14: JSON property names are inconsistent. This one is `LastDeployedStacks` and deployment projects is `deployment-projects`. When should use either pascal casing or hyphen across the file.

src/AWS.Deploy.Orchestration/DeploymentManifestEngine.cs
```diff
@@ -0,0 +1,222 @@
...
+                    $"for the deployment project stored at '{saveCdkDirectoryFullPath}'", ex);
+            }
+            
+        }
+
+        /// <summary>
+        /// This method updates the deployment manifest json file by adding the name of the stack that was most recently used.
+        /// If the manifest file does not exists then a new file is generated.
+        /// </summary>
+        public async Task UpdateLastDeployedStack(string stackName)
```
normj, line 87: Should this method also clean up any orphan stacks in the LastDeployedStacks for the current account and region? Or is there some other mechanism to clean up the file for stacks deleted in the console or some other tool.

src/AWS.Deploy.Orchestration/DeploymentManifestEngine.cs
```diff
@@ -0,0 +1,222 @@
...
+        /// </summary>
+        public async Task DeleteLastDeployedStack(string stackName)
+        {
+            try
+            {
+                var deploymentManifestModel = await GetDeploymentManifest();
+                var lastDeployedStack = deploymentManifestModel?.LastDeployedStacks
+                    .FirstOrDefault(x => x.AWSAccountId.Equals(_session.AWSAccountId) && x.AWSRegion.Equals(_session.AWSRegion));
+
+                if (deploymentManifestModel != null && lastDeployedStack != null)
```
normj, line 144: nit: lastDeployedStack has to be null if deploymentManifestModel so you could just check `lastDeployedStack != null`

src/AWS.Deploy.Orchestration/DeploymentManifestEngine.cs
```diff
@@ -0,0 +1,222 @@
...
+        /// This method parses the deployment-manifest file into a <see cref="DeploymentManifestModel"/>
+        /// </summary>
+        public async Task<DeploymentManifestModel?> GetDeploymentManifest()
+        {
+            var deploymentManifestFilePath = GetDeploymentManifestFilePath(_targetApplicationFullPath);
+
+            if (!_fileManager.Exists(deploymentManifestFilePath))
+                return null;
+            var manifestFilejsonString = await _fileManager.ReadAllTextAsync(deploymentManifestFilePath);
+            return JsonConvert.DeserializeObject<DeploymentManifestModel>(manifestFilejsonString);
```
normj, line 197: I think if we fail to deserialize `aws-deployments.json` we should log it and then return an empty `DeploymentManifestModel`. We shouldn't fail deployment because somebody dumb fingered this file of mostly optional information. Especially since if we fail to deserialize the error will manifest in probably confusing ways in the toolkit.

src/AWS.Deploy.Orchestration/DeploymentManifestEngine.cs
```diff
@@ -0,0 +1,222 @@
...
+        }
+
+        /// <summary>
+        /// This method parses the <see cref="DeploymentManifestModel"/> into a string and writes it to disk.
+        /// </summary>
+        private async Task WriteDeploymentManifestFile(DeploymentManifestModel deploymentManifestModel)
+        {
+            var deploymentManifestFilePath = GetDeploymentManifestFilePath(_targetApplicationFullPath);
+            var manifestFileJsonString = JsonConvert.SerializeObject(deploymentManifestModel, Formatting.Indented);
```
normj, line 206: Use something like [SerializeRecipeContractResolver](https://github.com/aws/aws-dotnet-deploy/blob/dev/src/AWS.Deploy.Common/SerializeRecipeContractResolver.cs) to not serialize empty lists.

src/AWS.Deploy.Orchestration/Utilities/DeployedApplicationQueryer.cs
```diff
@@ -76,6 +89,19 @@ public async Task<List<CloudApplication>> GetExistingDeployedApplications(IList<
...
             }
 
+            if (_deploymentManifestEngine != null &&
+                _session != null)
+            {
+                var deploymentManifest = await _deploymentManifestEngine.GetDeploymentManifest();
+                var lastDeployedStack = deploymentManifest?.LastDeployedStacks
+                    .FirstOrDefault(x => x.AWSAccountId.Equals(_session.AWSAccountId) && x.AWSRegion.Equals(_session.AWSRegion));
+
+                return apps.OrderBy(x => {
```
normj, line 99: I believe the plan was to sort the list of the stacks whether they are in or not of the `LastDeployedStacks` and then the stacks last update date. That way the stack you most recently deployed to when show up first. We should also return the last updated date in case the toolkit wants to show it so the users can see the order is date based and not alpha.

normj (changes requested):

src/AWS.Deploy.CLI/Commands/DeleteDeploymentCommand.cs
philasmar: Originally I was doing that but since modifying the aws-deployments.json file needs knowledge of the project path and the delete command does not have it, I decided against it. The alternative would be to modify the delete command to only run in a directory with a valid project (which i don't think we want to do)

normj: Seems like we could do a best attempt. If the current directory or if `--project-path` is set then we can check the local deployment file. If the stack is not listed or the file is not found then we don't do any clean up. Currently the delete command does still take in `--project-path`.

philasmar: Done

src/AWS.Deploy.Common/DeploymentManifest/DeploymentManifestModel.cs
philasmar: Fixed

src/AWS.Deploy.Orchestration/DeploymentManifestEngine.cs
philasmar: If i dont explicitly check deploymentManifestModel for null, I will get a Possible null reference error.

src/AWS.Deploy.Orchestration/DeploymentManifestEngine.cs
philasmar: Fixed

src/AWS.Deploy.Orchestration/DeploymentManifestEngine.cs
philasmar: Fixed

src/AWS.Deploy.Orchestration/Utilities/DeployedApplicationQueryer.cs
philasmar: Fixed

src/AWS.Deploy.Orchestration/DeploymentManifestEngine.cs
```diff
@@ -0,0 +1,264 @@
...
+            try
+            {
+                if (!_directoryManager.Exists(saveCdkDirectoryFullPath))
+                    return;
+
+                var deploymentManifestModel = await GetDeploymentManifest();
+                var targetApplicationDirectoryPath = _directoryManager.GetDirectoryInfo(_targetApplicationFullPath).Parent.FullName;
+                var saveCdkDirectoryRelativePath = _directoryManager.GetRelativePath(targetApplicationDirectoryPath, saveCdkDirectoryFullPath);
+
+                if (deploymentManifestModel != null && deploymentManifestModel.DeploymentManifestEntries != null)
```
normj, line 65: If `deploymentManifestModel` is not null then you should always use that object not just if `deploymentManifestModel.DeploymentManifestEntries` is not null. Otherwise you are wiping out the rest of the data in the `deploymentManifestModel` the first time something is added to `DeploymentManifestEntries`.

src/AWS.Deploy.Orchestration/DeploymentManifestEngine.cs
```diff
@@ -0,0 +1,264 @@
...
+                if (_session.AWSAccountId == null || _session.AWSRegion == null)
+                    throw new FailedToUpdateDeploymentManifestFileException("The AWS Account Id or Region is not defined.");
+
+                var deploymentManifestModel = await GetDeploymentManifest();
+                var lastDeployedStack = deploymentManifestModel?.LastDeployedStacks?
+                    .FirstOrDefault(x => x.AWSAccountId.Equals(_session.AWSAccountId) && x.AWSRegion.Equals(_session.AWSRegion));
+
+                if (deploymentManifestModel != null && lastDeployedStack != null)
+                {
+                    if (lastDeployedStack != null)
```
normj, line 102: Since we are sorting by the CloudFormation update timestamp we don't need to keep changing the order of this file. I think that will cause frustration with a pair of developers both deploying to their own stacks in the same account but the file keeps getting changes and then being listed as a changed file for change control. If we wanted to have a deterministic  list just always keep the list in alpha order.

src/AWS.Deploy.Orchestration/DeploymentManifestEngine.cs
```diff
@@ -0,0 +1,264 @@
...
+        /// This method parses the <see cref="DeploymentManifestModel"/> into a string and writes it to disk.
+        /// </summary>
+        private async Task WriteDeploymentManifestFile(DeploymentManifestModel deploymentManifestModel)
+        {
+            var deploymentManifestFilePath = GetDeploymentManifestFilePath(_targetApplicationFullPath);
+            var manifestFileJsonString = JsonConvert.SerializeObject(deploymentManifestModel, new JsonSerializerSettings
+            {
+                Formatting = Formatting.Indented,
+                NullValueHandling = NullValueHandling.Ignore,
+                ContractResolver = new SerializeRecipeContractResolver()
```
normj, line 247: Should we change the name of this resolver class to something more generic since this is not a recipe file.

src/AWS.Deploy.CLI/Commands/DeployCommand.cs
```diff
@@ -18,6 +18,7 @@
 using AWS.Deploy.Orchestration.Data;
```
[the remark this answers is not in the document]
normj: Looking at this a bit more it seems like we have a divergence between how the CLI is determining the list and ServerMode. ServerMode is calling `GetExistingDeployedApplications` with the list of possible recommendations to filter. The CLI is calling `GetExistingDeployedApplications` without passing in a list of recommendations to filter and then calling a separate `GetCompatibleApplications` method to filter which is creating a new list and disrupting your sorting. We need to resolve this logic to have a single solution otherwise we are going to have bugs here.

normj (changes requested):

src/AWS.Deploy.Common/DeploymentManifest/DeploymentManifestModel.cs
```diff
@@ -11,11 +11,13 @@ namespace AWS.Deploy.Common.DeploymentManifest
     /// </summary>
     public class DeploymentManifestModel
     {
-        [JsonProperty("deployment-projects")]
-        public List<DeploymentManifestEntry> DeploymentManifestEntries { get; set; }
+        public List<LastDeployedStack>? LastDeployedStacks { get; set; }
 
-        public DeploymentManifestModel(List<DeploymentManifestEntry> deploymentManifestEntries)
+        public List<DeploymentManifestEntry>? DeploymentManifestEntries { get; set; }
```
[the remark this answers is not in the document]
96malhar: We already have JsonProperty called "deployment-projects". But its being removed as part of this PR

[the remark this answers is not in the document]
normj: @96malhar Yes, I asked in previous comment to make the name follow the same casing pattern. I'm good with using deployment project but we should probably use `DeploymentProjects` instead of `deployment-projects`.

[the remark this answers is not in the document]
philasmar: Fixed

src/AWS.Deploy.Common/DeploymentManifest/DeploymentManifestModel.cs
```diff
@@ -11,11 +11,13 @@ namespace AWS.Deploy.Common.DeploymentManifest
     /// </summary>
     public class DeploymentManifestModel
     {
-        [JsonProperty("deployment-projects")]
-        public List<DeploymentManifestEntry> DeploymentManifestEntries { get; set; }
+        public List<LastDeployedStack>? LastDeployedStacks { get; set; }
```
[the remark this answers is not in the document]
philasmar: Fixed

src/AWS.Deploy.Orchestration/DeploymentManifestEngine.cs
philasmar: Done

src/AWS.Deploy.CLI/Commands/CommandFactory.cs
```diff
@@ -171,6 +168,14 @@ private Command BuildDeployCommand()
 
                     var dockerEngine = new DockerEngine.DockerEngine(projectDefinition);
 
+                    var targetApplicationFullPath = _directoryManager.GetDirectoryInfo(projectDefinition.ProjectPath).FullName;
+
+                    var deploymentManifestEngine = new DeploymentManifestEngine(_directoryManager, _fileManager, session, targetApplicationFullPath);
+
+                    var deployedApplicationQueryer = new DeployedApplicationQueryer(_awsResourceQueryer, deploymentManifestEngine, session, _orchestratorInteractiveService);
+
+                    var customRecipeLocator = new CustomRecipeLocator(deploymentManifestEngine, _orchestratorInteractiveService, _commandLineWrapper, _directoryManager);
```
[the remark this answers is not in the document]
philasmar: Fixed

src/AWS.Deploy.CLI/Commands/CommandFactory.cs
```diff
@@ -256,7 +262,27 @@ private Command BuildDeleteCommand()
...
+                    {
+                        var projectDefinition = await _projectParserUtility.Parse(input.ProjectPath ?? "");
+
+                        var callerIdentity = await _awsResourceQueryer.GetCallerIdentity();
+
+                        var session = new OrchestratorSession(
+                        projectDefinition,
+                        awsCredentials,
+                        awsRegion,
+                        callerIdentity.Account);
```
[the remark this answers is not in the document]
philasmar: Fixed

src/AWS.Deploy.CLI/Commands/CommandFactory.cs
```diff
@@ -256,7 +262,27 @@ private Command BuildDeleteCommand()
...
+
+                        var targetApplicationFullPath = _directoryManager.GetDirectoryInfo(projectDefinition.ProjectPath).FullName;
+
+                        deploymentManifestEngine = new DeploymentManifestEngine(_directoryManager, _fileManager, session, targetApplicationFullPath);
+                    }
+                    catch (FailedToFindDeployableTargetException) { }
```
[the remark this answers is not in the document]
philasmar: This is to address Norm's comment. We want to do a best effort in deleting the stack from the file, but if the user does not provide the project path as part of the delete command, then we cannot do this. This is why we catch the exception.

test/AWS.Deploy.CLI.IntegrationTests/SaveCdkDeploymentProject/SaveCdkDeploymentProjectTests.cs
```diff
@@ -14,52 +15,60 @@ namespace AWS.Deploy.CLI.IntegrationTests.SaveCdkDeploymentProject
     [Collection("SaveCdkDeploymentProjectTests")]
     public class SaveCdkDeploymentProjectTests : IDisposable
     {
-        private readonly string _targetApplicationProjectPath;
-        private readonly string _deploymentManifestFilePath;
+        //private readonly string _targetApplicationProjectPath;
+        //private readonly string _deploymentManifestFilePath;
```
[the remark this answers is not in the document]
philasmar: Done

test/AWS.Deploy.CLI.IntegrationTests/SaveCdkDeploymentProject/SaveCdkDeploymentProjectTests.cs
```diff
@@ -14,52 +15,60 @@ namespace AWS.Deploy.CLI.IntegrationTests.SaveCdkDeploymentProject
...
         private string _saveDirectoryPath;
+        private string _stackName;
 
         public SaveCdkDeploymentProjectTests()
         {
             var testAppsDirectoryPath = Utilities.ResolvePathToTestApps();
-            _targetApplicationProjectPath = Path.Combine(testAppsDirectoryPath, "WebAppWithDockerFile");
-            _deploymentManifestFilePath = Path.Combine(testAppsDirectoryPath, "WebAppWithDockerFile", "aws-deployments.json");
+            //_targetApplicationProjectPath = Path.Combine(testAppsDirectoryPath, "WebAppWithDockerFile");
+            //_deploymentManifestFilePath = Path.Combine(testAppsDirectoryPath, "WebAppWithDockerFile", "aws-deployments.json");
```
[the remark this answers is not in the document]
philasmar: Done

src/AWS.Deploy.Orchestration/DeploymentManifestEngine.cs
```diff
@@ -0,0 +1,271 @@
...
+
+                var deploymentManifestModel = await GetDeploymentManifest();
+                var lastDeployedStack = deploymentManifestModel?.LastDeployedStacks?
+                    .FirstOrDefault(x => x.AWSAccountId.Equals(_session.AWSAccountId) && x.AWSRegion.Equals(_session.AWSRegion));
+
+                if (deploymentManifestModel != null)
+                {
+                    if (lastDeployedStack != null)
+                    {
+                        if (lastDeployedStack.Stacks == null)
```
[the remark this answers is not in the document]
philasmar: Since we are ignoring empty lists when serializing, this can be null when serializing.

ganeshnj:
General question: what is the rational behind storing the timestamp in aws-deployments.json file than using `LastUpdatedTime` to order all the stacks deployed through deploy tool?

This approach will be more accurate in case stack gets updated out of the lifecycle of deploy tool and requires less maintenance on client side.

ganeshnj (changes requested):

ganeshnj:
Another downside of this approach is, given aws-deployments.json file is source controlled, as soon as multiple developers will commit their changes, this logic will not work.

By keeping stack tracking per system (somewhere in temp location) would solve this problem.

src/AWS.Deploy.Orchestration/Utilities/DeployedApplicationQueryer.cs
```diff
@@ -73,7 +87,32 @@ public async Task<List<CloudApplication>> GetExistingDeployedApplications(IList<
                     continue;
                 }
 
-                apps.Add(new CloudApplication(stack.StackName, recipeId));
+                apps.Add(new CloudApplication(stack.StackName, recipeId, stack.LastUpdatedTime));
```
96malhar, line 90: What time zone is `stack.LastUpdatedTime` returned in? Also, as per the documentation, this field will only be returned if the stack has been updated at least once. 

image

source - https://docs.aws.amazon.com/sdkfornet/v3/apidocs/items/CloudFormation/TStack.html

Will this be a problem?

src/AWS.Deploy.CLI/Commands/CommandHandlerInput/GenerateDeploymentProjectCommandHandlerInput.cs
```diff
@@ -8,6 +8,8 @@ namespace AWS.Deploy.CLI.Commands.CommandHandlerInput
     /// </summary>
     public class GenerateDeploymentProjectCommandHandlerInput
     {
+        public string? Profile { get; set; }
+        public string? Region { get; set; }
```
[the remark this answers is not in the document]
philasmar: Fixed. No need for AWS creds for this command.

src/AWS.Deploy.Orchestration/DeploymentManifestEngine.cs
```diff
@@ -0,0 +1,282 @@
...
+        Task DeleteLastDeployedStack(string stackName, string? awsAccountId, string? awsRegion, string targetApplicationFullPath);
+        Task<List<string>> GetRecipeDefinitionPaths(string targetApplicationFullPath);
+        Task CleanOrphanStacks(List<string> deployedStacks, string? awsAccountId, string? awsRegion, string targetApplicationFullPath);
+    }
+
+    /// <summary>
+    /// This class contains the helper methods to update the deployment manifest file
+    /// that keeps track of the save CDK deployment projects.
+    /// </summary>
+    public class DeploymentManifestEngine : IDeploymentManifestEngine
```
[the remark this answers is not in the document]
philasmar: Added tests

src/AWS.Deploy.CLI/Commands/DeployCommand.cs
```diff
@@ -131,24 +140,23 @@ public async Task<(Orchestrator, Recommendation, CloudApplication)> InitializeDe
...
                     _customRecipeLocator,
-                    new List<string> { RecipeLocator.FindRecipeDefinitionsPath() });
+                    new List<string> { RecipeLocator.FindRecipeDefinitionsPath() },
+                    _directoryManager);
 
             // Determine what recommendations are possible for the project.
             var recommendations = await GenerateDeploymentRecommendations(orchestrator, deploymentProjectPath);
 
-            // Get all existing applications that were previously deployed using our deploy tool.
-            var allDeployedApplications = await _deployedApplicationQueryer.GetExistingDeployedApplications();
```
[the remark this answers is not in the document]
philasmar: I added that back to get all stacks.

src/AWS.Deploy.CLI/Commands/DeployCommand.cs
```diff
@@ -131,24 +140,23 @@ public async Task<(Orchestrator, Recommendation, CloudApplication)> InitializeDe
...
-
             // Filter compatible applications that can be re-deployed  using the current set of recommendations.
-            var compatibleApplications = GetCompatibleApplications(allDeployedApplications, recommendations);
+            var compatibleApplications = await _deployedApplicationQueryer.GetExistingDeployedApplications(recommendations, _session);
 
             // Get Cloudformation stack name.
             var cloudApplicationName = GetCloudApplicationName(stackName, userDeploymentSettings, compatibleApplications);
 
             // Find existing application with the same CloudFormation stack name.
-            var deployedApplication = allDeployedApplications.FirstOrDefault(x => string.Equals(x.Name, cloudApplicationName));
```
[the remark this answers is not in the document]
philasmar: Fixed.

src/AWS.Deploy.CLI/Commands/DeployCommand.cs
```diff
@@ -180,23 +188,6 @@ public async Task<(Orchestrator, Recommendation, CloudApplication)> InitializeDe
...
-        {
-            var compatibleApplications = new List<CloudApplication>();
-            foreach (var app in allDeployedApplications)
-            {
-                if (recommendations.Any(rec => string.Equals(rec.Recipe.Id, app.RecipeId, StringComparison.Ordinal)))
-                    compatibleApplications.Add(app);
-            }
-            return compatibleApplications;
-        }
-
```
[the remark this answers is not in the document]
philasmar: This is not needed. It is done as part of GetExistingDeployedApplications

src/AWS.Deploy.CLI/Extensions/CustomServiceCollectionExtension.cs
```diff
@@ -5,12 +5,12 @@
 using AWS.Deploy.CLI.Commands.TypeHints;
 using AWS.Deploy.CLI.Utilities;
 using AWS.Deploy.Common;
-using AWS.Deploy.Common.DeploymentManifest;
 using AWS.Deploy.Common.Extensions;
 using AWS.Deploy.Common.IO;
 using AWS.Deploy.Orchestration;
 using AWS.Deploy.Orchestration.CDK;
 using AWS.Deploy.Orchestration.Data;
+using AWS.Deploy.Orchestration.DeploymentManifest;
```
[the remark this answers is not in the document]
philasmar: DeploymentManifestEntry and DeploymentManifestModel are POCO classes and could be used by other projects. They dont NEED to be in the orchestrator.

src/AWS.Deploy.CLI/ServerMode/Controllers/DeploymentController.cs
```diff
@@ -242,13 +244,15 @@ public async Task<IActionResult> GetExistingDeployments(string sessionId)
             var output = new GetExistingDeploymentsOutput();
 
             var deployedApplicationQueryer = serviceProvider.GetRequiredService<IDeployedApplicationQueryer>();
-            state.ExistingDeployments = await deployedApplicationQueryer.GetExistingDeployedApplications(state.NewRecommendations);
+            var session = CreateOrchestratorSession(state);
+            state.ExistingDeployments = await deployedApplicationQueryer.GetExistingDeployedApplications(state.NewRecommendations, session);
```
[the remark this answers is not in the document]
philasmar: This is what this API is meant to do.

src/AWS.Deploy.Orchestration/Utilities/DeployedApplicationQueryer.cs
philasmar: I checked and every stack deployed by the deploy tool has the Last Updated Time. I also tried to fetch them with the API and it was fine.

src/AWS.Deploy.CLI/Commands/DeployCommand.cs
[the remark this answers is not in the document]
96malhar: Move this method to `DeployedApplicationQueryer`  and use this to filter the `compatibleApplications` from `allDeployedApplications`. This way we dont need to make a second API call to CloudFormation.

test/AWS.Deploy.CLI.IntegrationTests/SaveCdkDeploymentProject/CustomRecipeLocatorTests.cs
```diff
@@ -51,7 +52,7 @@ public CustomRecipeLocatorTests()
             _solutionDirectoryPath = directoryManager.GetDirectoryInfo(solutionPath).Parent.FullName;
         }
 
-        [Fact]
+        [Fact(Skip = "Needs Fixing")]
```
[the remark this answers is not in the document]
philasmar: Done

src/AWS.Deploy.CLI/Commands/DeployCommand.cs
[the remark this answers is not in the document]
philasmar: Done

src/AWS.Deploy.Common/DeploymentManifest/DeploymentManifestEngine.cs
```diff
@@ -57,7 +57,14 @@ public async Task UpdateDeploymentManifestFile(string saveCdkDirectoryFullPath,
                 if (_fileManager.Exists(deploymentManifestFilePath))
                 {
                     deploymentManifestModel = await ReadManifestFile(deploymentManifestFilePath);
-                    deploymentManifestModel.DeploymentManifestEntries.Add(new DeploymentManifestEntry(saveCdkDirectoryRelativePath));
+                    if (deploymentManifestModel.DeploymentProjects == null)
+                    {
+                        deploymentManifestModel.DeploymentProjects = new List<DeploymentManifestEntry> { new DeploymentManifestEntry(saveCdkDirectoryRelativePath) };
+                    }
```
ganeshnj, line 63: This and L100 check look redundant, `DeploymentManifestModel` has constructor which always initializes with a non-null List<DeploymentManifestEntry>.

src/AWS.Deploy.Orchestration/LocalUserSettings/LocalUserSettingsEngine.cs
```diff
@@ -0,0 +1,208 @@
...
+
+                var localUserSettings = await GetLocalUserSettings();
+                var lastDeployedStack = localUserSettings?.LastDeployedStacks?
+                    .FirstOrDefault(x => x.Exists(awsAccountId, awsRegion, projectName));
+
+                if (localUserSettings != null)
+                {
+                    if (lastDeployedStack != null)
+                    {
+                        if (lastDeployedStack.Stacks == null)
```
ganeshnj, line 57: It is redundant check, `LastDeployedStack` has constructor that initializes Stacks with a non-nullable List<string>

src/AWS.Deploy.Common/CloudApplication.cs
```diff
@@ -27,16 +29,27 @@ public class CloudApplication
...
 
+        /// <summary>
+        /// Last updated time of CloudFormation stack
+        /// </summary>
+        public DateTime? LastUpdatedTime { get; set; }
+
+        /// <summary>
+        /// Indicates whether the Cloud Application has been redeployed from this machine
+        /// </summary>
+        public bool LocallyRedeployed { get; set; }
```
ganeshnj, line 40: This meant to be the current (logged in) user not "machine" in the comment. 

Also, it will make easier to understand if we rename it to `UpdatedByCurrentUser` IMO.

src/AWS.Deploy.Orchestration/LocalUserSettings/LocalUserSettingsEngine.cs
```diff
@@ -0,0 +1,208 @@
...
+                var localUserSettings = await GetLocalUserSettings();
+                var localStacks = localUserSettings?.LastDeployedStacks?
+                    .FirstOrDefault(x => x.Exists(awsAccountId, awsRegion, projectName));
+
+                if (localUserSettings == null || localStacks == null)
+                    return;
+
+                var validStacks = deployedStacks.Intersect(localStacks.Stacks);
+
+                localStacks.Stacks = validStacks.ToList();
```
ganeshnj, line 150: Where is `validStacks` used? Possibly both statements can be removed.
```suggestion
```

ganeshnj (changes requested):

src/AWS.Deploy.Common/DeploymentManifest/DeploymentManifestEngine.cs
philasmar: They are not redundant. When we are deserializing them, if they don't exist in the file, they will be null. Ive ran into null exceptions multiple times which is why i added the check.

src/AWS.Deploy.Orchestration/LocalUserSettings/LocalUserSettingsEngine.cs
philasmar: They are not redundant. When we are deserializing them, if they don't exist in the file, they will be null. Ive ran into null exceptions multiple times which is why i added the check.

src/AWS.Deploy.Common/CloudApplication.cs
philasmar: Done

src/AWS.Deploy.Orchestration/LocalUserSettings/LocalUserSettingsEngine.cs
philasmar: validStacks is being set to  localStacks.Stacks
So we are basically getting the intersection, setting it in the model and then writing the file back.

src/AWS.Deploy.Common/DeploymentManifest/DeploymentManifestEngine.cs
ganeshnj: That's interesting, there must be some config that we are missing here. We can check it later.

src/AWS.Deploy.Orchestration/LocalUserSettings/LocalUserSettingsEngine.cs
ganeshnj: Makes sense, it was hard to follow in one go.

ganeshnj (approved):

96malhar (approved):

PR #295, opened 2021-08-16
fix: Correctly resolve the Solution path on non-Windows targets

96malhar:
*Issue #, if available:*

*Description of changes:*
This PR correctly resolves the Solution path while parsing the project definition on non-Windows targets.

### DIFF
--- a/src/AWS.Deploy.Common/ProjectDefinitionParser.cs
+++ b/src/AWS.Deploy.Common/ProjectDefinitionParser.cs
@@ -126,7 +126,16 @@ namespace AWS.Deploy.Common
 
             var lines = await _fileManager.ReadAllLinesAsync(solutionFile);
             var projectLines = lines.Where(x => x.StartsWith("Project"));
-            var projectPaths = projectLines.Select(x => x.Split(',')[1].Replace('\"', ' ').Trim()).ToList();
+            var projectPaths =
+                projectLines
+                    .Select(x => x.Split(','))
+                    .Where(x => x.Length > 1)
+                    .Select(x =>
+                            x[1]
+                                .Replace('\"', ' ')
+                                .Trim())
+                    .Select(x => x.Replace('\\', Path.DirectorySeparatorChar))
+                    .ToList();
 
             //Validate project exists in solution
             return projectPaths.Select(x => Path.GetFileName(x)).Any(x => x.Equals(projectFileName));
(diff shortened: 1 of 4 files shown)


philasmar (approved):

src/AWS.Deploy.Common/ProjectDefinitionParser.cs
```diff
@@ -126,7 +126,9 @@ private async Task<bool> ValidateProjectInSolution(string projectPath, string so
 
             var lines = await _fileManager.ReadAllLinesAsync(solutionFile);
             var projectLines = lines.Where(x => x.StartsWith("Project"));
-            var projectPaths = projectLines.Select(x => x.Split(',')[1].Replace('\"', ' ').Trim()).ToList();
+            var projectPaths = projectLines.Select(x => x.Split(',')[1].Replace('\"', ' ').Trim())
+                .Select(x => x.Replace('\\', Path.DirectorySeparatorChar))
+                .ToList();
```
ganeshnj, line 131: We should remove this method than fixing  and start using https://github.com/aws/aws-dotnet-deploy/blob/cc7cbcffb353d41aa2c73ead71468d66c80ab72c/src/AWS.Deploy.Orchestration/DeploymentBundleHandler.cs#L195

ganeshnj (changes requested):

src/AWS.Deploy.Common/ProjectDefinitionParser.cs
```diff
@@ -126,7 +126,9 @@ private async Task<bool> ValidateProjectInSolution(string projectPath, string so
 
             var lines = await _fileManager.ReadAllLinesAsync(solutionFile);
             var projectLines = lines.Where(x => x.StartsWith("Project"));
-            var projectPaths = projectLines.Select(x => x.Split(',')[1].Replace('\"', ' ').Trim()).ToList();
+            var projectPaths = projectLines.Select(x => x.Split(',')[1].Replace('\"', ' ').Trim())
```
normj, line 129: I think we should be defensive an not always assume there is a `1` element. I know that is not new in this PR.

normj (approved):
Approved but I suspect we are missing some unit tests for this function.

src/AWS.Deploy.Common/ProjectDefinitionParser.cs
96malhar: This method is redundant. So removed it.

ganeshnj (approved):

normj (approved):

philasmar (approved):

PR #298, opened 2021-08-16
feat: detect CDK version by parsing deployment project csproj file

ganeshnj:
*Issue #, if available:*

*Description of changes:*
Currently, Deployment Tool uses static CDK version which needs to be changed as Amazon.CDK.* package references are updated. This change allows Deployment Tool to detect the minimum compatible version of Amazon.CDK.* dependency which must be installed or will be installed by the tool.

### DIFF
--- a/src/AWS.Deploy.CLI/Commands/CommandFactory.cs
+++ b/src/AWS.Deploy.CLI/Commands/CommandFactory.cs
@@ -2,6 +2,7 @@
 // SPDX-License-Identifier: Apache-2.0
 
 using System;
+using System.Collections;
 using System.CommandLine;
 using System.CommandLine.Invocation;
 using System.IO;
@@ -67,6 +68,7 @@ namespace AWS.Deploy.CLI.Commands
         private readonly IDeploymentManifestEngine _deploymentManifestEngine;
         private readonly ICustomRecipeLocator _customRecipeLocator;
         private readonly ILocalUserSettingsEngine _localUserSettingsEngine;
+        private readonly ICDKVersionDetector _cdkVersionDetector;
 
         public CommandFactory(
             IToolInteractiveService toolInteractiveService,
@@ -90,7 +92,8 @@ namespace AWS.Deploy.CLI.Commands
             IFileManager fileManager,
             IDeploymentManifestEngine deploymentManifestEngine,
             ICustomRecipeLocator customRecipeLocator,
-            ILocalUserSettingsEngine localUserSettingsEngine)
+            ILocalUserSettingsEngine localUserSettingsEngine,
+            ICDKVersionDetector cdkVersionDetector)
         {
             _toolInteractiveService = toolInteractiveService;
             _orchestratorInteractiveService = orchestratorInteractiveService;
@@ -114,6 +117,7 @@ namespace AWS.Deploy.CLI.Commands
             _deploymentManifestEngine = deploymentManifestEngine;
             _customRecipeLocator = customRecipeLocator;
             _localUserSettingsEngine = localUserSettingsEngine;
+            _cdkVersionDetector = cdkVersionDetector;
         }
 
         public Command BuildRootCommand()
@@ -193,6 +197,7 @@ namespace AWS.Deploy.CLI.Commands
                         _orchestratorInteractiveService,
                         _cdkProjectHandler,
                         _cdkManager,
+                        _cdkVersionDetector,
                         _deploymentBundleHandler,
                         dockerEngine,
                         _awsResourceQueryer,
@@ -205,7 +210,8 @@ namespace AWS.Deploy.CLI.Commands
                         _consoleUtilities,
                         _customRecipeLocator,
                         _systemCapabilityEvaluator,
-                        session);
+                        session,
+                        _directoryManager);
 
                     var deploymentProjectPath = input.DeploymentProject ?? string.Empty;
                     if (!string.IsNullOrEmpty(deploymentProjectPath))
@@ -338,7 +344,7 @@ namespace AWS.Deploy.CLI.Commands
                 listCommand.Add(_optionRegion);
                 listCommand.Add(_optionProjectPath);
                 listCommand.Add(_optionDiagnosticLogging);
-            } 
+            }
 
             listCommand.Handler = CommandHandler.Create(async (ListCommandHandlerInput input) =>
             {
--- a/src/AWS.Deploy.CLI/Commands/DeployCommand.cs
+++ b/src/AWS.Deploy.CLI/Commands/DeployCommand.cs
@@ -42,12 +42,15 @@ namespace AWS.Deploy.CLI.Commands
         private readonly ICustomRecipeLocator _customRecipeLocator;
         private readonly ISystemCapabilityEvaluator _systemCapabilityEvaluator;
         private readonly OrchestratorSession _session;
+        private readonly IDirectoryManager _directoryManager;
+        private ICDKVersionDetector _cdkVersionDetector;
 
         public DeployCommand(
             IToolInteractiveService toolInteractiveService,
             IOrchestratorInteractiveService orchestratorInteractiveService,
             ICdkProjectHandler cdkProjectHandler,
             ICDKManager cdkManager,
+            ICDKVersionDetector cdkVersionDetector,
             IDeploymentBundleHandler deploymentBundleHandler,
             IDockerEngine dockerEngine,
             IAWSResourceQueryer awsResourceQueryer,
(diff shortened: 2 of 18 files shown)


src/AWS.Deploy.CLI/Commands/CommandFactory.cs
```diff
@@ -308,7 +312,7 @@ private Command BuildListCommand()
                 listCommand.Add(_optionRegion);
                 listCommand.Add(_optionProjectPath);
                 listCommand.Add(_optionDiagnosticLogging);
-            } 
+            }
```
philasmar, line 315: nit: can you remove whitespace change?

src/AWS.Deploy.CLI/Commands/DeployCommand.cs
```diff
@@ -162,7 +166,7 @@ public async Task<(Orchestrator, Recommendation, CloudApplication)> InitializeDe
 
                 // preset settings for deployment based on last deployment.
                 selectedRecommendation = await GetSelectedRecommendationFromPreviousDeployment(recommendations, deployedApplication, userDeploymentSettings);
-            } 
+            }
```
philasmar, line 169: there are several whitespace changes in this class. what is causing them?

src/AWS.Deploy.Orchestration/CDK/CDKVersionDetector.cs
```diff
@@ -0,0 +1,87 @@
...
+
+            foreach (var node in document.DescendantNodes())
+            {
+                if (node is not XElement element || element.Name.ToString() != "PackageReference")
+                {
+                    continue;
+                }
+
+                var includeAttribute = element.Attribute("Include");
+                if (includeAttribute == null || !includeAttribute.Value.StartsWith("Amazon.CDK."))
```
philasmar, line 50: Could you add "Amazon.CDK." as a constant? Also isn't there a Nuget package called "Amazon.CDK" (without the . )? This won't work for that specific package.

src/AWS.Deploy.Orchestration/Orchestrator.cs
```diff
@@ -131,16 +134,22 @@ public async Task DeployRecommendation(CloudApplication cloudApplication, Recomm
...
             {
                 case DeploymentTypes.CdkProject:
-                    await _cdkProjectHandler.CreateCdkDeployment(_session, cloudApplication, recommendation);
+                    var cdkProject = await _cdkProjectHandler.ConfigureCdkProject(_session, cloudApplication, recommendation);
+                    _interactiveService.LogMessageLine("AWS CDK is being configured.");
+
+                    if (_cdkVersionDetector == null)
+                    {
+                        throw new InvalidOperationException($"{nameof(_cdkVersionDetector)} must not be null.");
+                    }
```
philasmar, line 146: can we do this check earlier and error out before we configure the cdk project.

src/AWS.Deploy.CLI/Commands/CommandFactory.cs
ganeshnj: I would rather keep them, there shouldn't be a whitespace once the line has ended.

src/AWS.Deploy.CLI/Commands/DeployCommand.cs
ganeshnj: same as above.

src/AWS.Deploy.Orchestration/CDK/CDKVersionDetector.cs
ganeshnj: Fair call, Amazon.CDK would be a better check.

src/AWS.Deploy.Orchestration/Orchestrator.cs
ganeshnj: nope that's not going to work for following reasons, we need csproj file in order to detect which is only available once the deployment is generated. It could be in temp dir or customer codebase.

src/AWS.Deploy.Orchestration/CdkProjectHandler.cs
```diff
@@ -53,23 +48,34 @@ public async Task CreateCdkDeployment(OrchestratorSession session, CloudApplicat
...
+        }
+
+        public async Task DeployCdkProject(OrchestratorSession session, string cdkProjectPath, Recommendation recommendation)
+        {
+            var recipeInfo = $"{recommendation.Recipe.Id}_{recommendation.Recipe.Version}";
+            var environmentVariables = new Dictionary<string, string>
             {
-                await appSettingsFile.WriteAsync(appSettingsBody);
-            }
+                { EnvironmentVariableKeys.AWS_EXECUTION_ENV, recipeInfo }
```
normj, line 68: We are messing up the execution env which is configured `App.cs`. It should always start with the current value set for execution env which in the CLI would be `aws-dotnet-deploy-cli` and then we can add recipe info as additional information.

src/AWS.Deploy.Orchestration/CdkProjectHandler.cs
```diff
@@ -53,23 +48,34 @@ public async Task CreateCdkDeployment(OrchestratorSession session, CloudApplicat
...
+            };
 
             _interactiveService.LogMessageLine("Starting deployment of CDK Project");
 
             // Ensure region is bootstrapped
             await _commandLineWrapper.Run($"npx cdk bootstrap aws://{session.AWSAccountId}/{session.AWSRegion}",
                 needAwsCredentials: true);
 
+            var appSettingsFilePath = Path.Combine(cdkProjectPath, "appsettings.json");
```
normj, line 77: If this is a saved deployment project we should still be writing the `appsettings.json` file to `~/.aws-dotnet-deploy/<random-folder>`. We don't want to be writing a temporary file into the user's source control folder.

src/AWS.Deploy.Orchestration/Orchestrator.cs
```diff
@@ -131,16 +134,22 @@ public async Task DeployRecommendation(CloudApplication cloudApplication, Recomm
...
+
+                    var csprojFiles = Directory.GetFiles(cdkProject, "*.csproj") ?? new string[]{};
```
normj, line 148: Can we move this into a utility method to get a project file for a directory and then also look for fsproj files? I know it is highly unlikely that somebody will write a CDK project in F# but I want us to in the practice for always looking for either type of project file.

normj (changes requested):

src/AWS.Deploy.Orchestration/Orchestrator.cs
philasmar: I'm not sure i get what you mean. _cdkVersionDetector will be null on line 140 and 143. So instead of checking it on 143, I'd like to check it on before 140

src/AWS.Deploy.Orchestration/CdkProjectHandler.cs
ganeshnj: This is already taken care [here](https://github.com/aws/aws-dotnet-deploy/blob/4c541282dce9a3582afcfe1621b99b6253947e45/src/AWS.Deploy.CLI/Utilities/CommandLineWrapper.cs#L127)

src/AWS.Deploy.Orchestration/CdkProjectHandler.cs
ganeshnj: This should be a separate task, this change is required in order to split the `CreateCdkDeployment` into two parts which are `ConfigureCdkProject` and `DeployCdkProject`.

src/AWS.Deploy.Orchestration/Orchestrator.cs
ganeshnj: I don't like that either, why even dependencies are marked as nullable which is the root cause of the problem here. Moving the code two lines above does not solve the problem.

src/AWS.Deploy.Orchestration/Orchestrator.cs
ganeshnj: 100%

ganeshnj: Fixed

src/AWS.Deploy.Orchestration/Orchestrator.cs
ganeshnj: Moved to as first line to unblock this PR. We have backlog now DOTNET-5358.

src/AWS.Deploy.Orchestration/CDK/CDKVersionDetector.cs
```diff
@@ -47,7 +49,7 @@ public Version Detect(string csprojPath)
                 }
 
                 var includeAttribute = element.Attribute("Include");
-                if (includeAttribute == null || !includeAttribute.Value.StartsWith("Amazon.CDK."))
+                if (includeAttribute == null || !includeAttribute.Value.StartsWith(AMAZON_CDK_PACKAGE_REFERENCE_PREFIX))
```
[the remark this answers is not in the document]
ganeshnj: Fixed and added test case too.

normj (approved):

philasmar (approved):

PR #300, opened 2021-08-17
feat: check if system capabilities are missing before deployment in server mode

philasmar:
*Issue #, if available:*

*Description of changes:*
feat: check if system capabilities are missing before deployment in server mode

### DIFF
--- a/src/AWS.Deploy.CLI/ServerMode/Controllers/DeploymentController.cs
+++ b/src/AWS.Deploy.CLI/ServerMode/Controllers/DeploymentController.cs
@@ -353,7 +353,7 @@ namespace AWS.Deploy.CLI.ServerMode.Controllers
         [HttpPost("session/<sessionId>/execute")]
         [SwaggerOperation(OperationId = "StartDeployment")]
         [Authorize]
-        public IActionResult StartDeployment(string sessionId)
+        public async Task<IActionResult> StartDeployment(string sessionId)
         {
             var state = _stateServer.Get(sessionId);
             if (state == null)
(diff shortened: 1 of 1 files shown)


src/AWS.Deploy.CLI/ServerMode/Controllers/DeploymentController.cs
```diff
@@ -368,6 +368,13 @@ public IActionResult StartDeployment(string sessionId)
             if (state.SelectedRecommendation == null)
                 throw new SelectedRecommendationIsNullException("The selected recommendation is null or invalid.");
 
+            var systemCapabilityEvaluator = serviceProvider.GetRequiredService<ISystemCapabilityEvaluator>();
+
+            var capabilities = await systemCapabilityEvaluator.EvaluateSystemCapabilities(state.SelectedRecommendation);
+
+            if (capabilities.Any())
+                return Problem("Unable to start deployment due to missing system capabilities.");
```
normj, line 376: We should return back what are the missing capabilities in the error message.

normj (changes requested):

src/AWS.Deploy.CLI/ServerMode/Controllers/DeploymentController.cs
ganeshnj: I agree with @normj assuming this message will be shown in some window/label/alert and we would like to be very specific here so that customer can take appropriate action.

philasmar: Updated

normj (approved):

ganeshnj (approved):

PR #304, opened 2021-08-20
feat: make the encryption key info mode the default one

philasmar:
*Issue #, if available:*
IDE-5674

*Description of changes:*
ZX Security found that the mechanism that configures the Authorization header’s
encryption is entirely optional, and not enabled by default. The following is a valid command, even though it lacks the --encryption-keyinfo-stdin parameter:
dotnet aws server-mode --port 10000 --parent-pid 1512
In this mode, the Authorization header does not get encrypted, and can be easily
read or forged.

This PR makes the encryption key info mode the default one


src/AWS.Deploy.CLI/Commands/CommandFactory.cs
```diff
@@ -485,6 +485,7 @@ private Command BuildServerModeCommand()
                 serverModeCommand.Add(new Option<int>(new[] { "--port" }, description: "Port the server mode will listen to."));
                 serverModeCommand.Add(new Option<int>(new[] { "--parent-pid" }, description: "The ID of the process that is launching server mode. Server mode will exit when the parent pid terminates."));
                 serverModeCommand.Add(new Option<bool>(new[] { "--encryption-keyinfo-stdin" }, description: "If set the cli reads encryption key info from stdin to use for decryption."));
+                serverModeCommand.Add(new Option<bool>(new[] { "--no-encryption-keyinfo" }, description: "If set the cli bypasses the need for an encryption key."));
```
ganeshnj, line 488: `--encryption-keyinfo-stdin` and `--no-encryption-keyinfo` are too confusing now given both are bool options. From the description, I couldn't understand what will be the end result.

Not related to this change but
```
"If set the cli reads encryption key info from stdin to use for decryption."
```
"encryption key" looks wrong, either it should be a symmetric or asymmetric key.

src/AWS.Deploy.CLI/Commands/ServerModeCommand.cs
```diff
@@ -117,7 +119,14 @@ private IEncryptionProvider CreateEncryptionProvider()
             }
             else
             {
-                encryptionProvider = new NoEncryptionProvider();
+                if (!_noEncryptionKeyInfo)
+                {
+                    throw new InvalidEncryptionKeyInfoException($"The encryption key info has not been passed to Stdin.");
```
ganeshnj, line 124: Same comment as above on "encryption key"

ganeshnj (changes requested):

src/AWS.Deploy.CLI/Commands/CommandFactory.cs
normj: I would be fine and I think that would meet AppSec requirement if we remove `--encryption-keyinfo-stdin` and always read from stdin unless `--no-encryption-keyinfo` is set. I would also rename `--no-encryption-keyinfo` to be `--unsecure-mode`. Encryption keyinfo is kind of an implementation detail and the point we want to put out is they are choosing to a less secure mode..

philasmar: Fixed

src/AWS.Deploy.CLI/Commands/ServerModeCommand.cs
philasmar: Fixed

PR #305, opened 2021-08-23
fix: sanitize deploy tool path to prevent command injection

philasmar:
*Issue #, if available:*
IDE-5675
*Description of changes:*
Pentest recommendation: ensure the CLI path is properly escaped before being passed to the Process.Start() method in .NET. File paths should be properly framed so as to avoid spaces being interpreted as command line syntax rather than part of the file path.


src/AWS.Deploy.ServerMode.Client/ServerModeSession.cs
```diff
@@ -216,6 +217,28 @@ private Task<bool> IsServerAvailable(CancellationToken cancellationToken)
...
+                return false;
+
+            if (Path.GetInvalidPathChars().Where(x => deployToolPath.Contains(x)).Any())
+                return false;
+
+            if (!File.Exists(_deployToolPath))
+                return false;
+
+            return true;
+        }
```
ganeshnj, line 240: This is generic method and should be part of Path class. It can be added as extension.
```
bool Path.IsValid(string? path)
```

ganeshnj (changes requested):

src/AWS.Deploy.ServerMode.Client/ServerModeSession.cs
philasmar: Path is a static class and cannot add extension methods to. I created a utilities class and put the method there.

PR #311, opened 2021-08-27
fix: disallow SMB shares from file paths

philasmar:
*Issue #, if available:*
IDE-5678

*Description of changes:*
Pentest recommendation:

Disallow SMB shares from being considered valid file paths by default.
If SMB shares are required, consider allowing specific servers to be explicitly allowed in a configuration file, and deny all other file paths.

### DIFF
--- a/src/AWS.Deploy.Common/IO/DirectoryManager.cs
+++ b/src/AWS.Deploy.Common/IO/DirectoryManager.cs
@@ -5,6 +5,7 @@ using System;
 using System.Collections.Generic;
 using System.IO;
 using System.Linq;
+using AWS.Deploy.Common.Utilities;
 
 namespace AWS.Deploy.Common.IO
 {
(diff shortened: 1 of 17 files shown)


src/AWS.Deploy.Common/IO/DirectoryManager.cs
```diff
@@ -61,5 +61,27 @@ public string[] GetProjFiles(string path)
         {
             return Directory.GetFiles(path).Where(filePath => _projFileExtensions.Contains(Path.GetExtension(filePath).ToLower())).ToArray();
         }
+
+        private bool IsDirectoryValid(string directoryPath)
+        {
+            directoryPath = directoryPath.Trim();
```
normj, line 67: Since this logic is the same as `IsFileValid` except for the Exists call can we put all of the string checks into a separate utility function that both methods call.

normj (changes requested):

src/AWS.Deploy.Common/IO/DirectoryManager.cs
philasmar: Done

normj (approved):

ganeshnj (approved):

PR #318, opened 2021-09-02
feat: Enable re-deployment using a compatible CDK deployment project.

96malhar:
*Issue #, if available:*
DOTNET-5328

*Description of changes:*
This PR enables re-deployments to an existing Cloudformation stack using a compatible CDK project

A sample workflow of this feature is as follows:
1. Perform an initial deployment using the **ECS Fargate recipe**.
1.1 `dotnet aws deploy --project-path [path-to-project] --stack-name [stack-name]`.
2. Create a persisted CDK deployment project using the **ECS Fargate recipe**.
2.1 `dotnet aws deployment-project generate --project-path [path-to-project] --output [path-to-save-cdk-project]`
3. Perform a re-deployment to the **same stack** using the persisted CDK project.
3.1 `dotnet aws deploy --project-path [path-to-project] --stack-name [stack-name] --deployment-project [path-to-saved-cdk-project]`

To enable this feature, a new property called `BaseRecipeId` is added as part of the `RecipeDefinition`. This property is set while generating the recipe snapshot when saving a custom CDK project to customer's source control. 

A re-deployment using the custom CDK project is only permitted when the `BaseRecipeId` is equal to the RecipeId used during the initial deployment. 

### DIFF
--- a/src/AWS.Deploy.CLI/Commands/DeployCommand.cs
+++ b/src/AWS.Deploy.CLI/Commands/DeployCommand.cs
@@ -20,6 +20,7 @@ using AWS.Deploy.Orchestration.Utilities;
 using AWS.Deploy.Orchestration.DisplayedResources;
 using AWS.Deploy.Common.IO;
 using AWS.Deploy.Orchestration.LocalUserSettings;
+using Newtonsoft.Json;
 
 namespace AWS.Deploy.CLI.Commands
 {
@@ -45,6 +46,7 @@ namespace AWS.Deploy.CLI.Commands
         private readonly ISystemCapabilityEvaluator _systemCapabilityEvaluator;
         private readonly OrchestratorSession _session;
         private readonly IDirectoryManager _directoryManager;
+        private readonly IFileManager _fileManager;
         private readonly ICDKVersionDetector _cdkVersionDetector;
 
         public DeployCommand(
@@ -66,7 +68,8 @@ namespace AWS.Deploy.CLI.Commands
             ICustomRecipeLocator customRecipeLocator,
             ISystemCapabilityEvaluator systemCapabilityEvaluator,
             OrchestratorSession session,
-            IDirectoryManager directoryManager)
+            IDirectoryManager directoryManager,
+            IFileManager fileManager)
         {
             _toolInteractiveService = toolInteractiveService;
             _orchestratorInteractiveService = orchestratorInteractiveService;
@@ -83,6 +86,7 @@ namespace AWS.Deploy.CLI.Commands
             _consoleUtilities = consoleUtilities;
             _session = session;
             _directoryManager = directoryManager;
+            _fileManager = fileManager;
             _cdkVersionDetector = cdkVersionDetector;
             _cdkManager = cdkManager;
             _customRecipeLocator = customRecipeLocator;
@@ -174,12 +178,12 @@ namespace AWS.Deploy.CLI.Commands
                 // Verify that the target application can be deployed using the current set of recommendations
                 if (!compatibleApplications.Any(app => app.StackName.Equals(deployedApplication.StackName, StringComparison.Ordinal)))
                 {
-                    var errorMessage = $"{deployedApplication.StackName} already exists as a Cloudformation stack but a compatible recommendation to perform a redeployment was no found";
+                    var errorMessage = $"{deployedApplication.StackName} already exists as a Cloudformation stack but a compatible recommendation to perform a redeployment was not found";
                     throw new FailedToFindCompatibleRecipeException(errorMessage);
                 }
 
                 // preset settings for deployment based on last deployment.
-                selectedRecommendation = await GetSelectedRecommendationFromPreviousDeployment(recommendations, deployedApplication, userDeploymentSettings);
+                selectedRecommendation = await GetSelectedRecommendationFromPreviousDeployment(recommendations, deployedApplication, userDeploymentSettings, deploymentProjectPath);
             }
             else
             {
@@ -264,15 +268,14 @@ namespace AWS.Deploy.CLI.Commands
             return recommendations;
         }
 
-        private async Task<Recommendation> GetSelectedRecommendationFromPreviousDeployment(List<Recommendation> recommendations, CloudApplication deployedApplication, UserDeploymentSettings? userDeploymentSettings)
+        private async Task<Recommendation> GetSelectedRecommendationFromPreviousDeployment(List<Recommendation> recommendations, CloudApplication deployedApplication, UserDeploymentSettings? userDeploymentSettings, string deploymentProjectPath)
         {
             var existingCloudApplicationMetadata = await _templateMetadataReader.LoadCloudApplicationMetadata(deployedApplication.Name);
             var deploymentSettingRecipeId = userDeploymentSettings?.RecipeId;
-            var selectedRecommendation = recommendations.FirstOrDefault(x => string.Equals(x.Recipe.Id, deployedApplication.RecipeId, StringComparison.InvariantCultureIgnoreCase));
-
+            var selectedRecommendation = await GetRecommendationForRedeployment(recommendations, deployedApplication, deploymentProjectPath);
             if (selectedRecommendation == null)
             {
-                var errorMessage = $"{deployedApplication.StackName} already exists as a Cloudformation stack but the recommendation used to deploy to the stack was not found.";
+                var errorMessage = $"{deployedApplication.StackName} already exists as a Cloudformation stack but a compatible recommendation used to perform a re-deployment was not found.";
                 throw new FailedToFindCompatibleRecipeException(errorMessage);
             }
             if (!string.IsNullOrEmpty(deploymentSettingRecipeId) && !string.Equals(deploymentSettingRecipeId, selectedRecommendation.Recipe.Id, StringComparison.InvariantCultureIgnoreCase))
@@ -315,6 +318,51 @@ namespace AWS.Deploy.CLI.Commands
             }
 
             return selectedRecommendation;
+        }
+
+        private async Task<Recommendation?> GetRecommendationForRedeployment(List<Recommendation> recommendations, CloudApplication deployedApplication, string deploymentProjectPath)
+        {
+            var targetRecipeId = !string.IsNullOrEmpty(deploymentProjectPath) ?
+                await GetDeploymentProjectRecipeId(deploymentProjectPath) : deployedApplication.RecipeId;
+
+            foreach (var recommendation in recommendations)
+            {
+                if (string.Equals(recommendation.Recipe.Id, targetRecipeId) && _deployedApplicationQueryer.IsCompatible(deployedApplication, recommendation))
+                {
+                    return recommendation;
+                }
+            }
+            return null;
+        }
+
+        private async Task<string> GetDeploymentProjectRecipeId(string deploymentProjectPath)
+        {
+            if (!_directoryManager.Exists(deploymentProjectPath))
+            {
+                throw new InvalidOperationException($"Invalid deployment project path. {deploymentProjectPath} does not exist on the file system.");
+            }
+
+            try
+            {
+                var recipeFiles = _directoryManager.GetFiles(deploymentProjectPath, "*.recipe");
+                if (recipeFiles.Length == 0)
+                {
+                    throw new InvalidOperationException($"Failed to find a recipe file at {deploymentProjectPath}");
+                }
+                if (recipeFiles.Length > 1)
+                {
+                    throw new InvalidOperationException($"Found more than one recipe files at {deploymentProjectPath}. Only one recipe file per deployment project is supported.");
+                }
+
+                var recipeFilePath = recipeFiles.First();
+                var recipeBody = await _fileManager.ReadAllTextAsync(recipeFilePath);
+                var recipe = JsonConvert.DeserializeObject<RecipeDefinition>(recipeBody);
+                return recipe.Id;
+            }
+            catch (Exception ex)
+            {
+                throw new FailedToFindDeploymentProjectRecipeIdException($"Failed to find a recipe ID for the deployment project located at {deploymentProjectPath}", ex);
+            }
         }
 
         /// <summary>
--- a/src/AWS.Deploy.Common/Recipes/RecipeDefinition.cs
+++ b/src/AWS.Deploy.Common/Recipes/RecipeDefinition.cs
@@ -108,6 +108,11 @@ namespace AWS.Deploy.Common.Recipes
         /// </summary>
         public bool PersistedDeploymentProject { get; set; }
 
+        /// <summary>
+        /// The recipe ID of the parent recipe which was used to create the CDK deployment project. 
+        /// </summary>
+        public string? BaseRecipeId { get; set; }
+
         public RecipeDefinition(
             string id,
             string version,
--- a/src/AWS.Deploy.Orchestration/Utilities/DeployedApplicationQueryer.cs
+++ b/src/AWS.Deploy.Orchestration/Utilities/DeployedApplicationQueryer.cs
@@ -26,6 +26,11 @@ namespace AWS.Deploy.Orchestration.Utilities
         /// Get the list of compatible applications based on the matching elements of the deployed stack and recommendation, such as Recipe Id.
         /// </summary>
         Task<List<CloudApplication>> GetCompatibleApplications(List<Recommendation> recommendations, List<CloudApplication>? allDeployedApplications = null, OrchestratorSession? session = null);
+
+        /// <summary>
+        /// Checks if the given recommendation can be used for a redeployment to an existing cloudformation stack.
+        /// </summary>
+        bool IsCompatible(CloudApplication application, Recommendation recommendation);
     }
 
     public class DeployedApplicationQueryer : IDeployedApplicationQueryer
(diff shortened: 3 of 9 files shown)


test/AWS.Deploy.CLI.IntegrationTests/SaveCdkDeploymentProject/ReDeploymentTests/CompatibleReDeployment.cs
```diff
@@ -0,0 +1,182 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
```
ganeshnj, line 1: It is not ideal to grow the integration suite at this pace. There are other ways to test (especially unit test).

At max, there should be one integration test that validates that a saved deployment project deploys successfully.

Remember test pyramid

image

ganeshnj (changes requested):

test/AWS.Deploy.CLI.IntegrationTests/SaveCdkDeploymentProject/ReDeploymentTests/CompatibleReDeployment.cs
96malhar: I have streamlined the the compatible and incompatible re-deployment in a single integration test. This is the first (and the only) integration test that performs any sort of deployment using the persisted CDK project.

src/AWS.Deploy.Common/Recipes/RecipeDefinition.cs
```diff
@@ -108,6 +108,11 @@ public class RecipeDefinition
         /// </summary>
         public bool PersistedDeploymentProject { get; set; }
 
+        /// <summary>
+        /// The recipe ID of the base recommendation using which the persisted deployment project is created.
```
normj, line 112: Some grammar issues in this doc makes it hard to follow what this is. I assume it should be something like "The recipe ID of the recipe a saved deployment project was created from."

src/AWS.Deploy.Orchestration/Utilities/DeployedApplicationQueryer.cs
```diff
@@ -100,7 +100,14 @@ public async Task<List<CloudApplication>> GetCompatibleApplications(List<Recomme
             foreach (var app in allDeployedApplications)
             {
                 if (recommendations.Any(rec => string.Equals(rec.Recipe.Id, app.RecipeId, StringComparison.Ordinal)))
+                {
                     compatibleApplications.Add(app);
+                }
+                // check for persisted custom deployment recommendations
+                else if (recommendations.Any(rec => string.Equals(rec.Recipe.BaseRecipeId, app.RecipeId, StringComparison.Ordinal)))
```
normj, line 107: This 2 check process of for finding a recommendation for a given target recipe id is repeated here and on lines 271 and 275 of the DeployCommand. This seems like an easy thing in the future to forget doing 2 check process if we have another use case. Can we put this in a utility method?

normj (changes requested):

src/AWS.Deploy.Common/Recipes/RecipeDefinition.cs
96malhar: Fixed

src/AWS.Deploy.Orchestration/Utilities/DeployedApplicationQueryer.cs
96malhar: Fixed

src/AWS.Deploy.CLI/Commands/DeployCommand.cs
```diff
@@ -272,7 +272,12 @@ private async Task<Recommendation> GetSelectedRecommendationFromPreviousDeployme
 
             if (selectedRecommendation == null)
             {
-                var errorMessage = $"{deployedApplication.StackName} already exists as a Cloudformation stack but the recommendation used to deploy to the stack was not found.";
+                //check for persisted custom deployment recommendations
+                selectedRecommendation = recommendations.FirstOrDefault(x => string.Equals(x.Recipe.BaseRecipeId, deployedApplication.RecipeId, StringComparison.InvariantCultureIgnoreCase));
```
normj, line 276: If you change your `IsCompatible` function to be `GetRecommendation` and have it return back the recommendation instead of a bool you can use it here as well and replace likes 271-277 to be just

```
var selectedRecommenation = GetRecommendation(deployedApplication.RecipeId, recommendations);
```

In the `DeployedApplicationQueryer` that is currently calling `IsCompatible` you can change that to to call `GetRecommendation` and check to see if it returns back null.

normj (changes requested):

src/AWS.Deploy.CLI/Commands/DeployCommand.cs
96malhar: Fixed

src/AWS.Deploy.Orchestration/Utilities/DeployedApplicationQueryer.cs
```diff
@@ -137,5 +144,24 @@ public async Task<List<CloudApplication>> GetCompatibleApplications(List<Recomme
                 .OrderByDescending(x => x.LastUpdatedTime)
                 .ToList();
         }
+
+        /// <summary>
+        /// Gets the appropriate recommendation that will be used for a redeployment to the same Cloudformation stack. 
+        /// </summary>
+        public Recommendation? GetRecommendationForRedeployment(CloudApplication application, List<Recommendation> recommendations)
```
normj, line 151: I don't think this logic is going to work. Then GetRecommendationForRedeployment should return back a list or this method should take in a targetRecipeId. If a project currently has multiple saved deployment projects and the current application is using the parent recipe id. As I see this we are just taking whatever is the first match. Then up above we are comparing with the passed in deployment recipe id. And if that one wasn't the first one found here then we fail.

normj (changes requested):

src/AWS.Deploy.CLI/Commands/DeployCommand.cs
```diff
@@ -317,6 +316,22 @@ private async Task<Recommendation> GetSelectedRecommendationFromPreviousDeployme
             return selectedRecommendation;
         }
 
+        private Recommendation? GetRecommendationForRedeployment(List<Recommendation> recommendations, CloudApplication deployedApplication, string deploymentProjectPath)
+        {
+            if (!string.IsNullOrEmpty(deploymentProjectPath))
+            {
+                // There will only be ONE available recommendation if the deployment project path is provided. 
+                var recommendation = recommendations.First();
+                if (!string.Equals(recommendation.Recipe.BaseRecipeId, deployedApplication.RecipeId, StringComparison.Ordinal))
```
normj, line 325: You should also check if `recommendation.Recipe.BaseRecipeId` is also equal in case the user passes a deployment project for an application that is already using that project. Which gets back to my original comment of needing the double check. I suspose here you could just call your `IsCompatible` method.

normj (changes requested):

src/AWS.Deploy.CLI/Commands/DeployCommand.cs
96malhar: I have reverted a the `IsCompatible` function to just return a boolean. I have added a separate private method called `GetRecommendationForRedeployment` in the `DeployCommand` which performs this double check.

src/AWS.Deploy.Orchestration/Utilities/DeployedApplicationQueryer.cs
96malhar: Refer - https://github.com/aws/aws-dotnet-deploy/pull/318#discussion_r720445729

src/AWS.Deploy.CLI/Commands/DeployCommand.cs
```diff
@@ -317,6 +320,41 @@ private async Task<Recommendation> GetSelectedRecommendationFromPreviousDeployme
...
+
+        private async Task<string> GetDeploymentProjectRecipeId(string deploymentProjectPath)
+        {
+            if (!_directoryManager.Exists(deploymentProjectPath))
+            {
+                throw new InvalidOperationException($"Invalid deployment project path. {deploymentProjectPath} does not exist on the file system.");
+            }
+            try
+            {
+                var recipeFileName = _directoryManager.GetDirectoryInfo(deploymentProjectPath).Name + ".recipe";
```
normj, line 346: This is making the assumption the recipe file is the same name as the directory. We don't have that rule. Search the directory for the recipe file.

normj (changes requested):

src/AWS.Deploy.CLI/Commands/DeployCommand.cs
96malhar: Fixed

normj (approved):

test/AWS.Deploy.CLI.IntegrationTests/SaveCdkDeploymentProject/RedeploymentTests.cs
```diff
@@ -0,0 +1,175 @@
...
+
+        private async Task DeleteStack()
+        {
+            // Arrange input for delete
+            await _interactiveService.StdInWriter.WriteAsync("y"); // Confirm delete
+            await _interactiveService.StdInWriter.FlushAsync();
+            var deleteArgs = new[] { "delete-deployment", _stackName };
+
+            // Delete
+            await _app.Run(deleteArgs);
```
ganeshnj, line 142: assert on the returned value, it must be zero.

test/AWS.Deploy.CLI.IntegrationTests/SaveCdkDeploymentProject/RedeploymentTests.cs
```diff
@@ -0,0 +1,175 @@
...
+            var applicationUrl = deployStdOut.First(line => line.Trim().StartsWith("Endpoint:"))
+                .Split(" ")[1]
+                .Trim();
+
+            // URL could take few more minutes to come live, therefore, we want to wait and keep trying for a specified timeout
+            await _httpHelper.WaitUntilSuccessStatusCode(applicationUrl, TimeSpan.FromSeconds(5), TimeSpan.FromMinutes(5));
+
+            // list
+            var listArgs = new[] { "list-deployments" };
+            await _app.Run(listArgs);
```
ganeshnj, line 127: assert on the returned value, it must be zero.

test/AWS.Deploy.CLI.IntegrationTests/SaveCdkDeploymentProject/RedeploymentTests.cs
```diff
@@ -0,0 +1,175 @@
...
+        private async Task PerformInitialDeployment(string projectPath)
+        {
+            // Arrange input for deploy
+            await _interactiveService.StdInWriter.WriteLineAsync("1"); // Select ECS Fargate recommendation
+            await _interactiveService.StdInWriter.WriteAsync(Environment.NewLine); // Select default option settings
+            await _interactiveService.StdInWriter.FlushAsync();
+
+            // Deploy
+            var deployArgs = new[] { "deploy", "--project-path", projectPath, "--stack-name", _stackName, "--diagnostics" };
+            await _app.Run(deployArgs);
```
ganeshnj, line 108: assert on the returned value, it must be zero.

src/AWS.Deploy.CLI/Commands/DeployCommand.cs
```diff
@@ -317,6 +320,51 @@ private async Task<Recommendation> GetSelectedRecommendationFromPreviousDeployme
...
+            try
+            {
+                var recipeFiles = _directoryManager.GetFiles(deploymentProjectPath, "*.recipe");
+                if (recipeFiles.Length == 0)
+                {
+                    throw new InvalidOperationException($"Failed to find a recipe file at {deploymentProjectPath}");
+                }
+                if (recipeFiles.Length > 1)
+                {
+                    throw new InvalidOperationException($"Found more than one recipe file at {deploymentProjectPath}");
```
ganeshnj, line 354: It tells **what** but doesn't tell why there must not exist more than one recipe files. 

Something like this would make more sense
```
Found more than one recipe files at {path}. Only one recipe file per deployment project is supported.
```

src/AWS.Deploy.Common/Recipes/RecipeDefinition.cs
```diff
@@ -108,6 +108,11 @@ public class RecipeDefinition
         /// </summary>
         public bool PersistedDeploymentProject { get; set; }
 
+        /// <summary>
+        /// The recipe ID of the parent recipe which was used to create the CDK deployment project. 
+        /// </summary>
+        public string? BaseRecipeId { get; set; }
```
ganeshnj, line 114: Update recipe schema to reflect the new property.

test/AWS.Deploy.CLI.IntegrationTests/SaveCdkDeploymentProject/RedeploymentTests.cs
```diff
@@ -0,0 +1,175 @@
...
+
+            _app = serviceProvider.GetService<App>();
+            Assert.NotNull(_app);
+
+            _interactiveService = serviceProvider.GetService<InMemoryInteractiveService>();
+            Assert.NotNull(_interactiveService);
+
+            _testAppManager = new TestAppManager();
+
+            _commandLineWrapper = new CommandLineWrapper(new ConsoleOrchestratorLogger(new ConsoleInteractiveServiceImpl()));
```
ganeshnj, line 58: In integration tests, always use `InMemoryInteractiveService` for logging. This allows buffering of the output and writing on the stdout on the completion of the test. A integ test **must** not write directly to the console.

test/AWS.Deploy.CLI.IntegrationTests/SaveCdkDeploymentProject/RedeploymentTests.cs
```diff
@@ -0,0 +1,175 @@
...
+        {
+            if (_isDisposed) return;
+
+            if (disposing)
+            {
+                var isStackDeleted = _cloudFormationHelper.IsStackDeleted(_stackName).GetAwaiter().GetResult();
+                if (!isStackDeleted)
+                {
+                    _cloudFormationHelper.DeleteStack(_stackName).GetAwaiter().GetResult();
+                }
```
ganeshnj, line 164: write all the buffered stdout/err here from `InMemoryInteractiveService`

Ex. https://github.com/aws/aws-dotnet-deploy/blob/main/test/AWS.Deploy.CLI.IntegrationTests/BlazorWasmTests.cs#L145

test/AWS.Deploy.CLI.IntegrationTests/SaveCdkDeploymentProject/RedeploymentTests.cs
```diff
@@ -0,0 +1,175 @@
...
+
+            _commandLineWrapper = new CommandLineWrapper(new ConsoleOrchestratorLogger(new ConsoleInteractiveServiceImpl()));
+        }
+
+        [Fact]
+        public async Task AttemptWorkFlow()
+        {
+            _stackName = $"WebAppWithDockerFile{Guid.NewGuid().ToString().Split('-').Last()}";
+            var tempDirectoryPath = new TestAppManager().GetProjectPath(string.Empty);
+            await _commandLineWrapper.Run("git init", tempDirectoryPath);
```
ganeshnj, line 66: Is this required? Can this test be without git repo? I don't see it is a absolute necessity to make the test work.

test/AWS.Deploy.CLI.IntegrationTests/SaveCdkDeploymentProject/RedeploymentTests.cs
```diff
@@ -0,0 +1,175 @@
...
+
+            // Deploy
+            var deployArgs = new[] { "deploy", "--project-path", projectPath, "--stack-name", _stackName, "--diagnostics" };
+            await _app.Run(deployArgs);
+
+            // Verify application is deployed and running
+            Assert.Equal(StackStatus.CREATE_COMPLETE, await _cloudFormationHelper.GetStackStatus(_stackName));
+
+            var cluster = await _ecsHelper.GetCluster(_stackName);
+            Assert.Equal("ACTIVE", cluster.Status);
```
ganeshnj, line 114: Can use ECS constants for ACTIVE?  Also L93.

https://docs.aws.amazon.com/sdkfornet/v3/apidocs/items/ECS/TTaskDefinitionStatus.html

ganeshnj (changes requested):

test/AWS.Deploy.CLI.IntegrationTests/SaveCdkDeploymentProject/RedeploymentTests.cs
96malhar: Git repo is not required. Removed it.

test/AWS.Deploy.CLI.IntegrationTests/SaveCdkDeploymentProject/RedeploymentTests.cs
96malhar: The `_commandLineWrapper` was being used to initialize a Git repository. Removed it as this test does not require a Git repo.

test/AWS.Deploy.CLI.IntegrationTests/SaveCdkDeploymentProject/RedeploymentTests.cs
96malhar: Added.

test/AWS.Deploy.CLI.IntegrationTests/SaveCdkDeploymentProject/RedeploymentTests.cs
96malhar: Fixed

ganeshnj (approved):

PR #322, opened 2021-09-05
feat: Update ECS Fargate Console recipe to new style

normj:
*Description of changes:*
Update ECS Fargate service recipe to the new style of writing recipes.

Also added the ability to configure AutoScaling for the service using either CPU or Memory as scaling alarms.


src/AWS.Deploy.Recipes/CdkTemplates/ConsoleAppECSFargateService/Generated/Configurations/AutoScalingConfiguration.cs
```diff
@@ -0,0 +1,46 @@
...
+
+// This is a generated file from the original deployment recipe. It contains properties for
+// all of the settings defined in the recipe file. It is recommended to not modify this file in order
+// to allow easy updates to the file when the original recipe that this project was created from has updates.
+// This class is marked as a partial class. If you add new settings to the recipe file, those settings should be
+// added to partial versions of this class outside of the Generated folder for example in the Configuration folder.
+
+namespace ConsoleAppEcsFargateService.Configurations
+{
+    public partial class AutoScalingConfiguration
```
ganeshnj, line 16: I observed a weird behavior working with @96malhar, even though auto scaling was not configured, still metadata in the CF template had auto scaling configs due to default values.

Can Cpu* and Memory* be nullable? or something like in L36?

src/AWS.Deploy.Recipes/CdkTemplates/ConsoleAppECSFargateService/Generated/Configurations/AutoScalingConfiguration.cs
```diff
@@ -0,0 +1,46 @@
...
+
+        public ScalingTypeEnum? ScalingType { get; set; }
+
+
+
+        public double CpuTypeTargetUtilizationPercent { get; set; } = 70;
+
+        public int CpuTypeScaleInCooldownSeconds { get; set; } = defaultCooldown;
+
+        public int CpuTypeScaleOutCooldownSeconds { get; set; } = defaultCooldown;
```
ganeshnj, line 36:
```suggestion
        public CpuTypeScaling? CpuTypeScaling { get; set; }

```

```
    public class CpuTypeScaling
    {
        public double TargetUtilizationPercent { get; set; } = 70;
        public int ScaleInCooldownSeconds { get; set; } = 300;
        public int ScaleOutCooldownSeconds { get; set; } = 300;
    }
```

src/AWS.Deploy.Recipes/CdkTemplates/ConsoleAppECSFargateService/Generated/Recipe.cs
```diff
@@ -0,0 +1,235 @@
...
+        public FargateService? AppFargateService { get; set; }
+
+        public ScalableTaskCount? AutoScalingConfiguration { get; private set; }
+
+        public string AutoScaleTypeCpuType { get; } = "AutoScaleTypeCpuType";
+        public string AutoScaleTypeMemoryType { get; } = "AutoScaleTypeMemoryType";
+
+
+        public Recipe(Construct scope, IRecipeProps<Configuration> props)
+            : base(scope, "Recipe")
```
ganeshnj, line 50: There should be a comment explaining why the construct id is constant here. And changing it will impact the stack.

ganeshnj (changes requested):

src/AWS.Deploy.Recipes/CdkTemplates/ConsoleAppECSFargateService/Generated/Recipe.cs
normj: Done

src/AWS.Deploy.Recipes/CdkTemplates/ConsoleAppECSFargateService/Generated/Configurations/AutoScalingConfiguration.cs
normj: Since the nesting of the configuration parameters map 1 to 1 to the option settings I don't want to create another level of option settings in the recipe file. So I want to keep the settings flat here.

src/AWS.Deploy.Recipes/CdkTemplates/ConsoleAppECSFargateService/Generated/Configurations/AutoScalingConfiguration.cs
normj: Since we settings have default values in the recipe file the values will always have a value here as well and be serialized. We would need to develop a system that allows us to control serialization based on the `DependsOn` settings to make it so the non-used values don't get put in the CF template.

@96malhar issue turned out not to have anything to do with the the non-used values being put in the template and it was a bug in how we were serializing enums.

src/AWS.Deploy.Recipes/CdkTemplates/AspNetAppEcsFargate/AppStack.cs
```diff
@@ -43,17 +43,17 @@ private void CustomizeCDKProps(CustomizePropsEventArgs<Recipe> evnt)
...
+            {
+                if (evnt.Props is ContainerDefinitionOptions props)
+                {
+                    Console.WriteLine("Customizing AppContainerDefinition");
+                    if (props.Environment == null)
+                        props.Environment = new Dictionary<string, string>();
+
+                    props.Environment["EXAMPLE_ENV1"] = "EXAMPLE_VALUE1";
+                }
+            }
```
philasmar, line 56: this should be commented right?

src/AWS.Deploy.Recipes/CdkTemplates/ConsoleAppECSFargateService/ConsoleAppECSFargateService.csproj
```diff
@@ -44,7 +43,7 @@
   </ItemGroup>
   <!-- The project has run through the project template engine and "AWSDeployRecipesCDKCommonVersion" has been replaced with valid version number so a PackageReference is used. -->
   <ItemGroup Condition=" '$(RecipeCDKCommonVersion.ToUpper())' != 'AWSDEPLOYRECIPESCDKCOMMONVERSION' ">
-    <PackageReference Include="AWS.Deploy.Recipes.CDK.Common" Version="AWSDeployRecipesCDKCommonVersion"/>
+    <PackageReference Include="AWS.Deploy.Recipes.CDK.Common" Version="AWSDeployRecipesCDKCommonVersion" />
```
philasmar, line 46: nit: undo whitespace change

src/AWS.Deploy.Recipes/CdkTemplates/ConsoleAppECSFargateService/Generated/Configurations/AutoScalingConfiguration.cs
```diff
@@ -0,0 +1,46 @@
...
+        public int MinCapacity { get; set; } = 3;
+
+        public int MaxCapacity { get; set; } = 6;
+
+        public enum ScalingTypeEnum { Cpu, Memory }
+
+        public ScalingTypeEnum? ScalingType { get; set; }
+
+
+
```
philasmar, line 31: nit: can you remove the extra new lines?

src/AWS.Deploy.Recipes/CdkTemplates/ConsoleAppECSFargateService/Generated/Configurations/AutoScalingConfiguration.cs
```diff
@@ -0,0 +1,46 @@
...
+        public double CpuTypeTargetUtilizationPercent { get; set; } = 70;
+
+        public int CpuTypeScaleInCooldownSeconds { get; set; } = defaultCooldown;
+
+        public int CpuTypeScaleOutCooldownSeconds { get; set; } = defaultCooldown;
+
+
+
```
philasmar, line 39: nit: can you remove the extra new lines?

src/AWS.Deploy.Recipes/RecipeDefinitions/ConsoleAppECSFargateService.recipe
```diff
@@ -124,7 +124,7 @@
                     "Validators": [
                         {
                             "ValidatorType": "Regex",
-                            "Configuration" : {
+                            "Configuration": {
```
philasmar, line 127: nit: there's a lot of whitespace changes. Can you remove them?

philasmar (changes requested):

src/AWS.Deploy.Recipes/CdkTemplates/ConsoleAppECSFargateService/Generated/Configurations/AutoScalingConfiguration.cs
ganeshnj: There are multiple ways to solve this, 
- Make serialization intelligent (as you suggested)
- Don't assign default values to the actual configuration rather use some other constant if the config value is not set


Something like
```
public string? Value { get; set; }
public const string ValueDefault = "MyDefaultValue";
```

and use conditional logic to get the final value. This should be easy to do and avoid unnecessary data in CF template.

Looking at the [CDK API](https://docs.aws.amazon.com/cdk/api/latest/docs/@aws-cdk_aws-ecs.RequestCountScalingProps.html#scaleincooldown), they have default values already same as CDK template i.e. 300. Can [ScaleInCooldown ](https://docs.aws.amazon.com/cdk/api/latest/docs/@aws-cdk_aws-ecs.RequestCountScalingProps.html#scaleincooldown)and [ScaleOutCooldown ](https://docs.aws.amazon.com/cdk/api/latest/docs/@aws-cdk_aws-ecs.RequestCountScalingProps.html#scaleincooldown)defaults be removed? Is there a reason that deploy tool has to keep a different default than CDK?

src/AWS.Deploy.Recipes/CdkTemplates/ConsoleAppECSFargateService/Generated/Configurations/AutoScalingConfiguration.cs
ganeshnj: Fair enough.

src/AWS.Deploy.Recipes/CdkTemplates/ConsoleAppECSFargateService/Generated/Configurations/AutoScalingConfiguration.cs
normj: The default values are coming from the recipe json file. Whether I put defaults here doesn't matter the process of creating the appsettings.json file from the orchestrator to the CDK project will put the default values in it. I just put the default values in the CDK code as well to avoid checking that will be null. We no they won't be null because the recipe always passes in a value.

src/AWS.Deploy.Recipes/CdkTemplates/AspNetAppEcsFargate/AppStack.cs
normj: Yes and fixed

src/AWS.Deploy.Recipes/CdkTemplates/ConsoleAppECSFargateService/ConsoleAppECSFargateService.csproj
normj: Actually I think the space is the default formatting VS does. So I think we should make this change so VS won't keep changing the value.

src/AWS.Deploy.Recipes/CdkTemplates/ConsoleAppECSFargateService/Generated/Configurations/AutoScalingConfiguration.cs
normj: I actually prefer those lines because it is grouping the CPU and Memory autoscaling settings.

src/AWS.Deploy.Recipes/CdkTemplates/ConsoleAppECSFargateService/Generated/Configurations/AutoScalingConfiguration.cs
normj: I actually prefer those lines because it is grouping the CPU and Memory autoscaling settings.

src/AWS.Deploy.Recipes/RecipeDefinitions/ConsoleAppECSFargateService.recipe
normj: I think the VS did the autoformatting and it is making it the same as the other lines. So I think we should make the change to avoid getting this in the future.

src/AWS.Deploy.Recipes/CdkTemplates/ConsoleAppECSFargateService/Generated/Configurations/AutoScalingConfiguration.cs
ganeshnj: If default values are coming from recipe files, adding default values in C# code here can hide issues. You can use force unwrapping using `!` for optional fields/properties if it is 100% guaranteed that they will not be null.

Caution: on a null value, the app will crash on a force unwrap.

philasmar (approved):

ganeshnj (approved):

PR #328, opened 2021-09-13
Several bug fixes

normj:
*Description of changes:*
Fixing a few bugs discovered while doing some end to end testing:

* Using a recipe settings file with settings that were of type `double` failed to serialize.
* NPE when the tool was cleaning up orphan stacks in user settings
* Make the `Configuration` object accessible in the recipes to make customization easier
* Remove dead code now that all recipes are using the new style


src/AWS.Deploy.Recipes/CdkTemplates/AspNetAppAppRunner/AppStack.cs
```diff
@@ -19,9 +19,13 @@ namespace AspNetAppAppRunner
 {
     public class AppStack : Stack
     {
+        private readonly Configuration _configuration;
```
ganeshnj, line 22: There is no usage of _configuration, is this only to let people there is configuration object?

normj: Yes, if a user wants access to the configuration in the customization they have to know to how to capture during the constructor. I want to make that easy for users to find the settings during the customization callback. 

It was an awkward part of my recent demo.

ganeshnj (approved):

philasmar (approved):

PR #329, opened 2021-09-13
fix: Server-mode GetRecipe API supports finding recipes for custom deployment project recipe ids.

normj:
*Description of changes:*
Fix issue with the GetRecipe REST API from server mode that was not using custom recipe location paths to search for recipes.

### DIFF
--- a/src/AWS.Deploy.Orchestration/RecipeHandler.cs
+++ b/src/AWS.Deploy.Orchestration/RecipeHandler.cs
@@ -4,6 +4,9 @@
 using System;
 using System.Collections.Generic;
 using System.IO;
+using System.Linq;
+using System.Threading.Tasks;
+using AWS.Deploy.Common;
 using AWS.Deploy.Common.Recipes;
 using AWS.Deploy.Recipes;
 using Newtonsoft.Json;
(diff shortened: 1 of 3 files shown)


test/AWS.Deploy.CLI.UnitTests/ServerModeTests.cs
```diff
@@ -10,6 +10,18 @@
...
 
+using Moq;
+using AWS.Deploy.Common.IO;
+using AWS.Deploy.Common.DeploymentManifest;
+using AWS.Deploy.Orchestration.LocalUserSettings;
+using AWS.Deploy.CLI.Utilities;
+using AWS.Deploy.Common;
+using AWS.Deploy.CLI.UnitTests.Utilities;
+
+using Moq;
```
ganeshnj, line 21: GitHub bot at action.

src/AWS.Deploy.Orchestration/RecipeHandler.cs
```diff
@@ -12,24 +15,37 @@ namespace AWS.Deploy.Orchestration
...
-                        throw new Exception($"Failed to Deserialize Recipe Definition [{recipeDefinitionFile}]: {e.Message}", e);
+                        try
+                        {
+                            var content = File.ReadAllText(recipeDefinitionFile);
+                            var definition = JsonConvert.DeserializeObject<RecipeDefinition>(content);
+                            recipeDefinitions.Add(definition);
+                        }
+                        catch (Exception e)
+                        {
+                            throw new Exception($"Failed to Deserialize Recipe Definition [{recipeDefinitionFile}]: {e.Message}", e);
```
ganeshnj, line 47: This exception message is too generic, if file IO fails, it will show the same message which isn't actually deserialization.

test/AWS.Deploy.CLI.UnitTests/ServerModeTests.cs
normj: Fixed

src/AWS.Deploy.Orchestration/RecipeHandler.cs
normj: The message also contains the underlying exception message. For an exception that shouldn't happen that should be good enough. I don't think adding a `catch(IOException e)` and then still just rethrowing with a our own message will improve things.

ganeshnj (approved):

philasmar (approved):

PR #332, opened 2021-09-17
fix: Replace default values for option setting items with stack name as prefix

96malhar:
*Description of changes:*
Currently in the deploy tool we are prefixing the default values for certain parameter (ECS Service name, Elastic Beanstalk Environment etc..) with the `ProjectName`. However, the `ProjectName` is not unique because we use the same project for multiple integration tests. This is causing an issue while deletion of stacks because the Cloudformation resources are shared between multiple stacks. 

In order to avoid this issue, we are replacing the `ProjectName` prefix is replaced with the `StackName`, which is unique per test. As a result, all Cloudformation resources created by a test are exclusive to the stack that it created

The same is also extended to the server mode.

### DIFF
--- a/src/AWS.Deploy.CLI/Commands/DeployCommand.cs
+++ b/src/AWS.Deploy.CLI/Commands/DeployCommand.cs
@@ -25,6 +25,8 @@ namespace AWS.Deploy.CLI.Commands
 {
     public class DeployCommand
     {
+        public const string REPLACE_TOKEN_STACK_NAME = "{StackName}";
+
         private readonly IToolInteractiveService _toolInteractiveService;
         private readonly IOrchestratorInteractiveService _orchestratorInteractiveService;
         private readonly ICdkProjectHandler _cdkProjectHandler;
@@ -162,7 +164,7 @@ namespace AWS.Deploy.CLI.Commands
 
             // Get Cloudformation stack name.
             var cloudApplicationName = GetCloudApplicationName(stackName, userDeploymentSettings, compatibleApplications);
-
+            
             // Find existing application with the same CloudFormation stack name.
             var deployedApplication = allDeployedApplications.FirstOrDefault(x => string.Equals(x.Name, cloudApplicationName));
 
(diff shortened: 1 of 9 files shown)


philasmar (approved):

src/AWS.Deploy.CLI/ServerMode/Controllers/DeploymentController.cs
```diff
@@ -36,6 +36,8 @@ namespace AWS.Deploy.CLI.ServerMode.Controllers
     [Route("api/v1/[controller]")]
     public class DeploymentController : ControllerBase
     {
+        private const string REPLACE_TOKEN_STACK_NAME = "{StackName}";
```
normj, line 39: The constant should not be defined more then once.

src/AWS.Deploy.CLI/Commands/DeployCommand.cs
```diff
@@ -25,6 +25,8 @@ namespace AWS.Deploy.CLI.Commands
 {
     public class DeployCommand
     {
+        private const string REPLACE_TOKEN_STACK_NAME = "{StackName}";
```
normj, line 28: Why was this changed from `ProjectName` to `StackName`? I assume it is fine but these are the types of changes that would be breaking changes after we have users saving their deployment projects.

normj (changes requested):
Lets us be better about PR descriptions. Explain what the root cause of the issue was and how this change fixes the issue.

src/AWS.Deploy.CLI/Commands/DeployCommand.cs
96malhar: Updated the PR description to provide clarity.

src/AWS.Deploy.CLI/ServerMode/Controllers/DeploymentController.cs
96malhar: I have only defined this once in this class.

normj: It is also defined in `DeployCommand` which is the same project.

normj (changes requested):

src/AWS.Deploy.CLI/ServerMode/Controllers/DeploymentController.cs
96malhar: Fixed

PR #335, opened 2021-09-23
Integration test improvements

ganeshnj:
*Issue #, if available:*

*Description of changes:*

1. **Serialization of parallel integration tests ouput**
Integration tests run in parallel which emits logs in a non-deterministic order, this creates log reading painful and hard to debug. This change combines all types of logs to one stream and puts on the Console at the end of test class disposal.
2. All test commands now use diagnostics flag and assert on the return value [Run method](https://github.com/aws/aws-dotnet-deploy/blob/main/src/AWS.Deploy.CLI/App.cs#L25)
3. Fix various test failures
4. Log exceptions that are caught be never rethrown

### DIFF
--- a/src/AWS.Deploy.CLI/AWSUtilities.cs
+++ b/src/AWS.Deploy.CLI/AWSUtilities.cs
@@ -1,6 +1,7 @@
 // Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
 // SPDX-License-Identifier: Apache-2.0
 
+using System;
 using System.Collections.Generic;
 using System.Linq;
 using System.Threading.Tasks;
@@ -8,7 +9,9 @@ using Amazon.Runtime;
 using Amazon.Runtime.CredentialManagement;
 using Amazon.EC2.Model;
 using System.IO;
+using AWS.Deploy.CLI.Commands.CommandHandlerInput;
 using AWS.Deploy.CLI.Utilities;
+using AWS.Deploy.Common;
 using AWS.Deploy.Common.IO;
 
 namespace AWS.Deploy.CLI
@@ -25,7 +28,10 @@ namespace AWS.Deploy.CLI
         private readonly IConsoleUtilities _consoleUtilities;
         private readonly IDirectoryManager _directoryManager;
 
-        public AWSUtilities(IToolInteractiveService toolInteractiveService, IConsoleUtilities consoleUtilities, IDirectoryManager directoryManager)
+        public AWSUtilities(
+            IToolInteractiveService toolInteractiveService,
+            IConsoleUtilities consoleUtilities,
+            IDirectoryManager directoryManager)
         {
             _toolInteractiveService = toolInteractiveService;
             _consoleUtilities = consoleUtilities;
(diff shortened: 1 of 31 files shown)


src/AWS.Deploy.CLI/AWSUtilities.cs
```diff
@@ -24,63 +27,73 @@ public class AWSUtilities : IAWSUtilities
...
+                    {
+                        _toolInteractiveService.WriteLine("Configuring AWS Credentials using AWS SDK credential search.");
                         return fallbackCredentials;
+                    }
+                }
+                catch (AmazonServiceException ex)
+                {
+                    // FallbackCredentialsFactory throws an exception if no credentials are found. Burying exception because if no credentials are found
+                    // we want to continue and ask the user to select a profile.
+                    if (_commandInputService.Diagnostics)
```
normj, line 80: Add a `WriteDiagnostics` method to IToolInteractiveServices and internal to that method check if diagnostics is enabled.

src/AWS.Deploy.CLI/Commands/CommandHandlerInput/CommandInputService.cs
```diff
@@ -0,0 +1,25 @@
...
+    {
+        bool Diagnostics { get; }
+        DeleteCommandHandlerInput? DeleteInput { get; set; }
+        DeployCommandHandlerInput? DeployInput { get; set; }
+        GenerateDeploymentProjectCommandHandlerInput? GenerateDeploymentProjectInput { get; set; }
+        ListCommandHandlerInput? List { get; set; }
+        ServerModeCommandHandlerInput? ServerModeInput { get; set; }
+    }
+
+    public class CommandInputService : ICommandInputService
```
normj, line 16: This is weird. A base class that has a property to each of the sub classes. Is the goal to have input that has some global settings like `Diagnostics` and the command specific settings?

If so I would recommend using generics instead of a union of all of the children. Something like

```
    public interface ICommandInputService
    {
        bool Diagnostics { get; }
    }

    public interface ICommandInputService<T> : ICommandInputService
    {
        T Input { get; }
    }

    public CommandInputService<T> : ICommandInputService<T>
    {
          public bool Diagnostics => ...
          T Input {get;set;}
    }
```

src/AWS.Deploy.Orchestration/Utilities/DeployedApplicationQueryer.cs
```diff
@@ -124,16 +124,24 @@ public async Task<List<CloudApplication>> GetCompatibleApplications(List<Recomme
                 catch (FailedToUpdateLocalUserSettingsFileException ex)
                 {
                     _orchestratorInteractiveService.LogErrorMessageLine(ex.Message);
+                    if (session.Diagnostics)
```
normj, line 127: Again we should push the diagnostics check into the _orchestratorInteractiveService.

src/AWS.Deploy.CLI/Commands/CommandHandlerInput/CommandInputService.cs
normj: I'm not convinced we need `ICommandInputService` especially if we the Diagnostics flag into the interactive services. All I see it doing is passing around Diagnostics and the input type. But the input shape isn't even used once the command is executed. The input types are currently just used to translate the CLI switches into the arguments of the Execute commands then throw away. This change keeps them around for no new reason.

normj (changes requested):

src/AWS.Deploy.Orchestration/Utilities/DeployedApplicationQueryer.cs
ganeshnj: Fixed

src/AWS.Deploy.CLI/Commands/CommandHandlerInput/CommandInputService.cs
ganeshnj: Removed `ICommandInputService` for now, at this point not needed.

src/AWS.Deploy.CLI/AWSUtilities.cs
ganeshnj: Fixed

normj (approved):

philasmar (approved):

PR #339, opened 2021-09-28
test: make sure only one thread at a time allowed to perform cdk inst…

ganeshnj:
*Issue #, if available:*
A thread is installing the CDK, while another is trying to use CDK CLI while has not finished installation. In this race condition, multiple threads can lead to using broken CDK CLI.

*Description of changes:*
Use SemaphoreSlim to lock async code and made sure that only one thread allowed to perform check at a time.

Removed installing cdk cli from CI CodeBuild job. This makes sure that the release pipeline and CI job has the same operating conditions.

### DIFF
--- a/src/AWS.Deploy.Orchestration/Orchestrator.cs
+++ b/src/AWS.Deploy.Orchestration/Orchestrator.cs
@@ -5,6 +5,7 @@ using System;
 using System.Collections.Generic;
 using System.IO;
 using System.Linq;
+using System.Threading;
 using System.Threading.Tasks;
 using AWS.Deploy.Common;
 using AWS.Deploy.Common.IO;
@@ -25,6 +26,7 @@ namespace AWS.Deploy.Orchestration
     /// </summary>
     public class Orchestrator
     {
+        private static readonly SemaphoreSlim s_cdkManagerSemaphoreSlim = new(1,1);
         private const string REPLACE_TOKEN_LATEST_DOTNET_BEANSTALK_PLATFORM_ARN = "{LatestDotnetBeanstalkPlatformArn}";
 
         private readonly ICdkProjectHandler? _cdkProjectHandler;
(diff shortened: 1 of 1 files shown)


96malhar (approved):

src/AWS.Deploy.Orchestration/Orchestrator.cs
```diff
@@ -163,7 +165,19 @@ public async Task DeployRecommendation(CloudApplication cloudApplication, Recomm
...
-                    await _cdkManager.EnsureCompatibleCDKExists(Constants.CDK.DeployToolWorkspaceDirectoryRoot, cdkVersion);
+
+                    try
+                    {
+                        await s_cdkManagerSemaphoreSlim.WaitAsync();
+                        await _cdkManager.EnsureCompatibleCDKExists(Constants.CDK.DeployToolWorkspaceDirectoryRoot, cdkVersion);
+                    }
+                    finally
+                    {
+                        if (s_cdkManagerSemaphoreSlim.CurrentCount == 0)
```
normj, line 176: Why would we not always call `Release`? Seems like the `if` statement is just adding complexity but if we enter the block we should always release on the way out.

normj (changes requested):

src/AWS.Deploy.Orchestration/Orchestrator.cs
ganeshnj: Because release can throw an exception (SemaphoreFullException)

https://docs.microsoft.com/en-us/dotnet/api/system.threading.semaphore.release?view=net-5.0

ganeshnj: This can happen if my WaitAsync call fails and in finally block, it tries to release the semaphore.

normj: The exception would be because you have a design issue where you are calling release more often then entering the lock. If you did have that design issue this `if` statement would be hiding that issue.

normj (approved):

PR #342, opened 2021-09-30
fix: update StartDeploymentSession to return default stack name

philasmar:
*Issue #, if available:*
IDE-5945

*Description of changes:*
fix: update StartDeploymentSession to return default stack name

### DIFF
--- a/src/AWS.Deploy.CLI/ServerMode/Controllers/DeploymentController.cs
+++ b/src/AWS.Deploy.CLI/ServerMode/Controllers/DeploymentController.cs
@@ -81,7 +81,21 @@ namespace AWS.Deploy.CLI.ServerMode.Controllers
 
             _stateServer.Save(output.SessionId, state);
 
-            output.DefaultDeploymentName = _cloudApplicationNameGenerator.GenerateValidName(state.ProjectDefinition, new List<CloudApplication>());
+            var deployedApplicationQueryer = serviceProvider.GetRequiredService<IDeployedApplicationQueryer>();
+            var session = CreateOrchestratorSession(state);
+            var orchestrator = CreateOrchestrator(state);
+
+            // Determine what recommendations are possible for the project.
+            var recommendations = await orchestrator.GenerateDeploymentRecommendations();
+            state.NewRecommendations = recommendations;
+
+            // Get all existing applications that were previously deployed using our deploy tool.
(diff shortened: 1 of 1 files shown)


src/AWS.Deploy.CLI/ServerMode/Controllers/DeploymentController.cs
```diff
@@ -81,7 +81,19 @@ await _projectParserUtility.Parse(input.ProjectPath)
 
             _stateServer.Save(output.SessionId, state);
 
-            output.DefaultDeploymentName = _cloudApplicationNameGenerator.GenerateValidName(state.ProjectDefinition, new List<CloudApplication>());
+            var deployedApplicationQueryer = serviceProvider.GetRequiredService<IDeployedApplicationQueryer>();
+            var session = CreateOrchestratorSession(state);
+            var orchestrator = CreateOrchestrator(state);
+
+            // Determine what recommendations are possible for the project.
+            var recommendations = await orchestrator.GenerateDeploymentRecommendations();
```
normj, line 89: Should we save `recommendations` to `state.NewRecommendations` and `existingApplications` to `state.ExistingDeployments` so we don't have to recomputed those? That would mean those controller getter methods should look to see if the state already has a value before doing all the work again.

normj (changes requested):
There should be a test to go along with this change.

src/AWS.Deploy.CLI/ServerMode/Controllers/DeploymentController.cs
philasmar: Fixed

normj (approved):

src/AWS.Deploy.CLI/ServerMode/Controllers/DeploymentController.cs
```diff
@@ -116,7 +130,7 @@ public async Task<IActionResult> GetRecommendations(string sessionId)
 
             var output = new GetRecommendationsOutput();
 
-            state.NewRecommendations = await orchestrator.GenerateDeploymentRecommendations();
+            state.NewRecommendations ??= await orchestrator.GenerateDeploymentRecommendations();
```
ganeshnj, line 133: above `NewRecommendations` is replaced with newly available recommendations but here only if it is null. A comment/description would help why it is done like this.

ganeshnj (changes requested):

PR #344, opened 2021-10-01
fix: update recipe descriptions to be more informative

philasmar:
*Issue #, if available:*
IDE-5944

*Description of changes:*
fix: update recipe descriptions to be more informative


ganeshnj (approved):

96malhar (approved):

normj (changes requested):
Do not put things in ALL CAPS. It looks like we are yelling at our customers and is inconsistent with Visual Studio.

Also do not say "USING DOCKER" in the recommendation because I'm pretty sure Fargate used containerd not docker to run the containers.

philasmar:
> Do not put things in ALL CAPS. It looks like we are yelling at our customers and is inconsistent with Visual Studio.
> 
> Also do not say "USING DOCKER" in the recommendation because I'm pretty sure Fargate used containerd not docker to run the containers.

Fixed

PR #345, opened 2021-10-06
Fix sourceName for template config to the correct value

normj:
*Description of changes:*
Noticed when saving a deployment-project for Blazor WebAssembly the project name and namespace names were not being correctly changed with the project name. The was a copy/paste with the template.config using the name of another recipe.


src/AWS.Deploy.Recipes/CdkTemplates/BlazorWasm/AppStack.cs
```diff
@@ -48,7 +48,6 @@ private void CustomizeCDKProps(CustomizePropsEventArgs<Recipe> evnt)
             {
                 if (evnt.Props is DistributionProps props)
                 {
```
philasmar, line 50: nit: Can you leave the write line and comment the entire block like we are doing in other templates?

philasmar (approved):

src/AWS.Deploy.Recipes/CdkTemplates/BlazorWasm/AppStack.cs
normj: Done

ganeshnj (approved):

Issue #165, opened 2021-03-19
docker build image for react and angular SPA doesn't have node installed


ganeshnj:
**Describe the bug**
docker build image for react and angular SPA doesn't have node installed which is required for building the project.

**To Reproduce**
1. Create a sample Angular/React application
2. `dotnet aws deploy`
3. Observe the npm command failure

**Logs**
 > [publish 1/1] RUN dotnet publish "Demo.csproj" -c Release -o /app/publish:
#16 0.607 Microsoft (R) Build Engine version 16.7.2+b60ddb6f4 for .NET
#16 0.607 Copyright (C) Microsoft Corporation. All rights reserved.
#16 0.607
#16 1.023   Determining projects to restore...
#16 1.283   All projects are up-to-date for restore.
#16 1.718   Demo-> /src/Demo/bin/Release/netcoreapp3.1/Demo.dll
#16 1.720   Demo-> /src/Demo/bin/Release/netcoreapp3.1/Demo.Views.dll
#16 1.763   /bin/sh: 2: /tmp/tmp43d60b22442c4399bb30e4d6e4a0f8d4.exec.cmd: npm: not found
#16 1.764 /src/Demo/Demo.csproj(35,5): error MSB3073: The command "npm install" exited with code 127.
------
executor failed running [/bin/sh -c dotnet publish "Demo.csproj" -c Release -o /app/publish]: exit code: 1
We were unable to build the docker image due to the following error:
Exception of type 'AWS.Deploy.Orchestrator.DockerBuildFailedException' was thrown.
Docker builds usually fail due to executing them from a working directory that is incompatible with the Dockerfile.
You can try setting the 'Docker Execution Directory' in the option settings.

**Expected behavior**
`node` must be installed prior to the project build.

**Solution**
```
RUN apt-get update -yq \
    && apt-get install curl gnupg -yq \
    && curl -sL https://deb.nodesource.com/setup_10.x | bash \
    && apt-get install nodejs -yq
```

**Screenshots**
![image](https://user-images.githubusercontent.com/8882380/111828299-56f96400-88a8-11eb-8314-0d6f4b1c5555.png)

**Screen recording**
If applicable, add screen recording to help explain your problem.

**Environment (please complete the following information):**
 - CLI Version: [eg. 1.0.0] 
 - Include the output of `dotnet --info`
 - OS Info: 

**Additional context**

---

This is a :bug: bug report

PR #348, opened 2021-10-06
fix: install nodejs in publish image to support SPA app publish

ganeshnj:
*Issue #, if available:*
#165 

*Description of changes:*
final image doesn't have any nodejs dependencies, this change is only for publish image


normj (approved):
Curious if this will bother people to always have a slower docker build due to the node.js install but willing to approve and see what feedback we get.

PR #354, opened 2021-10-08
fix: option setting updatable flag causes issues in toolkit

philasmar:
*Issue #, if available:*

*Description of changes:*
Break up Updatable flag into ReadOnly and Visible flags

### DIFF
--- a/src/AWS.Deploy.CLI/ServerMode/Controllers/DeploymentController.cs
+++ b/src/AWS.Deploy.CLI/ServerMode/Controllers/DeploymentController.cs
@@ -185,7 +185,8 @@ namespace AWS.Deploy.CLI.ServerMode.Controllers
                     TypeHint = setting.TypeHint?.ToString(),
                     Value = recommendation.GetOptionSettingValue(setting),
                     Advanced = setting.AdvancedSetting,
-                    Updatable = (!recommendation.IsExistingCloudApplication || setting.Updatable) && recommendation.IsOptionSettingDisplayable(setting),
+                    ReadOnly = recommendation.IsExistingCloudApplication && !setting.Updatable,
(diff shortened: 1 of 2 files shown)


src/AWS.Deploy.CLI/ServerMode/Controllers/DeploymentController.cs
```diff
@@ -185,7 +185,8 @@ private List<OptionSettingItemSummary> ListOptionSettingSummary(Recommendation r
                     TypeHint = setting.TypeHint?.ToString(),
                     Value = recommendation.GetOptionSettingValue(setting),
                     Advanced = setting.AdvancedSetting,
-                    Updatable = (!recommendation.IsExistingCloudApplication || setting.Updatable) && recommendation.IsOptionSettingDisplayable(setting),
+                    ReadOnly = !(!recommendation.IsExistingCloudApplication || setting.Updatable),
```
normj, line 188: No double **not**, it makes it too confusing. This should be 

```
recommendation.IsExistingCloudApplication && !setting.Updatable
```

src/AWS.Deploy.CLI/ServerMode/Controllers/DeploymentController.cs
```diff
@@ -185,7 +185,8 @@ private List<OptionSettingItemSummary> ListOptionSettingSummary(Recommendation r
...
+                    Visible = recommendation.IsOptionSettingDisplayable(setting),
                     AllowedValues = setting.AllowedValues,
```
ganeshnj, line 190: Is there a test covering this? Looks like a easily breakable code.

ganeshnj: Let's add a backlog, good for now.

PR #360, opened 2021-10-11
feat: add API to get config setting resources

philasmar:
*Issue #, if available:*
DOTNET-5450

*Description of changes:*
add API to get config setting resources

### DIFF
--- a/src/AWS.Deploy.CLI/Commands/TypeHints/BeanstalkEnvironmentCommand.cs
+++ b/src/AWS.Deploy.CLI/Commands/TypeHints/BeanstalkEnvironmentCommand.cs
@@ -1,11 +1,15 @@
 // Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
 // SPDX-License-Identifier: Apache-2.0
 
+using System.Collections.Generic;
 using System.Linq;
 using System.Threading.Tasks;
+using Amazon.ElasticBeanstalk.Model;
 using AWS.Deploy.Common;
 using AWS.Deploy.Common.Recipes;
+using AWS.Deploy.Common.TypeHintData;
 using AWS.Deploy.Orchestration.Data;
+using Newtonsoft.Json;
 
 namespace AWS.Deploy.CLI.Commands.TypeHints
 {
@@ -20,13 +24,23 @@ namespace AWS.Deploy.CLI.Commands.TypeHints
             _consoleUtilities = consoleUtilities;
         }
 
+        private async Task<List<EnvironmentDescription>> GetData(Recommendation recommendation, OptionSettingItem optionSetting)
+        {
+            var applicationOptionSetting = recommendation.GetOptionSetting(optionSetting.ParentSettingId);
+            var applicationName = recommendation.GetOptionSettingValue(applicationOptionSetting) as string;
+            return await _awsResourceQueryer.ListOfElasticBeanstalkEnvironments(applicationName);
+        }
+
+        public async Task<List<TypeHintResource>?> GetResources(Recommendation recommendation, OptionSettingItem optionSetting)
+        {
+            var environments = await GetData(recommendation, optionSetting);
+            return environments.Select(x => new TypeHintResource(x.EnvironmentName, x.EnvironmentName)).ToList();
+        }
+
         public async Task<object> Execute(Recommendation recommendation, OptionSettingItem optionSetting)
         {
             var currentValue = recommendation.GetOptionSettingValue(optionSetting);
-            var applicationOptionSetting = recommendation.GetOptionSetting(optionSetting.ParentSettingId);
-
-            var applicationName = recommendation.GetOptionSettingValue(applicationOptionSetting) as string;
-            var environments = await _awsResourceQueryer.ListOfElasticBeanstalkEnvironments(applicationName);
+            var environments = await GetData(recommendation, optionSetting);
 
             var userResponse = _consoleUtilities.AskUserToChooseOrCreateNew(
                 options: environments.Select(env => env.EnvironmentName),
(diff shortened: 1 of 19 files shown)


src/AWS.Deploy.CLI/ServerMode/Controllers/DeploymentController.cs
```diff
@@ -546,5 +582,62 @@ private Orchestrator CreateOrchestrator(SessionState state, IServiceProvider? se
                                     serviceProvider.GetRequiredService<IDirectoryManager>()
                                 );
         }
+
+        private async Task<Dictionary<string, string>?> GetResourcesByTypeHint(IAWSResourceQueryer awsResourceQueryer, Recommendation recommendation, OptionSettingItem optionSetting)
```
normj, line 586: We need to unify the CLI and ServerMode to use the same code to fetch resources. The goal should be if I want to add another simple typehint of list of AWS resources should just require an update in one area and both CLI and ServerMode get the update.

src/AWS.Deploy.CLI/ServerMode/Models/GetConfigSettingResourcesOutput.cs
```diff
@@ -0,0 +1,12 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.\r
+// SPDX-License-Identifier: Apache-2.0
+
+using System.Collections.Generic;
+
+namespace AWS.Deploy.CLI.ServerMode.Models
+{
+    public class GetConfigSettingResourcesOutput
+    {
+        public Dictionary<string, string>? Resources { get; set; }
```
normj, line 10: This should be a list of a container resource object. That object should have `DisplayName` and `SystemName` property.

src/AWS.Deploy.CLI/ServerMode/Controllers/DeploymentController.cs
philasmar: Done

src/AWS.Deploy.CLI/ServerMode/Models/GetConfigSettingResourcesOutput.cs
philasmar: Done

src/AWS.Deploy.CLI/Commands/TypeHints/BeanstalkEnvironmentCommand.cs
```diff
@@ -20,13 +24,24 @@ public BeanstalkEnvironmentCommand(IAWSResourceQueryer awsResourceQueryer, ICons
...
 
-        public async Task<object> Execute(Recommendation recommendation, OptionSettingItem optionSetting)
+        private async Task<T> GetData<T>(Recommendation recommendation, OptionSettingItem optionSetting)
         {
-            var currentValue = recommendation.GetOptionSettingValue(optionSetting);
             var applicationOptionSetting = recommendation.GetOptionSetting(optionSetting.ParentSettingId);
-
             var applicationName = recommendation.GetOptionSettingValue(applicationOptionSetting) as string;
             var environments = await _awsResourceQueryer.ListOfElasticBeanstalkEnvironments(applicationName);
+            return JsonConvert.DeserializeObject<T>(JsonConvert.SerializeObject(environments));
```
normj, line 32: This looks very silly to serialize and then immediately deserialize the object. Especially since I'm pretty sure we are always doing this to the same type of `EnvironmentDescription`. I think you are copying and pasting this pattern to all of the type hint commands but it looks really weird and we should fix this up.

src/AWS.Deploy.CLI/Commands/TypeHints/BeanstalkEnvironmentCommand.cs
```diff
@@ -20,13 +24,24 @@ public BeanstalkEnvironmentCommand(IAWSResourceQueryer awsResourceQueryer, ICons
...
+        }
+
+        public async Task<List<TypeHintResource>?> GetServerModeResource(Recommendation recommendation, OptionSettingItem optionSetting)
```
normj, line 35: I don't like calling these `ServerMode` implying they couldn't be used for anything else. Can we just call this method `GetResources`?

src/AWS.Deploy.CLI/ServerMode/Controllers/DeploymentController.cs
```diff
@@ -241,6 +244,45 @@ public IActionResult ApplyConfigSettings(string sessionId, [FromBody] ApplyConfi
...
+            var state = _stateServer.Get(sessionId);
+            if (state == null)
+            {
+                return NotFound($"Session ID {sessionId} not found.");
+            }
+
+            var serviceProvider = CreateSessionServiceProvider(state);
+            var typeHintCommandFactory = serviceProvider.GetRequiredService<ITypeHintCommandFactory>();
+
+            if (state.SelectedRecommendation == null)
```
normj, line 262: nit: Do this right after the `if` block on 254. Seems like we should start the method validating state. Once that is validated then create the session service provider and such.

src/AWS.Deploy.CLI/ServerMode/Controllers/DeploymentController.cs
```diff
@@ -241,6 +244,45 @@ public IActionResult ApplyConfigSettings(string sessionId, [FromBody] ApplyConfi
...
+            {
+                return NotFound($"A deployment target is not set for Session ID {sessionId}.");
+            }
+
+            var configSetting = state.SelectedRecommendation.GetOptionSetting(configSettingId);
+
+            if (configSetting.TypeHint.HasValue && typeHintCommandFactory.GetCommand(configSetting.TypeHint.Value) is var typeHintCommand && typeHintCommand != null)
+            {
+                var output = new GetConfigSettingResourcesOutput();
+                var resources = await typeHintCommand.GetServerModeResource(state.SelectedRecommendation, configSetting);
```
normj, line 272: Why would resources list ever be null?

src/AWS.Deploy.CLI/ServerMode/Controllers/DeploymentController.cs
philasmar: Done

PR #378, opened 2021-10-15
fix: last deployed stack local file is overwritten when different apps are used

philasmar:
*Issue #, if available:*
IDE-6121

*Description of changes:*
fix: last deployed stack local file is overwritten when different apps are used


src/AWS.Deploy.Orchestration/LocalUserSettings/LocalUserSettingsEngine.cs
```diff
@@ -67,12 +67,23 @@ public async Task UpdateLastDeployedStack(string stackName, string projectName,
                     }
                     else
                     {
-                        localUserSettings.LastDeployedStacks = new List<LastDeployedStack>() {
-                            new LastDeployedStack(
-                                awsAccountId,
-                                awsRegion,
-                                projectName,
-                                new List<string>() { stackName })};
+                        var currentStack = new LastDeployedStack(
```
normj, line 70: This object should be created inside the if block that is actually using it.

src/AWS.Deploy.Orchestration/LocalUserSettings/LocalUserSettingsEngine.cs
```diff
@@ -67,12 +67,23 @@ public async Task UpdateLastDeployedStack(string stackName, string projectName,
...
+                            awsRegion,
+                            projectName,
+                            new List<string>()
+                            {
+                                stackName
+                            });
+
+                        if (localUserSettings.LastDeployedStacks == null)
+                        {
+                            localUserSettings.LastDeployedStacks = new List<LastDeployedStack>() { currentStack };
```
ganeshnj, line 81: is it worth changing `LastDeployedStacks` to non-nullable with constructor initialization or with default empty collection?

philasmar: Done

philasmar:
> Is this possible to add a unit test using mocks?

Done

src/AWS.Deploy.Orchestration/LocalUserSettings/LocalUserSettingsEngine.cs
ganeshnj: you can remove the whole null check now

PR #399, opened 2021-11-05
feat: improve telemetry and error visibility in server mode

philasmar:
*Issue #, if available:*
DOTNET-5509
*Description of changes:*
The expected exceptions now include an error code that represents a failure. Error codes are not a 1:1 map with exceptions. 1 exception used in multiple places could have different error codes.
all expected exceptions now inherit from a Deploy Tool base exception, and that base exception is checked in the server mode controller GetDeploymentStatus.
We check if an exception occurred during the execution of the DeploymentTask and we surface it in the API.

### DIFF
--- a/src/AWS.Deploy.Common/Exceptions.cs
+++ b/src/AWS.Deploy.Common/Exceptions.cs
@@ -8,171 +8,223 @@ using AWS.Deploy.Common.Recipes.Validation;
 
 namespace AWS.Deploy.Common
 {
-    public class ProjectFileNotFoundException : Exception
+    public abstract class DeployToolException : Exception
     {
-        public ProjectFileNotFoundException(string projectPath)
-            : base($"A project was not found at the path {projectPath}.")
-        {
-            Path = projectPath;
-        }
+        public DeployToolErrorCode ErrorCode { get; set; }
 
-        public string Path { get; set; }
+        public DeployToolException(DeployToolErrorCode errorCode, string message, Exception? innerException = null) : base(message, innerException)
+        {
+            ErrorCode = errorCode;
+        }
+    }
+
+    public enum DeployToolErrorCode
+    {
+        ProjectPathNotFound = 10000100,
+        ProjectParserNoSdkAttribute = 10000200,
+        InvalidCliArguments = 10000300,
+        SilentArgumentNeedsStackNameArgument = 10000400,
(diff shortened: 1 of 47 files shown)


96malhar:
Since all expected exceptions are now identified using the `DeployToolException` base class, can we remove the  `AWSDeploymentExpectedException` attribute?

src/AWS.Deploy.Common/Exceptions.cs
```diff
@@ -8,148 +8,230 @@
 
 namespace AWS.Deploy.Common
 {
-    public class ProjectFileNotFoundException : Exception
+    [AWSDeploymentExpectedException]
+    public class DeployToolException : Exception
```
96malhar, line 12: Since we are never directly initializing an object of `DeployToolException`, can we mark this as an abstract class ?

philasmar: Done

philasmar:
> Since all expected exceptions are now identified using the `DeployToolException` base class, can we remove the `AWSDeploymentExpectedException` attribute?

I agree, it seems redundant at this point. I removed it.

96malhar (approved):

src/AWS.Deploy.Common/Exceptions.cs
```diff
@@ -8,171 +8,223 @@
...
+
+        public DeployToolException(DeployToolErrorCode errorCode, string message, Exception? innerException = null) : base(message, innerException)
         {
-            Path = projectPath;
+            ErrorCode = errorCode;
         }
+    }
 
-        public string Path { get; set; }
+    public enum DeployToolErrorCode
```
ganeshnj, line 21: This seems very fragile error code number system, if every a developer adds an enum value in between, everything after that value will get messed up.

Either we should make then int constants or atleast provide documentation that new values **must** be added at the end only

ganeshnj: Also, these enums will not be backward compatible.

ganeshnj (changes requested):

src/AWS.Deploy.Common/Exceptions.cs
philasmar: Added int values to enums and casted to string when passing through API. The client now accepts a string and has no knowledge of the enum.

ganeshnj (approved):

PR #408, opened 2021-11-23
feat: add health check url setting to beanstalk deployments

philasmar:
*Issue #, if available:*
DOTNET-4808

*Description of changes:*
add health check url setting to beanstalk deployments

### DIFF
--- a/src/AWS.Deploy.Recipes/CdkTemplates/AspNetAppElasticBeanstalkLinux/Generated/Recipe.cs
+++ b/src/AWS.Deploy.Recipes/CdkTemplates/AspNetAppElasticBeanstalkLinux/Generated/Recipe.cs
@@ -183,6 +183,27 @@ namespace AspNetAppElasticBeanstalkLinux
                         Value = settings.LoadBalancerType
                     }
                 );
+
+                if (!string.IsNullOrEmpty(settings.HealthCheckURL))
+                {
+                    optionSettingProperties.Add(
+                        new CfnEnvironment.OptionSettingProperty
+                        {
+                            Namespace = "aws:elasticbeanstalk:application",
+                            OptionName = "Application Healthcheck URL",
+                            Value = settings.HealthCheckURL
+                        }
+                    );
+
+                    optionSettingProperties.Add(
+                        new CfnEnvironment.OptionSettingProperty
+                        {
+                            Namespace = "aws:elasticbeanstalk:environment:process:default",
+                            OptionName = "HealthCheckPath",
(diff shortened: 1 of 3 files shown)


src/AWS.Deploy.Recipes/CdkTemplates/AspNetAppElasticBeanstalkLinux/Generated/Recipe.cs
```diff
@@ -214,6 +214,27 @@ private void ConfigureBeanstalkEnvironment(Configuration settings)
                 });
             }
 
+            if (!string.IsNullOrEmpty(settings.HealthCheckURL))
```
normj, line 217: We should be extra defensive and only set health check url if doing a load balanced deployment. Don't just rely on the recipe `depends` criteria.

normj (changes requested):

src/AWS.Deploy.Recipes/RecipeDefinitions/ASP.NETAppElasticBeanstalk.recipe
```diff
@@ -329,6 +329,21 @@
                     ]
                 }
             ]
+        },
+        {
+            "Id": "HealthCheckURL",
+            "Name": "Health Check URL",
+            "Description": "Customize the load balancer health check to ensure that your application, and not just the web server, is in a good state",
```
96malhar, line 336: Can you update the description to reflect that this OptionSettingItem is only required if the environment type is load balanced?

normj: Since the option will only be displayed if the environment type is load balanced I don't think we need to make the description extra wordy.

src/AWS.Deploy.Recipes/CdkTemplates/AspNetAppElasticBeanstalkLinux/Generated/Recipe.cs
philasmar: Done

src/AWS.Deploy.Recipes/RecipeDefinitions/ASP.NETAppElasticBeanstalk.recipe
philasmar: I agree with Norm. This is only displayed when the env is load balanced.

normj (approved):

96malhar (approved):

PR #410, opened 2021-11-24
feat: add environment variables config to beanstalk deployments

philasmar:
*Issue #, if available:*
DOTNET-4809

*Description of changes:*
add environment variables config to beanstalk deployments
PR410

Introduce new KeyValue type that supports a Dictionary.
The KeyValue type is Dictionary that serializes the same way as Object types are defined in config files that users can deploy from. This makes it harder to distinguish between an Object and a Dictionary.
image
In the image above, one is an Object and the other is a Dictionary. For this reason, I have updated the GetOptionSetting method to check if the OptionSettingItem is a KeyValue type to return it.

Updated UX to ask user for key value pairs with ability to Add, Update and Delete

### DIFF
--- a/src/AWS.Deploy.CLI/Commands/DeployCommand.cs
+++ b/src/AWS.Deploy.CLI/Commands/DeployCommand.cs
@@ -369,6 +369,9 @@ namespace AWS.Deploy.CLI.Commands
 
                 var optionSetting = recommendation.GetOptionSetting(optionSettingJsonPath);
 
+                if (optionSetting == null)
+                    throw new OptionSettingItemDoesNotExistException(DeployToolErrorCode.OptionSettingItemDoesNotExistInRecipe, $"The Option Setting Item {optionSettingJsonPath} does not exist.");
+
                 if (!recommendation.IsExistingCloudApplication || optionSetting.Updatable)
                 {
                     object settingValue;
@@ -387,6 +390,13 @@ namespace AWS.Deploy.CLI.Commands
                                 break;
                             case OptionSettingValueType.Double:
                                 settingValue = double.Parse(optionSettingValue);
+                                break;
+                            case OptionSettingValueType.KeyValue:
+                                var optionSettingKey = optionSettingJsonPath.Split(".").Last();
+                                var existingValue = recommendation.GetOptionSettingValue<Dictionary<string, string>>(optionSetting);
+                                existingValue ??= new Dictionary<string, string>();
+                                existingValue[optionSettingKey] = optionSettingValue;
+                                settingValue = existingValue;
                                 break;
                             default:
                                 throw new InvalidOverrideValueException(DeployToolErrorCode.InvalidValueForOptionSettingItem, $"Invalid value {optionSettingValue} for option setting item {optionSettingJsonPath}");
@@ -772,6 +782,9 @@ namespace AWS.Deploy.CLI.Commands
                             var answer = _consoleUtilities.AskYesNoQuestion(string.Empty, recommendation.GetOptionSettingValue(setting).ToString());
                             settingValue = answer == YesNo.Yes ? "true" : "false";
                             break;
+                        case OptionSettingValueType.KeyValue:
+                            settingValue = _consoleUtilities.AskUserForKeyValue(!string.IsNullOrEmpty(currentValue.ToString()) ? (Dictionary<string, string>) currentValue : new Dictionary<string, string>());
+                            break;
                         case OptionSettingValueType.Object:
                             foreach (var childSetting in setting.ChildOptionSettings)
                             {
@@ -809,6 +822,7 @@ namespace AWS.Deploy.CLI.Commands
         private void DisplayValue(Recommendation recommendation, OptionSettingItem optionSetting, int optionSettingNumber, int optionSettingsCount, Type? typeHintResponseType, DisplayOptionSettingsMode mode)
         {
             object? displayValue = null;
+            Dictionary<string, string>? keyValuePair = null;
             Dictionary<string, object>? objectValues = null;
             if (typeHintResponseType != null)
             {
@@ -823,7 +837,8 @@ namespace AWS.Deploy.CLI.Commands
             {
                 var value = recommendation.GetOptionSettingValue(optionSetting);
                 objectValues = value as Dictionary<string, object>;
-                displayValue = objectValues == null ? value : string.Empty;
+                keyValuePair = value as Dictionary<string, string>;
+                displayValue = objectValues == null && keyValuePair == null ? value : string.Empty;
             }
 
             if (mode == DisplayOptionSettingsMode.Editable)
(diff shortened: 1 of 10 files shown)


src/AWS.Deploy.CLI/ConsoleUtilities.cs
```diff
@@ -306,6 +307,68 @@ public string AskUserForValue(string message, string defaultValue, bool allowEmp
             return userValue;
         }
 
+        public Dictionary<string, string> AskUserForKeyValue(Dictionary<string, string> environmentVariables)
```
96malhar, line 310: Is this method only going to be used to prompt for environment variables or any general key-value pair? If it is general key-value pairs as well then we should change the argument name from `environmentVariables` to something generic.

If it is only used for prompt for environment variables, then change the function name to `AskForEnvironmentVariable`

96malhar:
Can you add a new config to https://github.com/aws/aws-dotnet-deploy/tree/main/test/AWS.Deploy.CLI.Common.UnitTests/ConfigFileDeployment/TestFiles/UnitTestFiles and add a unit test with a config file that has a key-value pair.

Make sure you add a **new** config file and not edit an existing one, since the existing files are also used in integ tests

src/AWS.Deploy.CLI/ConsoleUtilities.cs
philasmar: Done

src/AWS.Deploy.CLI/ConsoleUtilities.cs
```diff
@@ -306,6 +307,68 @@ public string AskUserForValue(string message, string defaultValue, bool allowEmp
...
+
+            if (keyValue.Keys.Count == 0)
+            {
+                AskToAddKeyValuePair(keyValue);
+                return keyValue;
+            }
+
+            const string ADD = "Add new environment variable";
+            const string UPDATE = "Update existing environment variable";
+            const string DELETE = "Delete environment variable";
```
96malhar, line 322: These prompts strictly refer to environment variables but are declared as constants inside the generic method to ask for key-value pairs. These message prompts should be parameters in the method signature and will change depending on the context of the key-value pairs.

src/AWS.Deploy.CLI/ConsoleUtilities.cs
```diff
@@ -306,6 +307,68 @@ public string AskUserForValue(string message, string defaultValue, bool allowEmp
...
+                string.Equals($"'{RESET}'", variableValue.Trim(), StringComparison.OrdinalIgnoreCase))
+                variableValue = keyValue.ContainsKey(variableName) ? keyValue[variableName] : "";
+
+            keyValue[variableName] = variableValue;
+        }
+
+        private void AskToUpdateKeyValuePair(Dictionary<string, string> keyValue)
+        {
+            var selectedKey = AskUserToChoose(keyValue.Keys.ToList(), "Select the environment variable you wish to update:", null);
+            var selectedValue = AskUserForValue("Enter the environment variable value:", keyValue[selectedKey], true);
```
96malhar, line 361: Refer to the above comment on message prompts.

src/AWS.Deploy.CLI/ConsoleUtilities.cs
```diff
@@ -306,6 +307,68 @@ public string AskUserForValue(string message, string defaultValue, bool allowEmp
...
+
+            keyValue[selectedKey] = selectedValue;
+        }
+
+        private void AskToDeleteKeyValuePair(Dictionary<string, string> keyValue)
+        {
+            var selectedKey = AskUserToChoose(keyValue.Keys.ToList(), "Select the environment variable you wish to delete:", null);
```
96malhar, line 368: Refer to the above comment on message prompts.

src/AWS.Deploy.CLI/ConsoleUtilities.cs
```diff
@@ -306,6 +307,68 @@ public string AskUserForValue(string message, string defaultValue, bool allowEmp
...
+        {
+            const string RESET = "<reset>";
+            var variableName = string.Empty;
+            while (string.IsNullOrEmpty(variableName))
+            {
+                _interactiveService.WriteLine("Enter the environment variable name:");
+                variableName = _interactiveService.ReadLine()?.Trim() ?? "";
+            }
+
+            _interactiveService.WriteLine($"Enter the environment variable value (type {RESET} to reset):");
```
96malhar, line 349: Refer to the above comment on message prompts.

src/AWS.Deploy.CLI/ConsoleUtilities.cs
philasmar: Done

src/AWS.Deploy.CLI/ConsoleUtilities.cs
philasmar: Done

src/AWS.Deploy.CLI/ConsoleUtilities.cs
philasmar: Done

96malhar (approved):

src/AWS.Deploy.CLI/Commands/DeployCommand.cs
```diff
@@ -365,9 +365,26 @@ private void ConfigureDeploymentFromConfigFile(Recommendation recommendation, Us
...
 
-                var optionSetting = recommendation.GetOptionSetting(optionSettingJsonPath);
+                OptionSettingItem? optionSetting = null;
+                try
+                {
+                    optionSetting = recommendation.GetOptionSetting(optionSettingJsonPath);
+                }
+                catch (OptionSettingItemDoesNotExistException)
+                {
+                    var optionSettingSplit = optionSettingJsonPath.Split(".");
```
normj, line 378: This needs a comment. I have no idea what this code is here for.

normj: From what I can gather the logic attempt to get an `optionSetting`. If it fails then it is likely a KeyValuePair and do some extra logic for for determing `optionSettingKey`.

Its not good to use control flow by exceptions and having a optionSettingKey variable that is there for all types but only used if it is a KVP adds even more confusion to the code. This method is getting to frail relying on subtle behavior with nothing documented.

src/AWS.Deploy.Recipes/RecipeDefinitions/ASP.NETAppElasticBeanstalk.recipe
```diff
@@ -387,6 +387,14 @@
             ],
             "AdvancedSetting": true,
             "Updatable": true
+        },
+        {
+            "Id": "ElasticBeanstalkEnvironmentVariables",
+            "Name": "Environment Variables",
+            "Description": "Configure environment properties for your application.",
+            "Type": "KeyValue",
+            "AdvancedSetting": true,
```
normj, line 396: Lets make environment variables **not** be advanced.

normj (changes requested):
The description of this PR should have a description of how you added key value pairs. What were the challenges and how did you solve them? You are adding an architecture change to the product so treat the description as the spec for the change.

Also since you are adding a new UX workflow there should be an animated gif of the experience.

src/AWS.Deploy.CLI/Commands/DeployCommand.cs
philasmar: This code has been revised and moved to GetOptionSetting

src/AWS.Deploy.Recipes/RecipeDefinitions/ASP.NETAppElasticBeanstalk.recipe
philasmar: done

normj (approved):

PR #414, opened 2021-12-02
feat: Add type hints on child settings

philasmar:
*Issue #, if available:*
DOTNET-5570
*Description of changes:*
* Add type hints on child settings to enable visual studio toolkit to query the child typehint and return the resources that can be selected. The CLI will still use the parent type hint to display the Object type. The toolkit can decide to use either.
* Added Typehint data for S3BucketName, DynamoDbTable, SNSTopicArns and SQSQueueUrl to choose whether the typehint list allows empty values or not.
* Added a filtering mechanism to retrieve instance types
PR414

### DIFF
--- a/src/AWS.Deploy.CLI/Commands/TypeHints/S3BucketNameCommand.cs
+++ b/src/AWS.Deploy.CLI/Commands/TypeHints/S3BucketNameCommand.cs
@@ -39,9 +39,12 @@ namespace AWS.Deploy.CLI.Commands.TypeHints
         {
             const string NO_VALUE = "*** Do not select bucket ***";
             var currentValue = recommendation.GetOptionSettingValue(optionSetting);
+            var typeHintData = optionSetting.GetTypeHintData<S3BucketNameTypeHintData>();
             var buckets = (await GetData()).Select(bucket => bucket.BucketName).ToList();
 
-            buckets.Add(NO_VALUE);
+            if (typeHintData?.AllowNoValue ?? false)
+                buckets.Add(NO_VALUE);
+
             var userResponse = _consoleUtilities.AskUserToChoose(
                 values: buckets,
                 title: "Select a S3 bucket:",
--- a/src/AWS.Deploy.Orchestration/Data/AWSResourceQueryer.cs
+++ b/src/AWS.Deploy.Orchestration/Data/AWSResourceQueryer.cs
@@ -39,6 +39,7 @@ namespace AWS.Deploy.Orchestration.Data
 {
     public interface IAWSResourceQueryer
     {
+        Task<List<InstanceTypeInfo>> ListOfAvailableInstanceTypes();
         Task<Amazon.AppRunner.Model.Service> DescribeAppRunnerService(string serviceArn);
         Task<List<StackResource>> DescribeCloudFormationResources(string stackName);
         Task<EnvironmentDescription> DescribeElasticBeanstalkEnvironment(string environmentId);
@@ -76,6 +77,20 @@ namespace AWS.Deploy.Orchestration.Data
         public AWSResourceQueryer(IAWSClientFactory awsClientFactory)
         {
             _awsClientFactory = awsClientFactory;
+        }
+
+        public async Task<List<InstanceTypeInfo>> ListOfAvailableInstanceTypes()
+        {
+            var ec2Client = _awsClientFactory.GetAWSClient<IAmazonEC2>();
+            var instanceTypes = new List<InstanceTypeInfo>();
+            var listInstanceTypesPaginator = ec2Client.Paginators.DescribeInstanceTypes(new DescribeInstanceTypesRequest());
+
+            await foreach (var response in listInstanceTypesPaginator.Responses)
+            {
+                instanceTypes.AddRange(response.InstanceTypes);
+            }
+
+            return instanceTypes;
         }
 
         public async Task<Amazon.AppRunner.Model.Service> DescribeAppRunnerService(string serviceArn)
--- a/src/AWS.Deploy.Recipes/RecipeDefinitions/ASP.NETAppElasticBeanstalk.recipe
+++ b/src/AWS.Deploy.Recipes/RecipeDefinitions/ASP.NETAppElasticBeanstalk.recipe
@@ -98,6 +98,7 @@
                     "Name": "Application Name",
                     "Description": "The Elastic Beanstalk application name.",
                     "Type": "String",
+                    "TypeHint": "ExistingBeanstalkApplication",
                     "DefaultValue": "{StackName}",
                     "AdvancedSetting": false,
                     "Updatable": false,
@@ -139,7 +140,6 @@
             "Description": "The EC2 instance type of the EC2 instances created for the environment.",
             "Type": "String",
             "TypeHint": "InstanceType",
-            "DefaultValue": "",
             "AdvancedSetting": true,
             "Updatable": true
         },
@@ -211,6 +211,10 @@
                     "Name": "Existing Role ARN",
                     "Description": "The ARN of the existing role to use.",
                     "Type": "String",
+                    "TypeHint": "ExistingIAMRole",
+                    "TypeHintData": {
+                        "ServicePrincipal": "elasticbeanstalk.amazonaws.com"
+                    },
                     "AdvancedSetting": false,
                     "Updatable": false,
                     "DependsOn": [
--- a/src/AWS.Deploy.CLI/Commands/TypeHints/DynamoDBTableCommand.cs
+++ b/src/AWS.Deploy.CLI/Commands/TypeHints/DynamoDBTableCommand.cs
@@ -38,9 +38,12 @@ namespace AWS.Deploy.CLI.Commands.TypeHints
         {
             const string NO_VALUE = "*** Do not select table ***";
             var currentValue = recommendation.GetOptionSettingValue(optionSetting);
+            var typeHintData = optionSetting.GetTypeHintData<DynamoDBTableTypeHintData>();
             var tables = await GetData();
 
-            tables.Add(NO_VALUE);
+            if (typeHintData?.AllowNoValue ?? false)
+                tables.Add(NO_VALUE);
+
             var userResponse = _consoleUtilities.AskUserToChoose(
                 values: tables,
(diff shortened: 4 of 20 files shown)


src/AWS.Deploy.Orchestration/Data/AWSResourceQueryer.cs
```diff
@@ -78,6 +81,96 @@ public AWSResourceQueryer(IAWSClientFactory awsClientFactory)
...
+                    }
+                }
+            });
+
+            await foreach (var response in listInstanceTypesPaginator.Responses)
+            {
+                instanceTypes.AddRange(response.InstanceTypes);
+            }
+
+            return instanceTypes.Select(x => (double) x.MemoryInfo.SizeInMiB).Distinct().OrderBy(x => x).ToList();
```
96malhar, line 137: Bases on the API ref, the `SizeInMiB` is an `int`. Why are we casting it into a double?

API ref - https://docs.aws.amazon.com/sdkfornet/v3/apidocs/items/EC2/TMemoryInfo.html

src/AWS.Deploy.Recipes/RecipeDefinitions/ASP.NETAppElasticBeanstalk.recipe
```diff
@@ -139,11 +140,67 @@
...
+                    ],
+                    "AdvancedSetting": true,
+                    "Updatable": true
+                },
+                {
+                    "Id": "InstanceCPUCores",
+                    "ParentSettingId": "InstanceType",
+                    "Name": "EC2 Instance CPU Cores",
+                    "Description": "The CPU cores of the EC2 instances created for the environment.",
+                    "Type": "Int",
```
96malhar, line 175: Can you add a range validator for the setting and add a validator unit test?

src/AWS.Deploy.Recipes/RecipeDefinitions/ASP.NETAppElasticBeanstalk.recipe
```diff
@@ -139,11 +140,67 @@
...
+                    "TypeHint": "InstanceCPUCores",
+                    "AdvancedSetting": true,
+                    "Updatable": true
+                },
+                {
+                    "Id": "InstanceMemory",
+                    "ParentSettingId": "InstanceType",
+                    "Name": "EC2 Instance Memory",
+                    "Description": "The memory in MB of the EC2 instances created for the environment.",
+                    "Type": "Double",
```
96malhar, line 186: Can you add a range validator for the setting and add a validator unit test?

src/AWS.Deploy.Orchestration/Data/AWSResourceQueryer.cs
philasmar: In the API ref you linked, it says that it is an Int64 which is mapped to a `long`. I casted the `long` to `double` since we currently support the type `Double` in our recipes.

src/AWS.Deploy.Recipes/RecipeDefinitions/ASP.NETAppElasticBeanstalk.recipe
philasmar: The CPU cores type is dynamically calculated based on the type hint. It is selected from a range of values presented to the end users. It is also not a real option setting item, it is only used to filter the INstance Type. There is no predefined list of values to set as a validator.

src/AWS.Deploy.Recipes/RecipeDefinitions/ASP.NETAppElasticBeanstalk.recipe
philasmar: The instance memory is dynamically calculated based on the type hint. It is selected from a range of values presented to the end users. It is also not a real option setting item, it is only used to filter the Instance Type. There is no predefined list of values to set as a validator.

src/AWS.Deploy.Recipes/RecipeDefinitions/ASP.NETAppElasticBeanstalk.recipe
96malhar: Currently, the type hints are not invoked if a user choses to perform a deployment via a configuration file. It is possible for a user to supply an arbitrarily large value or negative values in the config. Having a simple range validator will throw an actionable error before the CDK hand-off happens.

src/AWS.Deploy.Recipes/RecipeDefinitions/ASP.NETAppElasticBeanstalk.recipe
96malhar: Currently, the type hints are not invoked if a user choses to perform a deployment via a configuration file. It is possible for a user to supply an arbitrarily large value or negative values in the config. Having a simple range validator will throw an actionable error before the CDK hand-off happens.

src/AWS.Deploy.Recipes/RecipeDefinitions/ASP.NETAppElasticBeanstalk.recipe
philasmar: even if the user sets an invalid valid, this does not affect anything. These are filtering OSI, not actual OSI. They don't have a predetermined value, so we cannot create a random range to validate.

src/AWS.Deploy.Recipes/RecipeDefinitions/ASP.NETAppElasticBeanstalk.recipe
philasmar: even if the user sets an invalid valid, this does not affect anything. These are filtering OSI, not actual OSI. They don't have a predetermined value, so we cannot create a random range to validate.

src/AWS.Deploy.CLI/Commands/TypeHints/InstanceCPUCoresCommand.cs
```diff
@@ -0,0 +1,64 @@
...
+        {
+            _awsResourceQueryer = awsResourceQueryer;
+            _consoleUtilities = consoleUtilities;
+        }
+
+        private async Task<List<int>?> GetData(Recommendation recommendation, OptionSettingItem optionSetting)
+        {
+            var parentSetting = recommendation.GetOptionSetting(optionSetting.ParentSettingId);
+
+            var freeTierEligible = parentSetting.ChildOptionSettings.FirstOrDefault(x => x.Id.Equals(INSTANCE_FREE_TIER_ELIGIBLE));
```
normj, line 31: We shouldn't have string value type hint that is relying on other sibling settings. What if a recipe wants to use the instance type typehint but it doesn't define an `InstanceFreeTierEligible` or `InstanceArchitecture`

src/AWS.Deploy.Recipes/RecipeDefinitions/ASP.NETAppElasticBeanstalk.recipe
```diff
@@ -139,11 +140,67 @@
...
-            "Type": "String",
+            "Type": "Object",
             "TypeHint": "InstanceType",
-            "DefaultValue": "",
             "AdvancedSetting": true,
-            "Updatable": true
+            "Updatable": true,
+            "ChildOptionSettings": [
+                {
+                    "Id": "InstanceFreeTierEligible",
```
normj, line 149: I don't think this is right to add these settings like InstanceFreeTierEligible, InstanceArchitecture, InstanceCPUCores and InstanceMemory that are not actual settings by filter for the typehint.

src/AWS.Deploy.CLI/Commands/TypeHints/S3BucketNameCommand.cs
```diff
@@ -37,17 +37,15 @@ private async Task<List<S3Bucket>> GetData()
 
         public async Task<object> Execute(Recommendation recommendation, OptionSettingItem optionSetting)
         {
-            const string NO_VALUE = "*** Do not select bucket ***";
             var currentValue = recommendation.GetOptionSettingValue(optionSetting);
             var buckets = (await GetData()).Select(bucket => bucket.BucketName).ToList();
 
-            buckets.Add(NO_VALUE);
```
normj, line 44: Why did you remove the ability to not select a bucket?

src/AWS.Deploy.Recipes/RecipeDefinitions/ASP.NETAppElasticBeanstalk.recipe
```diff
@@ -139,11 +140,67 @@
             "Id": "InstanceType",
             "Name": "EC2 Instance Type",
             "Description": "The EC2 instance type of the EC2 instances created for the environment.",
-            "Type": "String",
+            "Type": "Object",
```
normj, line 143: This is a breaking change to any existing users. I know we are in preview but we should give this serious thought on whether we should be making breaking changes.

normj (changes requested):
The PR should have an animated gif showing the experience of the new type hints. Especially the instance type. As I see the code I'm concerned with the approach.

src/AWS.Deploy.CLI/Commands/TypeHints/InstanceCPUCoresCommand.cs
philasmar: I removed the extra settings

src/AWS.Deploy.Recipes/RecipeDefinitions/ASP.NETAppElasticBeanstalk.recipe
philasmar: Done

src/AWS.Deploy.CLI/Commands/TypeHints/S3BucketNameCommand.cs
philasmar: Added typehint data like we discussed

src/AWS.Deploy.Recipes/RecipeDefinitions/ASP.NETAppElasticBeanstalk.recipe
philasmar: Reverted

96malhar (approved):

PR #426, opened 2022-01-12
fix: InstanceType hint returns 500 in server mode

philasmar:
*Issue #, if available:*
DOTNET-5622

*Description of changes:*
Added ability to attach to running server mode to simplify debugging
Fixed bug with InstanceType type hint that was returning a 500 error in server mode due to sorting a list incorrectly.
Previously, using linq, we were converting the Instance types into a TypeHintResource and then doing an OrderBy. This was failing because we dont have a comparator for TypeHintResource. To fix this, I simply did the OrderBy before creating the TypeHintResource.

### DIFF
--- a/src/AWS.Deploy.ServerMode.Client/ServerModeSession.cs
+++ b/src/AWS.Deploy.ServerMode.Client/ServerModeSession.cs
@@ -186,6 +186,27 @@ namespace AWS.Deploy.ServerMode.Client
             return true;
         }
 
+        /// <summary>
+        /// Builds <see cref="IRestAPIClient"/> client based on a deploy tool server mode running on the specified port.
+        /// </summary>
+        public static bool TryGetRestAPIClient(int port, Aes? aes, Func<Task<AWSCredentials>> credentialsGenerator, out IRestAPIClient? restApiClient)
+        {
+            // This ensures that deploy tool CLI doesn't try on the in-use port
+            // because server availability task will return success response for
+            // an in-use port
+            if (!IsPortInUse(port))
+            {
+                restApiClient = null;
+                throw new PortUnavailableException($"There is no running process on port {port}.");
+            }
+
+            var baseUrl = $"http://localhost:{port}";
+
+            var httpClient = ServerModeHttpClientFactory.ConstructHttpClient(credentialsGenerator, aes);
+            restApiClient = new RestAPIClient(baseUrl, httpClient);
+            return true;
+        }
+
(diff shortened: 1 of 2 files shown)


src/AWS.Deploy.ServerMode.Client/ServerModeSession.cs
```diff
@@ -112,6 +129,30 @@ public ServerModeSession(int startPort = 10000, int endPort = 10100, string depl
             _deployToolPath = deployToolPath;
         }
 
+        public async Task AttachToRunnningSession(int port, CancellationToken cancellationToken)
```
normj, line 132: I'm uncomfortable having this method since it will be useless in the normal mode of server-mode running with encryption. It implies something more powerful then it really can do. Is the goal of this method just to make sure `_baseUrl` is set? If so maybe have `TryGetRestAPIClient` have a named parameter for `baseUrl`.

normj (changes requested):
In the PR description you should have describe how the list was incorrectly being sorted.

src/AWS.Deploy.ServerMode.Client/ServerModeSession.cs
philasmar: I removed this method and moved some of the inner logic to the new `TryGetUnsecureRestAPIClient`. I did not want to add `baseUrl` as a named parameter as that is created in the `Start` method where we go over a bunch of available ports. I didn't want to make the user responsible for that. Plus adding that would have been a breaking change.

src/AWS.Deploy.ServerMode.Client/ServerModeSession.cs
```diff
@@ -186,6 +227,19 @@ public bool TryGetRestAPIClient(Func<Task<AWSCredentials>> credentialsGenerator,
...
 
+        public bool TryGetUnsecureRestAPIClient(Func<Task<AWSCredentials>> credentialsGenerator, out IRestAPIClient? restApiClient)
+        {
+            if (_baseUrl == null)
+            {
+                restApiClient = null;
+                return false;
+            }
+
+            var httpClient = ServerModeHttpClientFactory.ConstructHttpClient(credentialsGenerator);
```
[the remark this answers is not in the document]
normj: But you aren't launching server mode you are connecting with an already started instance.

normj (changes requested):

normj (approved):

96malhar (approved):

PR #427, opened 2022-01-14
chore: Refactor the code to support non-cloudformation based deployments

96malhar:
**Note** - The CI pipeline is not automatically run when trying to merge a PR into a staging branch. I manually triggered the `aws-dotnet-deploy-ci` CodeBuild job and it ran successfully (build 204, 213, 219).

*Issue #, if available:*
DOTNET-5510

*Description of changes:*

* Replaced the `--stack-name` CLI argument to `--application-name` since our deployment targets are no longer restricted to CloudFormation stacks.
* Amended property names, user prompts, error messages and comments to reflect that our deployment targets are no longer restricted to CloudFormation stacks.
* Moved existing and added new constants to the `AWS.Deploy.Constants` shared project.
* Made the following changes to the `CloudApplication` class
   * Removed the `StackName` property as not all `CloudApplication` objects will be CloudFormation stacks
   * Added a new `UniqueIdentifier` property that can either be CloudFormation StackIDs or Elastic Beanstalk Environment ARNs.  For new deployments this property is set to `string.Empty`
   * Added a new `ResourceType` property that indicates the type of the AWS resource which serves as the deployment target. Current supported values are `CloudFormationStack`, `BeanstalkEnvironment` and `None`
   * Added a new `DisplayName` property which is shown to the user when the `CloudApplication` is presented as an existing re-deployment target.
* Extended the `DeploymentTypes` enum to to include `BeanstalkEnvironment`
* Added a boolean property `DisableNewDeployments` to `RecipeDefinition` class. If set to true then this recipe will not be displayed as a recommendation for **new deployments**
* The call to `_deployedApplicationQueryer.GetExistingDeployedApplications` now includes compatible Elastic Beanstalk environments
  * These environments are vetted to ensure that they were not created by our deploy tool and that they contain a dotnet compatible platform ARN.
* Created the `IDeploymentCommand` interface which serves as the blueprint for all current and future deployment commands.
* Created the `DeploymentCommandFactory`. It inspects the `DeploymentType` of the recipe and creates an instance of the appropriate deployment command
* Moved all the logic for a **CDK based deployment** to `CdkDeploymentCommand`. This class implements the `IDeploymentCommand`  interface.
   * In a follow up PR the logic to deploy to an existing Elastic Beanstalk environment will be supported by `BeanstalkEnvironmentDeploymentCommand`
* Added a new recipe to support deployments to existing Elastic Beanstalk environment. This recipe has its `DisableNewDeployments` property set to true

### DIFF
--- a/src/AWS.Deploy.Common/Recipes/RecipeDefinition.cs
+++ b/src/AWS.Deploy.Common/Recipes/RecipeDefinition.cs
@@ -29,6 +29,11 @@ namespace AWS.Deploy.Common.Recipes
         public string Name { get; set; }
 
         /// <summary>
+        /// Indicates if this recipe should be presented as an option during new deployments.
+        /// </summary>
+        public bool DisableNewDeployments { get; set; }
+
+        /// <summary>
         /// Description of the recipe informing the user what this recipe does and why it is recommended.
         /// </summary>
         public string Description { get; set; }
--- a/src/AWS.Deploy.Orchestration/DeploymentCommands/CdkDeploymentCommand.cs
+++ b/src/AWS.Deploy.Orchestration/DeploymentCommands/CdkDeploymentCommand.cs
@@ -1,0 +1,82 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+using AWS.Deploy.Common;
+using AWS.Deploy.Common.Recipes;
+using AWS.Deploy.Orchestration.DisplayedResources;
+
+namespace AWS.Deploy.Orchestration.DeploymentCommands
+{
+    public class CdkDeploymentCommand : IDeploymentCommand
+    {
+        public async Task ExecuteAsync(Orchestrator orchestrator, CloudApplication cloudApplication, Recommendation recommendation)
+        {
+            if (orchestrator._interactiveService == null)
+                throw new InvalidOperationException($"{nameof(orchestrator._interactiveService)} is null as part of the orchestartor object");
+            if (orchestrator._cdkManager == null)
+                throw new InvalidOperationException($"{nameof(orchestrator._cdkManager)} is null as part of the orchestartor object");
+            if (orchestrator._cdkProjectHandler == null)
+                throw new InvalidOperationException($"{nameof(CdkProjectHandler)} is null as part of the orchestartor object");
+            if (orchestrator._localUserSettingsEngine == null)
+                throw new InvalidOperationException($"{nameof(orchestrator._localUserSettingsEngine)} is null as part of the orchestartor object");
+            if (orchestrator._session == null)
+                throw new InvalidOperationException($"{nameof(orchestrator._session)} is null as part of the orchestartor object");
+            if (orchestrator._cdkVersionDetector == null)
+                throw new InvalidOperationException($"{nameof(orchestrator._cdkVersionDetector)} must not be null.");
+            if (orchestrator._directoryManager == null)
+                throw new InvalidOperationException($"{nameof(orchestrator._directoryManager)} must not be null.");
+
+            orchestrator._interactiveService.LogMessageLine(string.Empty);
+            orchestrator._interactiveService.LogMessageLine($"Initiating deployment: {recommendation.Name}");
+
+            orchestrator._interactiveService.LogMessageLine("Configuring AWS Cloud Development Kit (CDK)...");
+            var cdkProject = await orchestrator._cdkProjectHandler.ConfigureCdkProject(orchestrator._session, cloudApplication, recommendation);
+
+            var projFiles = orchestrator._directoryManager.GetProjFiles(cdkProject);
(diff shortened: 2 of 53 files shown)


src/AWS.Deploy.Common/Recipes/RecipeDefinition.cs
```diff
@@ -28,6 +28,11 @@ public class RecipeDefinition
         /// </summary>
         public string Name { get; set; }
 
+        /// <summary>
+        /// Indicates if this recipe should be presented as an option during new deployments.
+        /// </summary>
+        public bool Hidden { get; set; }
```
normj, line 34: I know this is taken from my doc but I want if we should just be more explicit in the name instead of relying on comments to understand what it is for. Something like `DisableNewDeployments`.

src/AWS.Deploy.Orchestration/DeploymentCommands/CdkDeploymentCommand.cs
```diff
@@ -0,0 +1,84 @@
...
+using System.Threading.Tasks;
+using AWS.Deploy.Common;
+using AWS.Deploy.Common.Recipes;
+using AWS.Deploy.Orchestration.DisplayedResources;
+
+namespace AWS.Deploy.Orchestration.DeploymentCommands
+{
+    public class CdkDeploymentCommand : IDeploymentCommand
+    {
+        public DeploymentStatus Status => throw new NotImplementedException();
```
normj, line 17: I assume this will eventually be implemented.

src/AWS.Deploy.Orchestration/DisplayedResources/DisplayedResourcesHandler.cs
```diff
@@ -16,8 +17,8 @@ public interface IDisplayedResourcesHandler
 
     public class DisplayedResourcesHandler : IDisplayedResourcesHandler
     {
-        private readonly IAWSResourceQueryer _awsResourceQueryer;
-        private readonly IDisplayedResourceCommandFactory _displayedResourcesFactory;
+        internal readonly IAWSResourceQueryer _awsResourceQueryer;
```
normj, line 20: Why are this fields now `internal`? Seems like we are leaking implementation details.

src/AWS.Deploy.Orchestration/Utilities/DeployedApplicationQueryer.cs
```diff
@@ -50,6 +51,18 @@ public class DeployedApplicationQueryer : IDeployedApplicationQueryer
         }
 
         public async Task<List<CloudApplication>> GetExistingDeployedApplications()
+        {
+            var existingApplications = new List<CloudApplication>();
+            existingApplications.AddRange(await GetExistingCloudFormationStacks());
+            existingApplications.AddRange(await GetExistingBeanstalkEnvironments());
```
normj, line 57: We should only be getting beanstalk environments if we are deploying a web application.

normj (changes requested):

src/AWS.Deploy.Common/Recipes/RecipeDefinition.cs
96malhar: Fixed

src/AWS.Deploy.Orchestration/Utilities/DeployedApplicationQueryer.cs
96malhar: If the client's application is not a web app, then the new _"Hidden"_ recipe will not be part of the recommendation list. Due to this all existing Beanstalk environments will fail the [IsCompatible check](https://github.com/aws/aws-dotnet-deploy/blob/main/src/AWS.Deploy.Orchestration/Utilities/DeployedApplicationQueryer.cs#L151) and will not be available for selection.

src/AWS.Deploy.Orchestration/DeploymentCommands/CdkDeploymentCommand.cs
96malhar: Synced offline and concluded that this field is not required. Hence removed it.

src/AWS.Deploy.Orchestration/DisplayedResources/DisplayedResourcesHandler.cs
96malhar: Converted these fields to `readonly` properties and included them in the `IDisplayedResourcesHandler` interface.

src/AWS.Deploy.Orchestration/Utilities/DeployedApplicationQueryer.cs
96malhar: Added an overload for `GetExistingDeployedApplications` that accepts a list of DeploymentTypes and acts as a filter for which resources to query.

normj (changes requested):

src/AWS.Deploy.Orchestration/Utilities/DeployedApplicationQueryer.cs
96malhar: This version was being used by the `ListDeployments` command. I agree that this can be misused. Got rid of this and updated the `ListDeployments` command to use the new version.

philasmar (approved):

PR #432, opened 2022-01-20
fix: KeyValue types not working with server mode

philasmar:
*Issue #, if available:*
IDE-6536

*Description of changes:*
Trying to set ElasticBeanstalkEnvironmentVariables which is of type KeyValue via server mode fails. The way we are we are handling the KeyValue type set via config files is not working with server mode. removing the code specific to config files works for both features.
In addition, when setting the KeyValue type, the values are not sticking because they are being handle as string instead of Dictionary. Additional handling was added to cover this case.

### DIFF
--- a/src/AWS.Deploy.Common/Recommendation.cs
+++ b/src/AWS.Deploy.Common/Recommendation.cs
@@ -91,7 +91,6 @@ namespace AWS.Deploy.Common
         /// Throws exception if there is no <see cref="OptionSettingItem" /> that matches <paramref name="jsonPath"/> />
         /// In case an option setting of type <see cref="OptionSettingValueType.KeyValue"/> is encountered,
         /// that <paramref name="jsonPath"/> can have the key value pair name as the leaf node with the option setting Id as the node before that.
-        /// In case there are multiple nodes after a <see cref="OptionSettingValueType.KeyValue"/>, then that indicates an invalid <paramref name="jsonPath"/>
         /// </summary>
         /// <param name="jsonPath">
         /// Dot (.) separated key values string pointing to an option setting.
@@ -118,11 +117,7 @@ namespace AWS.Deploy.Common
                 }
                 if (optionSetting.Type.Equals(OptionSettingValueType.KeyValue))
                 {
-                    if (i + 2 == ids.Length)
-                        return optionSetting;
-                    else
-                        throw new OptionSettingItemDoesNotExistException(DeployToolErrorCode.OptionSettingItemDoesNotExistInRecipe, $"The Option Setting Item {jsonPath} does not exist as part of the" +
-                    $" {Recipe.Name} recipe");
+                    return optionSetting;
                 }
             }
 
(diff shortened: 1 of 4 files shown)


src/AWS.Deploy.Common/Recommendation.cs
```diff
@@ -118,11 +117,7 @@ public OptionSettingItem GetOptionSetting(string? jsonPath)
                 }
                 if (optionSetting.Type.Equals(OptionSettingValueType.KeyValue))
                 {
-                    if (i + 2 == ids.Length)
-                        return optionSetting;
-                    else
-                        throw new OptionSettingItemDoesNotExistException(DeployToolErrorCode.OptionSettingItemDoesNotExistInRecipe, $"The Option Setting Item {jsonPath} does not exist as part of the" +
-                    $" {Recipe.Name} recipe");
+                    return optionSetting;
```
awschristou, line 120: Both changes appear to be unit testable. Is there something I'm overlooking?
It would be good to know that this function is working as intended, and doesn't introduce a regression when updated in the future.

philasmar: I added some unit tests

test/AWS.Deploy.CLI.UnitTests/SetOptionSettingTests.cs
```diff
@@ -65,7 +65,31 @@ public void SetOptionSettingTests_AllowedValues()
...
+        }
+
+        [Fact]
+        public void SetOptionSettingTests_KeyValueType()
+        {
+            var recommendation = _recommendations.First(r => r.Recipe.Id == Constants.ASPNET_CORE_BEANSTALK_RECIPE_ID);
+
+            var optionSetting = recommendation.Recipe.OptionSettings.First(x => x.Id.Equals("ElasticBeanstalkEnvironmentVariables"));
+            var values = new Dictionary<string, string>() { { "key", "value" } };
+            optionSetting.SetValueOverride(values);
```
awschristou, line 81: Does this exercise the change in src/AWS.Deploy.Common/Recommendation.cs on line 120, or does SetValueOverride need to be called with a valid json string in order to fire that code? I have limited understanding of this repo -- I'm guessing passing in valid json would simulate how the Toolkit is using this cli.

philasmar: It does exercise the change. The ID we are passing in is of KeyValue type which will trigger the section in SetValueOverride

awschristou: it looks like this code (above the change of interest) would be exercised (due to `valueOverride is Dictionary<string, string>`) :
```
            if (valueOverride is bool || valueOverride is int || valueOverride is long || valueOverride is double || valueOverride is Dictionary<string, string>)
            {
                _valueOverride = valueOverride;
            }
```
What am I missing?

philasmar: when it is being sent by server mode, it is not being recognized as a Dictionary. It is being detected as a string.

awschristou: That brings us to my initial question: can we make a test that calls SetValueOverride with a string (where the string is equivalent to the dictionary), so that we simulate the server mode case? Or is there something about the system under test that is being created in console mode?

philasmar: I added a test case that passes in a serialized dictionary.

awschristou (approved):

96malhar (approved):

normj (approved):

PR #433, opened 2022-01-21
feat: Improve telemetry by capturing CDK deployment failure message

philasmar:
*Issue #, if available:*
DOTNET-5575

*Description of changes:*
Based on the metrics gathered around the deploy tool usage, customers are experiencing deployment failures when they get to the `cdk deploy` stage. This part aims to break down the CDK deploy failures to have a better understanding of the customer's issues with the deploy tool.
This PR adds the ability to query CloudFormation to get the events of the deployment for a given stack. After we finish deploying to cdk (whether success or failure), we check the CloudFormation events for `CREATE_FAILED` or `UPDATE_FAILED` events. If they exist, we throw an exception with the error message of those events.
There is another part to this story where `cdk deploy` fails before creating a stack where we have to scrape the output text to understand the failures. That will be tackled as part of another PR.

### DIFF
--- a/src/AWS.Deploy.Orchestration/CdkProjectHandler.cs
+++ b/src/AWS.Deploy.Orchestration/CdkProjectHandler.cs
@@ -2,10 +2,13 @@ using System;
 using System.Collections;
 using System.Collections.Generic;
 using System.IO;
+using System.Linq;
 using System.Threading.Tasks;
+using Amazon.CloudFormation;
 using AWS.Deploy.Common;
 using AWS.Deploy.Common.IO;
 using AWS.Deploy.Orchestration.CDK;
+using AWS.Deploy.Orchestration.Data;
 using AWS.Deploy.Orchestration.Utilities;
 using AWS.Deploy.Recipes.CDK.Common;
 
@@ -15,7 +18,7 @@ namespace AWS.Deploy.Orchestration
     {
         Task<string> ConfigureCdkProject(OrchestratorSession session, CloudApplication cloudApplication, Recommendation recommendation);
         string CreateCdkProject(Recommendation recommendation, OrchestratorSession session, string? saveDirectoryPath = null);
-        Task DeployCdkProject(OrchestratorSession session, string cdkProjectPath, Recommendation recommendation);
+        Task DeployCdkProject(OrchestratorSession session, CloudApplication cloudApplication, string cdkProjectPath, Recommendation recommendation);
         void DeleteTemporaryCdkProject(OrchestratorSession session, string cdkProjectPath);
     }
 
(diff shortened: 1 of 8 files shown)


src/AWS.Deploy.Orchestration/CdkProjectHandler.cs
```diff
@@ -86,10 +94,35 @@ public async Task DeployCdkProject(OrchestratorSession session, string cdkProjec
...
                 throw new FailedToDeployCDKAppException(DeployToolErrorCode.FailedToDeployCdkApplication, "We had an issue deploying your application to AWS. Check the deployment output for more details.");
         }
 
+        private async Task CheckCdkDeploymentFailure(CloudApplication cloudApplication)
+        {
+            try
+            {
+                var stackEvents = await _awsResourceQueryer.GetCloudFormationStackEvents(cloudApplication.StackName);
+
+                var createdFailedEvent = stackEvents
```
normj, line 109: Variable name should be renamed since this is not just createFailedEvents.

src/AWS.Deploy.Orchestration/CdkProjectHandler.cs
```diff
@@ -86,10 +94,35 @@ public async Task DeployCdkProject(OrchestratorSession session, string cdkProjec
...
+
             if (cdkDeploy.ExitCode != 0)
                 throw new FailedToDeployCDKAppException(DeployToolErrorCode.FailedToDeployCdkApplication, "We had an issue deploying your application to AWS. Check the deployment output for more details.");
         }
 
+        private async Task CheckCdkDeploymentFailure(CloudApplication cloudApplication)
+        {
+            try
+            {
+                var stackEvents = await _awsResourceQueryer.GetCloudFormationStackEvents(cloudApplication.StackName);
```
normj, line 107: This looks like it is getting all events not just events since the start of the current deployment. So if the previous redeployment failed and the attempt to fix the issue on the next deployment this will still get the previous failed deployment update failed events.

normj (changes requested):

src/AWS.Deploy.Orchestration/CdkProjectHandler.cs
philasmar: Fixed

philasmar: Fixed

src/AWS.Deploy.Orchestration/CdkProjectHandler.cs
```diff
@@ -86,10 +94,56 @@ public async Task DeployCdkProject(OrchestratorSession session, string cdkProjec
                 redirectIO: true,
                 streamOutputToInteractiveService: true);
 
+            await CheckCdkDeploymentFailure(cloudApplication);
```
96malhar, line 97: Currently you are fetching **all** events associated with the CF stack We need to find the DateTime of the last event that occurred before the [cdkDeploy command](https://github.com/aws/aws-dotnet-deploy/pull/433/files#diff-1f36ff037bca7e527df601920aa6f58376c0e7aceb08c49238d40e9f8e2836beR90) is invoked and then pass it to the `CheckCdkDeploymentFailure` method. Inside this method, we only need to inspect events that occurred after the `lastEventTime`. If the CF stack is brand new, then set the `lastEventTime` to `DateTime.Now`

96malhar (changes requested):

src/AWS.Deploy.Orchestration/CdkProjectHandler.cs
philasmar: Fixed

96malhar (approved):

normj (approved):

PR #435, opened 2022-01-27
fix: beanstalk error at startup when in non-default regions

philasmar:
*Issue #, if available:*
DOTNET-5654

*Description of changes:*
* When recommendations are instantiated, all the recipe default values are check to create an inventory of the replacements needed.
* After a user selects the recommendation, a call is made to the orchestrator to do the token replacements if the recommendation has the replacement token in its inventory.
* When a user selects a beanstalk recommendation, if they are in a region that does not support beanstalk, the .net stack platform call will return empty which will throw an expected deploy tool exception instead of a generic one. The error message has been improved.
* In server mode, exceptions thrown in APIs now correctly return the error messages and error codes to the caller instead of having an empty body.

### DIFF
--- a/src/AWS.Deploy.Common/Recommendation.cs
+++ b/src/AWS.Deploy.Common/Recommendation.cs
@@ -4,6 +4,7 @@
 using System.Collections.Generic;
 using System.IO;
 using System.Linq;
+using System.Text.RegularExpressions;
 using AWS.Deploy.Common.Extensions;
 using AWS.Deploy.Common.Recipes;
 
@@ -31,7 +32,7 @@ namespace AWS.Deploy.Common
 
         public readonly List<OptionSettingItem> DeploymentBundleSettings = new ();
 
-        private readonly Dictionary<string, string> _replacementTokens = new();
+        public readonly Dictionary<string, string> ReplacementTokens = new();
 
         public Recommendation(RecipeDefinition recipe, ProjectDefinition projectDefinition, List<OptionSettingItem> deploymentBundleSettings, int computedPriority, Dictionary<string, string> additionalReplacements)
         {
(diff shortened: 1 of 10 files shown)


src/AWS.Deploy.Orchestration/Orchestrator.cs
```diff
@@ -110,21 +107,24 @@ public async Task<List<Recommendation>> GenerateRecommendationsFromSavedDeployme
                 throw new InvalidCliArgumentException(DeployToolErrorCode.DeploymentProjectPathNotFound, $"The path '{deploymentProjectPath}' does not exists on the file system. Please provide a valid deployment project path and try again.");
 
             var engine = new RecommendationEngine.RecommendationEngine(new List<string> { deploymentProjectPath }, _session);
-            var additionalReplacements = await GetReplacements();
-            return await engine.ComputeRecommendations(additionalReplacements);
+            return await engine.ComputeRecommendations();
         }
 
-        public async Task<Dictionary<string, string>> GetReplacements()
+        public async Task AddReplacementTokens(Recommendation recommendation, string cloudApplicationName)
```
normj, line 113: Rename method to something more appropriate.

src/AWS.Deploy.Common/Recommendation.cs
```diff
@@ -44,10 +45,30 @@ public Recommendation(RecipeDefinition recipe, ProjectDefinition projectDefiniti
...
+
             foreach (var replacement in additionalReplacements)
             {
-                if (!_replacementTokens.ContainsKey(replacement.Key))
-                    _replacementTokens[replacement.Key] = replacement.Value;
+                ReplacementTokens[replacement.Key] = replacement.Value;
+            }
+        }
+
+        private void GetRecommendationReplacementTokens(List<OptionSettingItem> optionSettings)
```
normj, line 56: Maybe rename to Collect or something more appropriate.

normj (changes requested):

src/AWS.Deploy.Orchestration/Orchestrator.cs
philasmar: Fixed

src/AWS.Deploy.Common/Recommendation.cs
philasmar: Fixed

normj (approved):

96malhar (approved):

PR #437, opened 2022-01-31
feat: add check for opt-in regions

philasmar:
*Issue #, if available:*
DOTNET-5640

*Description of changes:*
Update existing GetCallerIdentity call at Command startup which checks if a region is reachable. If it is not, we retrieve the region partition. If it is in the commercial partition, we do another GetCallerIdentity call to us-east-1(default region). If that succeeds, we conclude that the user is in an opt-in region and needs to enable it.

image

### DIFF
--- a/src/AWS.Deploy.Constants/CLI.cs
+++ b/src/AWS.Deploy.Constants/CLI.cs
@@ -6,6 +6,10 @@ namespace AWS.Deploy.Constants
 {
     internal static class CLI
     {
+        // Represents the default STS AWS region that is used for the purposes of
+        // retrieving the caller identity and determining if a user is in an opt-in region.
+        public const string DEFAULT_STS_AWS_REGION = "us-east-1";
+
         public const string CREATE_NEW_LABEL = "*** Create new ***";
         public const string DEFAULT_LABEL = "*** Default ***";
         public const string EMPTY_LABEL = "*** Empty ***";
--- a/src/AWS.Deploy.Orchestration/Data/AWSResourceQueryer.cs
+++ b/src/AWS.Deploy.Orchestration/Data/AWSResourceQueryer.cs
@@ -6,6 +6,7 @@ using System.Collections.Generic;
 using System.IO;
 using System.Linq;
 using System.Threading.Tasks;
+using Amazon;
 using Amazon.CloudFormation;
 using Amazon.CloudFormation.Model;
 using Amazon.CloudFront;
@@ -63,7 +64,7 @@ namespace AWS.Deploy.Orchestration.Data
         Task<List<Repository>> GetECRRepositories(List<string> repositoryNames);
         Task<Repository> CreateECRRepository(string repositoryName);
         Task<List<Stack>> GetCloudFormationStacks();
-        Task<GetCallerIdentityResponse> GetCallerIdentity();
+        Task<GetCallerIdentityResponse> GetCallerIdentity(string awsRegion);
         Task<List<Amazon.ElasticLoadBalancingV2.Model.LoadBalancer>> ListOfLoadBalancers(LoadBalancerTypeEnum loadBalancerType);
         Task<Distribution> GetCloudFrontDistribution(string distributionId);
         Task<List<string>> ListOfDyanmoDBTables();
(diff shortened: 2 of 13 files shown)


src/AWS.Deploy.Orchestration/Data/AWSResourceQueryer.cs
```diff
@@ -438,10 +440,56 @@ public async Task<List<Stack>> GetCloudFormationStacks()
...
+                    awsOption.Credentials = awsCredentials;
+                    awsOption.Region = RegionEndpoint.GetBySystemName(awsRegion);
+                });
+
+                using var stsClient = _awsClientFactory.GetAWSClient<IAmazonSecurityTokenService>();
+                return await stsClient.GetCallerIdentityAsync(request);
+            }
+            catch (Exception ex)
+            {
+                var regionEndpoint = RegionEndpoint.GetBySystemName(awsRegion);
```
awschristou, line 460: I believe this throws if awsRegion is not a value that the SDK knows about (for example, a newer region), and you're in a catch block. Is that handled as expected?

awschristou (commented):
Is it reasonable to add testing coverage to this change?

src/AWS.Deploy.Orchestration/Data/AWSResourceQueryer.cs
philasmar: I checked this and it does not throw an exception. It returns an endpoint with an "Unknown" name.

philasmar:
> Is it reasonable to add testing coverage to this change?

I added a unit test that asserts an error thrown if GetCallerIdentity fails.

src/AWS.Deploy.Orchestration/Data/AWSResourceQueryer.cs
```diff
@@ -438,10 +440,56 @@ public async Task<List<Stack>> GetCloudFormationStacks()
...
+                    }
+                    catch (Exception e)
+                    {
+                        throw new UnableToAccessAWSRegionException(
+                           DeployToolErrorCode.UnableToAccessAWSRegion,
+                           $"We were unable to access the AWS region '{awsRegion}'. Make sure you have correct permissions for that region and the region is accessible.",
+                           e);
+                    }
+
+                    throw new UnableToAccessAWSRegionException(
```
awschristou, line 482: Did you mean to return on line 472? This appears to be unconditionally thrown.
If that is the case, please add a test that verifies that line 472 returns as expected.

src/AWS.Deploy.Orchestration/Data/AWSResourceQueryer.cs
```diff
@@ -438,10 +440,56 @@ public async Task<List<Stack>> GetCloudFormationStacks()
...
-        public async Task<GetCallerIdentityResponse> GetCallerIdentity()
+        public async Task<GetCallerIdentityResponse> GetCallerIdentity(AWSCredentials awsCredentials, string awsRegion)
         {
-            using var stsClient = _awsClientFactory.GetAWSClient<IAmazonSecurityTokenService>();
-            return await stsClient.GetCallerIdentityAsync(new GetCallerIdentityRequest());
+            var request = new GetCallerIdentityRequest();
+
+            try
+            {
+                _awsClientFactory.ConfigureAWSOptions(awsOption =>
```
awschristou, line 449: This appears to mutate shared state. Can this lead to race conditions? Could this risk be mitigated by moving this lambda to an argument of _awsClientFactory.GetAWSClient ?

src/AWS.Deploy.Orchestration/Data/AWSResourceQueryer.cs
philasmar: No this is intentional. Any time you step into the catch block, and exception will be thrown. It's just a matter of which exception and error code needs to be thrown. I have added a unit test to cover that case.

src/AWS.Deploy.Orchestration/Data/AWSResourceQueryer.cs
philasmar: I have updated it to no longer need the lambda.

src/AWS.Deploy.Orchestration/Data/AWSResourceQueryer.cs
```diff
@@ -438,10 +439,44 @@ public async Task<List<Stack>> GetCloudFormationStacks()
...
+                    {
+                        throw new UnableToAccessAWSRegionException(
+                           DeployToolErrorCode.UnableToAccessAWSRegion,
+                           $"We were unable to access the AWS region '{awsRegion}'. Make sure you have correct permissions for that region and the region is accessible.",
+                           e);
+                    }
+
+                    throw new UnableToAccessAWSRegionException(
+                        DeployToolErrorCode.OptInRegionDisabled,
+                        $"We were unable to access the Opt-In region '{awsRegion}'. Please enable the AWS Region '{awsRegion}' and try again.",
```
awschristou, line 471: Consider adding a link to an appropriate AWS User Guide url, so that users understand how they can enable a region.

src/AWS.Deploy.Orchestration/Data/AWSResourceQueryer.cs
awschristou: (non blocking, at your discretion) Consider moving this whole catch block logic (line 451) to a separate method to help make it easier to understand the intent.

awschristou (commented):
This change looks reasonable to me. Deferring to active contributors on this project to approve.

src/AWS.Deploy.Orchestration/Data/AWSResourceQueryer.cs
philasmar: I added a link to provide more info.

src/AWS.Deploy.Orchestration/Data/AWSResourceQueryer.cs
```diff
@@ -438,10 +439,44 @@ public async Task<List<Stack>> GetCloudFormationStacks()
...
+                    {
+                        throw new UnableToAccessAWSRegionException(
+                           DeployToolErrorCode.UnableToAccessAWSRegion,
+                           $"We were unable to access the AWS region '{awsRegion}'. Make sure you have correct permissions for that region and the region is accessible.",
+                           e);
+                    }
+
+                    throw new UnableToAccessAWSRegionException(
+                        DeployToolErrorCode.OptInRegionDisabled,
+                        $"We were unable to access the Opt-In region '{awsRegion}'. Please enable the AWS Region '{awsRegion}' and try again. Additional details could be found at https://docs.aws.amazon.com/general/latest/gr/rande-manage.html",
```
awschristou, line 471: Can you check if there is a url with an english language code? ("gr" / "rande-manage")

src/AWS.Deploy.Constants/CLI.cs
```diff
@@ -6,6 +6,8 @@ namespace AWS.Deploy.Constants
 {
     internal static class CLI
     {
+        public const string DEFAULT_AWS_REGION = "us-east-1";
```
normj, line 9: I don't want accidental use of this constant. Can you rename to `DEFAULT_STS_AWS_REGION` with a comment that this should only be used in this use case.

src/AWS.Deploy.Orchestration/Data/AWSResourceQueryer.cs
normj: To be clear the display name will be `Unknown` and the system name will be what you pass in for `awsRegion`. It was the classes way of dealing with regions not defined in the `endpoints.json` file.

src/AWS.Deploy.Orchestration/Data/AWSResourceQueryer.cs
```diff
@@ -438,10 +439,44 @@ public async Task<List<Stack>> GetCloudFormationStacks()
...
+
+            try
+            {
+                using var stsClient = _awsClientFactory.GetAWSClient<IAmazonSecurityTokenService>(awsRegion);
+                return await stsClient.GetCallerIdentityAsync(request);
+            }
+            catch (Exception ex)
+            {
+                var regionEndpoint = RegionEndpoint.GetBySystemName(awsRegion);
+                if (regionEndpoint.PartitionName.Equals("aws") && !awsRegion.Equals(Constants.CLI.DEFAULT_AWS_REGION))
```
normj, line 454: `regionEndpoint.PartitionName` could be null if the region wasn't known.

normj (changes requested):

96malhar (approved):

src/AWS.Deploy.Constants/CLI.cs
philasmar: Fixed

normj (approved):

PR #438, opened 2022-02-01
feat: capture standard error in server mode session

philasmar:
*Issue #, if available:*
DOTNET-5204

*Description of changes:*
Redirected standard error in server mode session to capture any error thrown in startup.
In the case where the deploy tool is not installed, server mode session throws an error but does not return the message indicating that the deploy tool is not installed. This change makes it so that errors are returned.

### DIFF
--- a/src/AWS.Deploy.ServerMode.Client/CommandLineWrapper.cs
+++ b/src/AWS.Deploy.ServerMode.Client/CommandLineWrapper.cs
@@ -6,7 +6,9 @@ using System.Collections;
 using System.Collections.Generic;
 using System.Diagnostics;
 using System.Runtime.InteropServices;
+using System.Text;
 using System.Threading.Tasks;
+using AWS.Deploy.ServerMode.Client.Utilities;
 
 namespace AWS.Deploy.ServerMode.Client
 {
@@ -19,9 +21,12 @@ namespace AWS.Deploy.ServerMode.Client
             _diagnosticLoggingEnabled = diagnosticLoggingEnabled;
         }
 
-        public virtual async Task<int> Run(string command, params string[] stdIn)
+        public virtual async Task<RunResult> Run(string command, params string[] stdIn)
         {
             var arguments = RuntimeInformation.IsOSPlatform(OSPlatform.Windows) ? $"/c {command}" : $"-c \"{command}\"";
+
+            var strOutput = new CappedStringBuilder(100);
+            var strError = new CappedStringBuilder(50);
 
             var processStartInfo = new ProcessStartInfo
             {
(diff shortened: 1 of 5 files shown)


src/AWS.Deploy.ServerMode.Client/CommandLineWrapper.cs
```diff
@@ -43,9 +49,28 @@ public virtual async Task<int> Run(string command, params string[] stdIn)
...
+            process.BeginOutputReadLine();
+            process.BeginErrorReadLine();
+
             process.WaitForExit(-1);
 
-            return await Task.FromResult(process.ExitCode).ConfigureAwait(false);
+            var result = new RunResult
+            {
+                ExitCode = process.ExitCode,
+                StandardError = strError.ToString(),
```
normj, line 69: If the deploy tool is running for a long time and if we aren't careful what we write to stdout or stderr this string builders could grow very long. We should probably cap what we store so we don't have an indefinitely growing string StringBuilder behind the scenes. Also we should probably only return back to VS the last few lines.

normj (changes requested):

src/AWS.Deploy.ServerMode.Client/CommandLineWrapper.cs
philasmar: I created a CappedStringBuilder which essentially limits the number of lines the StringBuilder can hold.

src/AWS.Deploy.ServerMode.Client/Utilities/CappedStringBuilder.cs
```diff
@@ -0,0 +1,57 @@
...
+// SPDX-License-Identifier: Apache-2.0
+
+using System;
+using System.Collections.Generic;
+using System.Text;
+using System.Linq;
+
+namespace AWS.Deploy.ServerMode.Client.Utilities
+{
+    public class CappedStringBuilder
```
normj, line 11: Add some unit tests

src/AWS.Deploy.ServerMode.Client/Utilities/CappedStringBuilder.cs
```diff
@@ -0,0 +1,57 @@
...
+    {
+        public int LineLimit { get; set; }
+        public int LineLength { get; private set; } = 0;
+
+        private readonly StringBuilder _stringBuilder;
```
normj, line 16: I would get rid of the string builder and have a queue of lines of strings that you concat together whenever the ToString method is called.

normj (changes requested):
A bit of bike shedding on your new `CappedStringBuilder` class but we definitely should have unit tests on the class.

src/AWS.Deploy.ServerMode.Client/Utilities/CappedStringBuilder.cs
philasmar: Done

src/AWS.Deploy.ServerMode.Client/Utilities/CappedStringBuilder.cs
philasmar: Done

normj (approved):

PR #462, opened 2022-02-23
feat: Improve the CLI UX to support multiple deployment targets

96malhar:
*Issue #, if available:*
DOTNET-5711

*Description of changes:*
This PR improves the CLI UX to support multiple deployment targets. It adds the ability to generate the input prompts for `CloudApplication` name by inspecting the deployment type of the recipe at run time.

**This PR establishes the following UX** _(Switch to a non-dark mode on the browser to view the flowchart clearly)_

deploy-tool-ux

### DIFF
--- a/src/AWS.Deploy.CLI/Commands/DeployCommand.cs
+++ b/src/AWS.Deploy.CLI/Commands/DeployCommand.cs
@@ -145,6 +145,8 @@ namespace AWS.Deploy.CLI.Commands
         /// <returns>A tuple consisting of the Orchestrator object, Selected Recommendation, Cloud Application metadata.</returns>
         public async Task<(Orchestrator, Recommendation, CloudApplication)> InitializeDeployment(string applicationName, UserDeploymentSettings? userDeploymentSettings, string deploymentProjectPath)
         {
+            string cloudApplicationName;
+
             var orchestrator = new Orchestrator(
                     _session,
                     _orchestratorInteractiveService,
@@ -170,8 +172,12 @@ namespace AWS.Deploy.CLI.Commands
             // Filter compatible applications that can be re-deployed  using the current set of recommendations.
             var compatibleApplications = await _deployedApplicationQueryer.GetCompatibleApplications(recommendations, allDeployedApplications, _session);
 
-            // Get CloudApplication name.
-            var cloudApplicationName = GetCloudApplicationName(applicationName, userDeploymentSettings, compatibleApplications);
+            // Try finding the CloudApplication name via the --application-name CLI argument or user provided config settings.
+            cloudApplicationName = GetCloudApplicationNameFromDeploymentSettings(applicationName, userDeploymentSettings);
+
+            // Prompt the user with a choice to re-deploy to existing targets or deploy to a new cloud application.
+            if (string.IsNullOrEmpty(cloudApplicationName))
+                cloudApplicationName = AskForCloudApplicationNameFromDeployedApplications(compatibleApplications);
 
             // Find existing application with the same CloudApplication name.
             var deployedApplication = allDeployedApplications.FirstOrDefault(x => string.Equals(x.Name, cloudApplicationName));
@@ -200,6 +206,10 @@ namespace AWS.Deploy.CLI.Commands
                     // Filter the recommendation list for a NEW deployment with recipes which have the DisableNewDeployments property set to false.
                     selectedRecommendation = GetSelectedRecommendation(userDeploymentSettings, recommendations.Where(x => !x.Recipe.DisableNewDeployments).ToList());
                 }
+
+                // Ask the user for a new Cloud Application name based on the deployment type of the recipe.
+                if (string.IsNullOrEmpty(cloudApplicationName))
+                    cloudApplicationName = AskForNewCloudApplicationName(selectedRecommendation.Recipe.DeploymentType, compatibleApplications);
             }
 
             await orchestrator.ApplyAllReplacementTokens(selectedRecommendation, cloudApplicationName);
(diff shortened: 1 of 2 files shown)


src/AWS.Deploy.CLI/Commands/DeployCommand.cs
```diff
@@ -481,22 +493,99 @@ private string GetCloudApplicationName(string? applicationName, UserDeploymentSe
...
+        private string AskForCloudApplicationNameFromDeployedApplications(List<CloudApplication> deployedApplications)
+        {
+            if (!deployedApplications.Any())
+                return string.Empty;
+
+            var title = "Select an existing AWS deployment target to deploy your application to.";
+
+            var userResponse =
+                        _consoleUtilities.AskUserToChooseOrCreateNew(
+                            deployedApplications.Select(x => x.DisplayName),
```
philasmar, line 520: Instead of using the DisplayName for both the Key and Value of the AskUserToChooseOrCreateNew method, we could use the Name as the Key and DisplayName as the Value. That way we wouldn't need to parse the DisplayName and split it on line 531.

src/AWS.Deploy.CLI/Commands/DeployCommand.cs
```diff
@@ -481,22 +493,99 @@ private string GetCloudApplicationName(string? applicationName, UserDeploymentSe
...
+                            askNewName: false,
+                            defaultChoosePrompt: Constants.CLI.PROMPT_CHOOSE_DEPLOYMENT_TARGET,
+                            defaultCreateNewLabel: Constants.CLI.CREATE_NEW_APPLICATION_LABEL);
+
+            if (userResponse.CreateNew)
+                return string.Empty;
+
+            // Since the selected option will be the display name, we need to extract the actual name out of it.
+            // Ex - DisplayName = "MyAppStack (CloudFormationStack)", ActualName = "MyAppStack"
+            var cloudApplicationName = userResponse.SelectedOption?.Split().FirstOrDefault() ?? string.Empty;
```
philasmar, line 531: I don't like that we are doing string manipulation here. We should use the CloudApplication Name.

src/AWS.Deploy.CLI/Commands/DeployCommand.cs
```diff
@@ -481,22 +493,99 @@ private string GetCloudApplicationName(string? applicationName, UserDeploymentSe
...
+
+                string inputPrompt;
+
+                switch (deploymentType)
+                {
+                    case DeploymentTypes.CdkProject:
+                        inputPrompt = Constants.CLI.PROMPT_NEW_STACK_NAME;
+                        break;
+                    default:
+                        throw new InvalidOperationException($"The {nameof(DeploymentTypes)} {deploymentType} does not have an input prompt");
```
philasmar, line 574: I'd refrain from using non-deploy tool exceptions which will lead to an ugly stack trace returned to the user if thrown.

philasmar (changes requested):

src/AWS.Deploy.CLI/Commands/DeployCommand.cs
96malhar: The `AskUserToChooseOrCreateNew` method does not take a separate key and value. Am I missing something? Also this piece of logic was already there. Its not something new that I am introducing as part of this PR.

src/AWS.Deploy.CLI/Commands/DeployCommand.cs
96malhar: The AskUserToChooseOrCreateNew method does not take a separate key and value. Am I missing something? Also this piece of logic was already there. Its not something new that I am introducing as part of this PR.

src/AWS.Deploy.CLI/Commands/DeployCommand.cs
96malhar: The deploy-tool exceptions are presented only in case of a user initiated error. They are used to display errors which can be fixed by the user directly. In this case, the error will not occur due to a user action. Its an implementation bug and we want to fail with a full stack trace.

src/AWS.Deploy.CLI/Commands/DeployCommand.cs
philasmar: The is a `AskUserToChooseOrCreateNew` method that take in a `UserInputConfiguration` where you can specify a DisplaySelector to do what you are trying to achieve. I agree that you are not introducing anything new, but since you are modifying this code. If we can do better, we should.

src/AWS.Deploy.CLI/Commands/DeployCommand.cs
philasmar: The is a `AskUserToChooseOrCreateNew` method that take in a `UserInputConfiguration` where you can specify a DisplaySelector to do what you are trying to achieve. I agree that you are not introducing anything new, but since you are modifying this code. If we can do better, we should.

src/AWS.Deploy.CLI/Commands/DeployCommand.cs
philasmar: Sounds good

src/AWS.Deploy.CLI/Commands/DeployCommand.cs
96malhar: Changed this to use the `UserInputConfiguration`

src/AWS.Deploy.CLI/Commands/DeployCommand.cs
96malhar: Changes this to use the `UserInputConfiguration`

philasmar (approved):

src/AWS.Deploy.CLI/Commands/DeployCommand.cs
```diff
@@ -145,6 +145,8 @@ private void DisplayOutputResources(List<DisplayedResourceItem> displayedResourc
         /// <returns>A tuple consisting of the Orchestrator object, Selected Recommendation, Cloud Application metadata.</returns>
         public async Task<(Orchestrator, Recommendation, CloudApplication)> InitializeDeployment(string applicationName, UserDeploymentSettings? userDeploymentSettings, string deploymentProjectPath)
         {
+            string cloudApplicationName;
```
normj, line 148: nit - You don't need to declare this variable till you use the variable on line 176.

normj (approved):

PR #469, opened 2022-03-08
feat: Add support for worker project types

normj:
*Description of changes:*
Add support for Worker project types being deploy as a ECS Fargate service. To distinguish a worker project type the csproj `Sdk` attribute is set to `Microsoft.NET.Sdk.Worker`. 

The new sdk typed caused a challenge with our existing rules for the recipe because we look for `Microsoft.NET.Sdk` and if the `OutputType` property is set to `Exe`. With `Microsoft.NET.Sdk.Worker` there is no `OutputType` property because it is inherited from the SDK. The rules got rewritten to have the initial priority be negative meaning exclude and then only include the recipe if either the rule for `Microsoft.NET.Sdk` and `Exe` was true or if `Microsoft.NET.Sdk.Worker` was true by setting the priority to a positive number.

### DIFF
--- a/src/AWS.Deploy.Orchestration/RecommendationEngine/MSProjectSdkAttributeTest.cs
+++ b/src/AWS.Deploy.Orchestration/RecommendationEngine/MSProjectSdkAttributeTest.cs
@@ -15,7 +15,16 @@ namespace AWS.Deploy.Orchestration.RecommendationEngine
 
         public override Task<bool> Execute(RecommendationTestInput input)
         {
-            var result = string.Equals(input.ProjectDefinition.SdkType, input.Test.Condition.Value, StringComparison.InvariantCultureIgnoreCase);
+            bool result = false;
+            if(!string.IsNullOrEmpty(input.Test.Condition.Value))
+            {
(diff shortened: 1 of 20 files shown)


src/AWS.Deploy.Orchestration/RecommendationEngine/MSProjectSdkAttributeTest.cs
```diff
@@ -15,7 +15,16 @@ public class MSProjectSdkAttributeTest : BaseRecommendationTest
 
         public override Task<bool> Execute(RecommendationTestInput input)
         {
-            var result = string.Equals(input.ProjectDefinition.SdkType, input.Test.Condition.Value, StringComparison.InvariantCultureIgnoreCase);
+            bool result = false;
+            if(!string.IsNullOrEmpty(input.Test.Condition.Value))
```
normj, line 19: Initially I was going to add support for working by just updating the SdkType test to be able to use the AllowedValues. That ended up not working because I had the issue with the `OutputType` not being set for worker. I left this in here thinking it could be useful for future recipes.

philasmar (approved):

96malhar (approved):

PR #472, opened 2022-03-14
feat: Add List type and VPC Connector setting to App Runner

philasmar:
*Issue #, if available:*
DOTNET-5710

*Description of changes:*
* Add a List type as a supported option for Option Setting Items
* Added VPC Configuration Option Setting Item to App runner recipe, which allows you to specify a list of subnets and security groups that will allow App Runner to determine the proper VPC to use.

https://user-images.githubusercontent.com/53088140/158249532-0d9aa166-540b-463b-aad8-7aa39b36a392.mov

### DIFF
--- a/src/AWS.Deploy.CLI/ConsoleUtilities.cs
+++ b/src/AWS.Deploy.CLI/ConsoleUtilities.cs
@@ -32,6 +32,7 @@ namespace AWS.Deploy.CLI
         YesNo AskYesNoQuestion(string question, YesNo? defaultValue = default);
         void DisplayValues(Dictionary<string, object> objectValues, string indent);
         Dictionary<string, string> AskUserForKeyValue(Dictionary<string, string> keyValue);
+        SortedSet<string> AskUserForList(SortedSet<string> listValue);
     }
 
     public class ConsoleUtilities : IConsoleUtilities
@@ -307,6 +308,59 @@ namespace AWS.Deploy.CLI
             return userValue;
         }
 
+        public SortedSet<string> AskUserForList(SortedSet<string> listValues)
+        {
+            listValues ??= new SortedSet<string>();
+
+            if (listValues.Count == 0)
+            {
+                AskToAddListItem(listValues);
+                return listValues;
+            }
+
+            const string ADD = "Add new";
+            const string UPDATE = "Update existing";
+            const string DELETE = "Delete existing";
+            const string NOOP = "No action";
+            var operations = new List<string> { ADD, UPDATE, DELETE, NOOP };
+
+            var selectedOperation = AskUserToChoose(operations, "Select which operation you want to perform:", NOOP);
+
+            if (selectedOperation.Equals(ADD))
+                AskToAddListItem(listValues);
+            else if (selectedOperation.Equals(UPDATE))
+                AskToUpdateListItem(listValues);
+            else if (selectedOperation.Equals(DELETE))
+                AskToDeleteListItem(listValues);
+
+            return listValues;
+        }
+
+        private void AskToAddListItem(SortedSet<string> listValues)
+        {
+            _interactiveService.WriteLine("Enter a value:");
+            var listValue = _interactiveService.ReadLine()?.Trim() ?? "";
+            if (!string.IsNullOrEmpty(listValue))
+                listValues.Add(listValue);
+        }
+
+        private void AskToUpdateListItem(SortedSet<string> listValues)
+        {
+            var selectedItem = AskUserToChoose(listValues.ToList(), "Select the value you wish to update:", null);
+            var selectedValue = AskUserForValue("Enter the updated value:", selectedItem, true);
+            if (!string.IsNullOrEmpty(selectedValue))
+            {
+                listValues.Remove(selectedItem);
+                listValues.Add(selectedValue);
+            }
+        }
+
+        private void AskToDeleteListItem(SortedSet<string> listValues)
+        {
+            var selectedItem = AskUserToChoose(listValues.ToList(), "Select the value you wish to delete:", null);
+            listValues.Remove(selectedItem);
+        }
+
         public Dictionary<string, string> AskUserForKeyValue(Dictionary<string, string> keyValue)
         {
             keyValue ??= new Dictionary<string, string>();
@@ -468,6 +522,14 @@ namespace AWS.Deploy.CLI
                 {
                     _interactiveService.WriteLine($"{indent}{key}");
                     DisplayValues(childObjectValue, $"{indent}\t");
+                }
+                else if (value is SortedSet<string> listValues)
+                {
+                    _interactiveService.WriteLine($"{indent}{key}:");
+                    foreach (var listValue in listValues)
+                    {
+                        _interactiveService.WriteLine($"{indent}\t{listValue}");
+                    }
                 }
                 else if (value is string stringValue)
                 {
--- a/src/AWS.Deploy.Common/UserDeploymentSettings.cs
+++ b/src/AWS.Deploy.Common/UserDeploymentSettings.cs
@@ -4,6 +4,7 @@
 using System;
 using System.Collections.Generic;
 using System.IO;
+using System.Linq;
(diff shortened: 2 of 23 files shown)


src/AWS.Deploy.Recipes/RecipeDefinitions/ASP.NETAppAppRunner.recipe
```diff
@@ -363,6 +363,53 @@
...
+                    "AdvancedSetting": false,
+                    "Updatable": true
+                },
+                {
+                    "Id": "Subnets",
+                    "Name": "Subnets",
+                    "Description": "A list of IDs of subnets that App Runner should use when it associates your service with a custom Amazon VPC. Specify IDs of subnets of a single Amazon VPC. App Runner determines the Amazon VPC from the subnets you specify.",
+                    "Type": "List",
+                    "AdvancedSetting": false,
+                    "Updatable": true,
```
normj, line 390: Needs Validation RegEx

src/AWS.Deploy.Common/UserDeploymentSettings.cs
```diff
@@ -53,6 +54,13 @@ public class UserDeploymentSettings
         /// <param name="node">The current node that is being processed.</param>
         private void TraverseRootToLeaf(JToken node)
         {
+            if (!string.IsNullOrEmpty(node.Path) && node.Type.ToString().Equals("Array"))
```
normj, line 57: Can we have some unit tests that exercise this code path?

src/AWS.Deploy.Recipes/RecipeDefinitions/ASP.NETAppAppRunner.recipe
```diff
@@ -363,6 +363,53 @@
...
+                    "Type": "Bool",
+                    "DefaultValue": false,
+                    "AdvancedSetting": false,
+                    "Updatable": true
+                },
+                {
+                    "Id": "Subnets",
+                    "Name": "Subnets",
+                    "Description": "A list of IDs of subnets that App Runner should use when it associates your service with a custom Amazon VPC. Specify IDs of subnets of a single Amazon VPC. App Runner determines the Amazon VPC from the subnets you specify.",
+                    "Type": "List",
```
normj, line 388: Should there be a sub property like `ListType` that indicates the type of data for the list. In this case `string`?

normj (changes requested):
Mostly comments on the recipe changes. The list mechanisms seem fine. I do wonder if we should have a way of adding multiple items to the list without having to keep going back to the main screen. Most likely a user would at least 2 subnets for redundancy.

src/AWS.Deploy.Recipes/RecipeDefinitions/ASP.NETAppAppRunner.recipe
```diff
@@ -363,6 +363,73 @@
...
+                            "Id": "VPCConnector.CreateNew",
+                            "Value": true
+                        }
+                    ]
+                },
+                {
+                    "Id": "SecurityGroups",
+                    "Name": "SecurityGroups",
+                    "Description": "A list of IDs of security groups that App Runner should use for access to AWS resources under the specified subnets. If not specified, App Runner uses the default security group of the Amazon VPC. The default security group allows all outbound traffic.",
+                    "Type": "List",
```
ganeshnj, line 412: Needs update in schema that recipe now supports `List` in option setting Type

src/AWS.Deploy.CLI/ConsoleUtilities.cs
```diff
@@ -307,6 +308,60 @@ public string AskUserForValue(string message, string defaultValue, bool allowEmp
...
+            var selectedOperation = AskUserToChoose(operations, "Select which operation you want to perform:", ADD);
+
+            if (selectedOperation.Equals(ADD))
+                AskToAddListItem(listValues);
+
+            if (selectedOperation.Equals(UPDATE))
+                AskToUpdateListItem(listValues);
+
+            if (selectedOperation.Equals(DELETE))
+                AskToDeleteListItem(listValues);
```
ganeshnj, line 335:
```suggestion
            if (selectedOperation.Equals(ADD))
                AskToAddListItem(listValues);
            else if (selectedOperation.Equals(UPDATE))
                AskToUpdateListItem(listValues);
            else if (selectedOperation.Equals(DELETE))
                AskToDeleteListItem(listValues);
```

slightly better, there is no worth doing all the checks after first successful check.

ganeshnj (changes requested):

philasmar:
> Mostly comments on the recipe changes. The list mechanisms seem fine. I do wonder if we should have a way of adding multiple items to the list without having to keep going back to the main screen. Most likely a user would at least 2 subnets for redundancy.

I think for now let's leave this mechanism similar to what we have with environment variables (KeyValue pairs). If users want to add multiple, they could use a config file. If we receive feedback on the current UX, then we can update it in the future for both List and KeyValue (which use the same mechanism).

src/AWS.Deploy.Common/UserDeploymentSettings.cs
philasmar: Done

src/AWS.Deploy.Recipes/RecipeDefinitions/ASP.NETAppAppRunner.recipe
philasmar: Since there is no need for this at the moment, I would say let's stick with the assumption that by default a List type is a List<string>. In the future, if we add ListType to the recipe we could change the default from string to whichever type is specified. We wouldn't have any one way doors in this case.

src/AWS.Deploy.Recipes/RecipeDefinitions/ASP.NETAppAppRunner.recipe
philasmar: Done

src/AWS.Deploy.Recipes/RecipeDefinitions/ASP.NETAppAppRunner.recipe
```diff
@@ -363,6 +363,89 @@
...
+                },
+                {
+                    "Id": "VpcConnectorId",
+                    "Name": "Existing VPC Connector ID",
+                    "Description": "The ID of the existing VPC Connector to use.",
+                    "Type": "String",
+                    "TypeHint": "ExistingVpcConnector",
+                    "DefaultValue": null,
+                    "AdvancedSetting": false,
+                    "Updatable": false,
```
normj, line 392: Update = `true`

src/AWS.Deploy.CLI/ConsoleUtilities.cs
```diff
@@ -307,6 +308,58 @@ public string AskUserForValue(string message, string defaultValue, bool allowEmp
             return userValue;
         }
 
+        public SortedSet<string> AskUserForList(SortedSet<string> listValues)
```
normj, line 311: When going through the experience of adding a VPC connector I added a subnet and security group. I wanted to add a second subnet to my VPC connector which goes through all of the settings including both subnets and security groups. This was awkward because I didn't want to change my security group but I was only given the option of add/update/delete security groups.

We need some way of saying I don't want to change a list when going through the settings.

normj (changes requested):

src/AWS.Deploy.CLI/ConsoleUtilities.cs
philasmar: Right now, if you choose to Add a new value and leave it empty, it gives the behavior you are looking for. Though i need a more explicit way to do that. I'll look into it

src/AWS.Deploy.Recipes/RecipeDefinitions/ASP.NETAppAppRunner.recipe
philasmar: done

normj (approved):

ganeshnj (approved):

PR #473, opened 2022-03-18
feat: add Environment Variable setting to all recipes that support it

philasmar:
*Issue #, if available:*
DOTNET-5733

*Description of changes:*
* Add Environment Variable setting to all recipes that support it


normj (approved):

philasmar:
> 1. Should we be exercising these via any tests?
> 2. Should we update https://docs.aws.amazon.com/sdk-for-net/v3/developer-guide/deployment-tool-config-file.html to document how to set these in the config file?

We currently have a test for the KeyValue type which was introduced in this PR https://github.com/aws/aws-dotnet-deploy/pull/410/files#diff-d360d0e1d111d571e31b91bbf8de2b2d8a0f28f984128b7aac8f406fa95d259d. I wouldn't want to add a test for every new setting that is added.

Concerning the doc updates, we should get them updated, but I think since the doc team is understaffed and we are waiting for the Github pages solution, we should hold off on it.

ashovlin (approved):

PR #480, opened 2022-03-24
fix: cdk bootstrap using globally installed version of cdk

philasmar:
*Description of changes:*
CDK bootstrap is using the globally installed version of CDK. This is problematic when there is a version mismatch between the global one and the minimum version needed for the deploy tool.


normj (approved):
I think we need an overall better solution of detecting if bootstrap has already run and if the version is compatible with the recipe. If it is then don't run bootstrap. Also we should be looking at the recipe to see what version it needs. For now we can approve this as a stop gap but we should do some more work in this area.

ashovlin (approved):

PR #482, opened 2022-03-28
fix: stack name generator incorrectly increments projects

philasmar:
*Description of changes:*
Currently, if we have a project "WebApp1" and that project is already deployed, when we try to create a new stack based on that same project, the suggested name will be "WebApp11" instead of "WebApp2". 
This PR fixes this behavior by extracting the number from the end of the project name and increments that instead of just tacking a new number to the end of the project name.

### DIFF
--- a/src/AWS.Deploy.Orchestration/Utilities/CloudApplicationNameGenerator.cs
+++ b/src/AWS.Deploy.Orchestration/Utilities/CloudApplicationNameGenerator.cs
@@ -67,13 +67,23 @@ namespace AWS.Deploy.Orchestration.Utilities
 
             // make sure the recommendation doesn't exist already in existingApplications
             var recommendation = recommendedPrefix;
-            var suffix = 1;
+            var suffix = 0;
+            var recommendationCharArray = recommendation.ToCharArray();
+            for (var i = recommendationCharArray.Length - 1; i >= 0; i--)
+            {
+                if (char.IsDigit(recommendationCharArray[i]))
+                    suffix = suffix * 10 + (recommendationCharArray[i] - '0');
+                else
+                    break;
+            }
+
+            var prefix = suffix != 0 ? recommendation[..^suffix.ToString().Length] : recommendedPrefix;
             while (suffix < 100)
             {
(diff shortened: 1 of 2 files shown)


src/AWS.Deploy.Orchestration/Utilities/CloudApplicationNameGenerator.cs
```diff
@@ -68,12 +68,24 @@ public string GenerateValidName(ProjectDefinition target, List<CloudApplication>
...
+            var parseResult = int.TryParse(digits, out suffix);
+
             while (suffix < 100)
             {
                 if (existingApplications.All(x => x.Name != recommendation) && IsValidName(recommendation))
                     return recommendation;
 
-                recommendation = $"{recommendedPrefix}{suffix++}";
+                recommendation = $"{(parseResult ? recommendation.Substring(0, recommendation.Length - digits.Length) : recommendedPrefix)}{++suffix}";
             }
```
ganeshnj, line 89: some calculations and string parsing be avoided
```csharp
            var recommendationCharArray = recommendation.ToCharArray();
            for (var i = recommendationCharArray.Length - 1; i >= 0; i--)
            {
                if (char.IsDigit(recommendationCharArray[i]))
                    suffix = suffix * 10 + (recommendationCharArray[i] - '0');
                else
                    break;
            }

            var prefix = suffix != 0 ? recommendation[..^suffix.ToString().Length] : recommendedPrefix;
            while (suffix < 100)
            {
                if (existingApplications.All(x => x.Name != recommendation) && IsValidName(recommendation))
                    return recommendation;

                recommendation = $"{prefix}{++suffix}";
            }
```

ganeshnj (changes requested):

src/AWS.Deploy.Orchestration/Utilities/CloudApplicationNameGenerator.cs
philasmar: Done

ganeshnj (approved):

PR #483, opened 2022-03-31
chore: Update third party licenses with current dependencies

ashovlin:
*Issue #, if available:*
DOTNET-5765

*Description of changes:*
Updating THIRD_PARTY_LICENSES to match our current dependency graph. This was reviewed internally also.

The rough categories of changes:
1. New AWSSDK dependencies (and upgrading others to 3.7)
2. Swapping out CDK dependencies with V1 to V2 upgrade
3. Adding Swashbuckle and OpenApi for server mode

### DIFF
--- a/THIRD_PARTY_LICENSES
+++ b/THIRD_PARTY_LICENSES
@@ -1,246 +1,228 @@
-**  Amazon.CDK; version 1.89.0 -- https://www.nuget.org/packages/Amazon.CDK
-** Amazon.CDK.CloudAssembly.Schema; version 1.89.0 --
-https://www.nuget.org/packages/Amazon.CDK.CloudAssembly.Schema/
-** Amazon.CDK.CXAPI; version 1.89.0 --
-https://www.nuget.org/packages/Amazon.CDK.CXAPI
-** Amazon.CDK.RegionInfo; version 1.89.0 --
-https://www.nuget.org/packages/Amazon.CDK.RegionInfo
-** Amazon.JSII.Runtime; version 1.15.0 --
-https://www.nuget.org/packages/Amazon.JSII.Runtime
-** AWSSDK.CloudFormation; version 3.5.2.27 --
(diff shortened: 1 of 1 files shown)


THIRD_PARTY_LICENSES
```diff
@@ -1,246 +1,232 @@
...
+    Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+* For AWSSDK.SimpleNotificationService see also this required NOTICE:
+    Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+* For AWSSDK.SQS see also this required NOTICE:
+    Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
 * For Amazon.JSII.Runtime see also this required NOTICE:
     AWS Cloud Development Kit (AWS CDK)
-    Copyright 2018-2018 Amazon.com, Inc. or its affiliates. All Rights
-    Reserved.
+    Copyright 2018-2018 Amazon.com, Inc. or its affiliates. All Rights Reserved.
```
normj, line 229: Do we need the year for the copyright. If we do I'm sure 2018-2018 doesn't make sense.

normj (changes requested):

THIRD_PARTY_LICENSES
ashovlin: This was largely generated from the internal tooling based on what is in the LICENSE/NOTICE files in the dependent repos:
1. CDK did update to _"2018 - 2022"_ since we first took the dependency, updated in a new commit - https://github.com/aws/aws-cdk/blob/99e12b5e6520bdfb64724b409a8a0609f753ebc3/LICENSE#L189
2. JSII still only lists 2018, so prefer to match them - https://github.com/aws/jsii/blob/8da74aaa5df97571908cc34d91de9256b7e3fe48/LICENSE#L190

normj (approved):

ganeshnj (approved):

PR #484, opened 2022-03-31
feat: Add shutdown endpoint to server mode to stop the server

ashovlin:
*Issue #, if available:*
DOTNET-5665

*Description of changes:*
Adds an `/api/v1/Health/Shutdown` endpoint to server mode to allow consumers to stop the server. This uses [IHostApplicationLifetime.StopApplication Method](https://docs.microsoft.com/en-us/dotnet/api/microsoft.extensions.hosting.ihostapplicationlifetime.stopapplication).

This can be called from:
1. The `RestAPIClient`'s new `Shutdown`/`ShutdownAsync` methods
2. `ServerModeSession.Shutdown` method

Added integration tests for ~~both~~ routes.
  * _Update: removed the test via `ServerModeSession`, I hadn't realized that this relies on already having the CLI installed locally._

### DIFF
--- a/src/AWS.Deploy.ServerMode.Client/RestAPI.cs
+++ b/src/AWS.Deploy.ServerMode.Client/RestAPI.cs
@@ -174,6 +174,21 @@ namespace AWS.Deploy.ServerMode.Client
         /// <exception cref="ApiException">A server side error occurred.</exception>
         System.Threading.Tasks.Task<RecipeSummary> GetRecipeAsync(string recipeId, string projectPath, System.Threading.CancellationToken cancellationToken);
     
+        /// <summary>Requests to stop the deployment tool. Any open sessions are implicitly closed.
+        /// This may return Microsoft.AspNetCore.Mvc.OkResult prior to the server being stopped,
+        /// clients may need to wait or check the health after requesting shutdown.</summary>
+        /// <returns>Success</returns>
+        /// <exception cref="ApiException">A server side error occurred.</exception>
+        System.Threading.Tasks.Task ShutdownAsync();
+    
+        /// <param name="cancellationToken">A cancellation token that can be used by other objects or threads to receive notice of cancellation.</param>
+        /// <summary>Requests to stop the deployment tool. Any open sessions are implicitly closed.
+        /// This may return Microsoft.AspNetCore.Mvc.OkResult prior to the server being stopped,
+        /// clients may need to wait or check the health after requesting shutdown.</summary>
+        /// <returns>Success</returns>
+        /// <exception cref="ApiException">A server side error occurred.</exception>
+        System.Threading.Tasks.Task ShutdownAsync(System.Threading.CancellationToken cancellationToken);
+    
     }
     
     [System.CodeDom.Compiler.GeneratedCode("NSwag", "13.10.9.0 (NJsonSchema v10.4.1.0 (Newtonsoft.Json v13.0.0.0))")]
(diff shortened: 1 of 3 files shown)


normj (approved):

src/AWS.Deploy.ServerMode.Client/RestAPI.cs
```diff
@@ -1579,12 +1661,12 @@ public partial class GetConfigSettingResourcesOutput
...
-    [System.CodeDom.Compiler.GeneratedCode("NJsonSchema", "10.4.1.0 (Newtonsoft.Json v12.0.0.0)")]
+    [System.CodeDom.Compiler.GeneratedCode("NJsonSchema", "10.4.1.0 (Newtonsoft.Json v13.0.0.0)")]
     public partial class GetDeploymentDetailsOutput 
     {
-        /// <summary>The CloudFormation Stack ID</summary>
-        [Newtonsoft.Json.JsonProperty("stackId", Required = Newtonsoft.Json.Required.Default, NullValueHandling = Newtonsoft.Json.NullValueHandling.Ignore)]
-        public string StackId { get; set; }
+        /// <summary>The CloudApplication Name</summary>
+        [Newtonsoft.Json.JsonProperty("cloudApplicationName", Required = Newtonsoft.Json.Required.Default, NullValueHandling = Newtonsoft.Json.NullValueHandling.Ignore)]
+        public string CloudApplicationName { get; set; }
```
awschristou, line 1669: Is this change related to the shutdown feature? If not it should be moved to its own PR

ashovlin: Still trying to track down what happened here, the actual change to the model was made in https://github.com/aws/aws-dotnet-deploy/commit/8d395c2542aa489693e759421d345b829cfa7d04 but appears it was not included in the generated `RestAPI`

philasmar (changes requested):

src/AWS.Deploy.CLI/ServerMode/Controllers/HealthController.cs
```diff
@@ -25,5 +31,16 @@ public HealthStatusOutput Get()
...
         }
+
+        /// <summary>
+        /// Requests to stop the deployment tool. Any open sessions are implicitly closed.
+        /// </summary>
+        [HttpPost("Shutdown")]
+        [SwaggerOperation(OperationId = "Shutdown")]
+        public IActionResult Shutdown()
+        {
+            _applicationLifetime.StopApplication();
```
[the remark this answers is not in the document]
ashovlin: When you start server mode via new `new ServerModeSession(port, port, "", true).Start();` (which I believe is close to how VS is doing it) you end up with a process tree like:

```
ServerModeClient.exe  A    // this is a console app that I'm testing this with
    cmd.exe  B    // "C:\WINDOWS\system32\cmd.exe" /c "dotnet aws" server-mode --port 1234 --parent-pid A
         dotnet.exe C    // dotnet  aws server-mode --port 1234 --parent-pid A
             dotnet-aws.exe  D    // C:\Users\shovlia\.dotnet\tools\dotnet-aws.exe" server-mode --port 1234 --parent-pid A
```
Watching Process Explorer, the current implementation using `StopApplication` will leave A running and kill B, C, and D.

---

I had also prototyped storing a reference to the process we start during [Run in AWS.Deploy.ServerMode.Client](https://github.com/aws/aws-dotnet-deploy/blob/a78bf14e49ebbdc1226fc97ded2e911355d5b75f/src/AWS.Deploy.ServerMode.Client/CommandLineWrapper.cs#L42), and then calling `process.Kill()` later, but that appeared to only kill B above and leave C and D running.

src/AWS.Deploy.CLI/ServerMode/Controllers/HealthController.cs
```diff
@@ -25,5 +31,16 @@ public HealthStatusOutput Get()
                 Status = SystemStatus.Ready
             };
         }
+
+        /// <summary>
+        /// Requests to stop the deployment tool. Any open sessions are implicitly closed.
+        /// </summary>
+        [HttpPost("Shutdown")]
+        [SwaggerOperation(OperationId = "Shutdown")]
+        public IActionResult Shutdown()
```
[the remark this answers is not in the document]
ashovlin: 1. Added `[Authorize]`
2. Moved to `/api/v1/Server/Shutdown` via a new `ServerController`.

src/AWS.Deploy.ServerMode.Client/RestAPI.cs
ashovlin: We extracted this change and released it separately as part of https://github.com/aws/aws-dotnet-deploy/releases/tag/0.38.3, `RestAPI` should contain only changes for this PR now.

philasmar (approved):

PR #500, opened 2022-04-08
fix: CDK not cleaning up resources after bootstrap failure

philasmar:
*Issue #, if available:*
DOTNET-5794

*Description of changes:*
This PR passes an updated CDK bootstrap Cloud Formation template to the bootstrap command. The custom template changes the StagingBucket retention policy to Delete. This fixes an issue where users with not enough permissions were stuck after a failed bootstrapping since cdk was not able to delete the staging bucket.


src/AWS.Deploy.Orchestration/CdkProjectHandler.cs
```diff
@@ -30,6 +31,8 @@ public class CdkProjectHandler : ICdkProjectHandler
         private readonly IDirectoryManager _directoryManager;
         private readonly IAWSResourceQueryer _awsResourceQueryer;
 
+        private const string TemplateIdentifier = "AWS.Deploy.Orchestration.CDK.CDKBootstrapTemplate.yaml";
```
ashovlin, line 34: Perhaps just for my understanding as someone newer to this project, but how do we decide whether to declare here vs. in `AWS.Deploy.Constants`?

philasmar: We don't really have a specific rule on what goes into the Constants project, but I personally take it on a case by case basis. Generally the Constants project has consts that are used in multiple places and so it made sense to consolidate them. For this specific use case, `TemplateIdentifier` refers to the location of the bootstrap template which is only used in this class and is specific to the Orchestration assembly. It didn't make a lot of sense for me to strip it out and put it in the Constants project.

src/AWS.Deploy.Orchestration/CdkProjectHandler.cs
```diff
@@ -66,6 +69,13 @@ public async Task<string> ConfigureCdkProject(OrchestratorSession session, Cloud
             await using var appSettingsFile = new StreamWriter(appSettingsFilePath);
             await appSettingsFile.WriteAsync(appSettingsBody);
 
+            // The CDK bootstrap template can be generated by running 'cdk bootstrap --show-template'.
+            // We need to keep the template up to date while making sure that the 'Staging Bucket' retention policies are set to 'Delete'.
+            var cdkBootstrapTemplate = typeof(CdkProjectHandler).Assembly.ReadEmbeddedFile(TemplateIdentifier);
+            var cdkBootstrapTemplateFilePath = Path.Combine(cdkProjectPath, "CDKBootstrapTemplate.yaml");
```
normj, line 75: We shouldn't be writing the bootstrap template to the CDK project path. That will be really confusing for persisted deployment projects.

normj (changes requested):

src/AWS.Deploy.Orchestration/CdkProjectHandler.cs
philasmar: Done. It is now saved to the global temp folder.

normj (approved):

ashovlin (commented):
Changes look fine, but wondering if we should have a mechanism to prompt us to update this as CDK evolves?

Per https://docs.aws.amazon.com/cdk/v2/guide/bootstrapping.html#bootstrapping-contract
> If you provide your own bootstrap template, keep it up-to-date with the canonical default template to ensure that yours continues to work with all CDK features.

I'm imagining a test that could run `cdk bootstrap --show-template` and diff against our own copy, and then fail if it detects a difference.

philasmar:
The consensus right now is that we will update the template when we do any changes or updates to CDK in our tool. We already have a predefined version of CDK in the code and so a template update wont catch us by surprise. So I think we can stick with that process for the near future.

ashovlin (approved):

96malhar (approved):

PR #502, opened 2022-04-12
feat: add VPC, Subnets and Security Groups settings to Beanstalk Recipe

philasmar:
*Issue #, if available:*
DOTNET-5459

*Description of changes:*
* In order to set the security group in CDK, I needed to set the VPC and Subnets. This is why I added setting for all 3. Subnets depends on VPC not being empty, and Security Groups depend on VPC and Subnets not being empty.
* I added a new functionality for setting dependency where you can specify an operation which for now is limited to "NotEmpty". That checks if the dependency is not null or empty in case of a string type, or if the List is not empty in case of List type.


test/AWS.Deploy.CLI.UnitTests/RecommendationTests.cs
```diff
@@ -419,6 +419,52 @@ public async Task IsDisplayable_ManyDependencies()
             Assert.True(fargateRecommendation.IsOptionSettingDisplayable(vpcIdOptionSetting));
         }
 
+        [Fact]
+        public async Task IsDisplayable_NotEmptyOperation()
```
awschristou, line 423: Please verify that the server mode tests are properly accepting and handling these new fields

src/AWS.Deploy.Common/Recipes/PropertyDependency.cs
```diff
@@ -9,14 +9,22 @@ namespace AWS.Deploy.Common.Recipes
...
+            PropertyDependencyOperationType? operation,
             object value)
         {
             Id = id;
+            Operation = operation;
             Value = value;
         }
     }
+
+    public enum PropertyDependencyOperationType
```
awschristou, line 26: Is the Toolkit (a server mode consumer) going to need to understand how to work with this?

src/AWS.Deploy.Common/Recipes/PropertyDependency.cs
```diff
@@ -9,14 +9,22 @@ namespace AWS.Deploy.Common.Recipes
...
+    {
+        NotEmpty
```
normj, line 28: Can we add `Equals` in this enum as well which is the implicit default.

src/AWS.Deploy.Common/Recommendation.cs
```diff
@@ -200,10 +200,30 @@ public bool IsOptionSettingDisplayable(OptionSettingItem optionSetting)
                 var dependsOnOptionSettingValue = GetOptionSettingValue(dependsOnOptionSetting);
                 if (
                     dependsOnOptionSetting != null &&
-                    dependsOnOptionSettingValue != null &&
-                    !dependsOnOptionSettingValue.Equals(dependency.Value))
+                    dependsOnOptionSettingValue != null)
                 {
-                    return false;
+                    if (dependency.Operation != null)
```
normj, line 205: Once you add `Equals` as the implicit default this logic can rid of a level of `if` nesting.

src/AWS.Deploy.Common/Recommendation.cs
```diff
@@ -200,10 +200,30 @@ public bool IsOptionSettingDisplayable(OptionSettingItem optionSetting)
...
-                    dependsOnOptionSettingValue != null &&
-                    !dependsOnOptionSettingValue.Equals(dependency.Value))
+                    dependsOnOptionSettingValue != null)
                 {
-                    return false;
+                    if (dependency.Operation != null)
+                    {
+                        if (dependency.Operation == PropertyDependencyOperationType.NotEmpty)
+                        {
+                            if (dependsOnOptionSetting.Type == OptionSettingValueType.List &&
```
normj, line 209: We should add maps (i.e. Environment Variables) as well.

normj (changes requested):

test/AWS.Deploy.CLI.UnitTests/RecommendationTests.cs
philasmar: I added a test for the new ExistingVPC typehint. The other fields already have tests and no need to redo.

src/AWS.Deploy.Common/Recipes/PropertyDependency.cs
philasmar: The setting dependency is abstracted and the toolkit will not need to interface with these properties.

src/AWS.Deploy.Common/Recipes/PropertyDependency.cs
philasmar: Done

src/AWS.Deploy.Common/Recommendation.cs
philasmar: Done

src/AWS.Deploy.Common/Recommendation.cs
```diff
@@ -200,10 +200,35 @@ public bool IsOptionSettingDisplayable(OptionSettingItem optionSetting)
...
+                    if (dependency.Operation == null ||
+                           dependency.Operation == PropertyDependencyOperationType.Equals)
+                    {
+                        if (!dependsOnOptionSettingValue.Equals(dependency.Value))
+                            return false;
+                    }
+                    else if (dependency.Operation == PropertyDependencyOperationType.NotEmpty)
+                    {
+                        if (dependsOnOptionSetting.Type == OptionSettingValueType.List &&
+                            dependsOnOptionSettingValue != null &&
```
normj, line 214: If the list has never been set then `dependsOnOptionSettingValue` will be null. That should count as being empty and we should return `false`. But the way I read this logic the `dependsOnOptionSettingValue != null` condition will cause it to skip the return `false`.

src/AWS.Deploy.Common/Recommendation.cs
```diff
@@ -200,10 +200,35 @@ public bool IsOptionSettingDisplayable(OptionSettingItem optionSetting)
...
+                            dependsOnOptionSettingValue.TryDeserialize<SortedSet<string>>(out var listValue) &&
+                            !listValue.Any())
+                        {
+                            return false;
+                        }
+                        else if (dependsOnOptionSetting.Type == OptionSettingValueType.KeyValue &&
+                            dependsOnOptionSettingValue != null &&
```
normj, line 221: Same comment as above about null value for `dependsOnOptionSettingValue`.

src/AWS.Deploy.Common/Recommendation.cs
```diff
@@ -200,10 +200,35 @@ public bool IsOptionSettingDisplayable(OptionSettingItem optionSetting)
...
                     dependsOnOptionSetting != null &&
-                    dependsOnOptionSettingValue != null &&
-                    !dependsOnOptionSettingValue.Equals(dependency.Value))
+                    dependsOnOptionSettingValue != null)
                 {
-                    return false;
+                    if (dependency.Operation == null ||
+                           dependency.Operation == PropertyDependencyOperationType.Equals)
+                    {
+                        if (!dependsOnOptionSettingValue.Equals(dependency.Value))
```
normj, line 208: Can you pass in the InvaritantCulture for comparison?

normj (changes requested):

src/AWS.Deploy.Common/Recommendation.cs
philasmar: The condition on line 203 https://github.com/aws/aws-dotnet-deploy/blob/842ae771dc31933231b30fe19abcadc269ca3534/src/AWS.Deploy.Common/Recommendation.cs#L203 makes sure that `dependsOnOptionSettingValue` is not null. In any case, `dependsOnOptionSettingValue` is never null because of https://github.com/aws/aws-dotnet-deploy/blob/842ae771dc31933231b30fe19abcadc269ca3534/src/AWS.Deploy.Common/Recipes/OptionSettingItem.ValueOverride.cs#L53 
`dependsOnOptionSettingValue` will be an empty string which will be handled in the if condition.

src/AWS.Deploy.Common/Recommendation.cs
philasmar: The condition on lineThe condition on line 203 https://github.com/aws/aws-dotnet-deploy/blob/842ae771dc31933231b30fe19abcadc269ca3534/src/AWS.Deploy.Common/Recommendation.cs#L203 makes sure that `dependsOnOptionSettingValue` is not null. In any case, `dependsOnOptionSettingValue` is never null because of https://github.com/aws/aws-dotnet-deploy/blob/842ae771dc31933231b30fe19abcadc269ca3534/src/AWS.Deploy.Common/Recipes/OptionSettingItem.ValueOverride.cs#L53 
`dependsOnOptionSettingValue` will be an empty string which will be handled in the if condition.

src/AWS.Deploy.Common/Recommendation.cs
philasmar: both `dependsOnOptionSettingValue` and `dependency.Value` are objects not strings and i don't want to cast them to string to use StringComparison.InvariantCulture

src/AWS.Deploy.Common/Recommendation.cs
normj: But if `dependsOnOptionSettingValue` is null then the condition on line 203 will cause it to skip processing either the `Equals` or `NotEmpty` operations and then return true. Sounds like we are getting lucky now because we always return back at least an empty string but that is frail. Especially since we even do a null check in this method on a property that we say should never be null.

normj (changes requested):

src/AWS.Deploy.Common/Recommendation.cs
philasmar: fixed

src/AWS.Deploy.Common/Recommendation.cs
philasmar: Fixed

normj (approved):

src/AWS.Deploy.Common/Recipes/PropertyDependency.cs
ashovlin: Could you improve documentation here anyway? As a new developer working on the tool this isn't clear to me, and I imagine developers writing custom recipes may need to understand this.
* Document enum values here
* Add a property for this to aws-deploy-recipe.schema-json (with a description, and backfill descriptions for `Id` and `value` too

src/AWS.Deploy.CLI/Commands/TypeHints/ExistingVpcCommand.cs
```diff
@@ -0,0 +1,91 @@
...
+using AWS.Deploy.Orchestration.Data;
+
+namespace AWS.Deploy.CLI.Commands.TypeHints
+{
+    public class ExistingVpcCommand : ITypeHintCommand
+    {
+        private readonly IAWSResourceQueryer _awsResourceQueryer;
+        private readonly IConsoleUtilities _consoleUtilities;
+
+        public ExistingVpcCommand(IAWSResourceQueryer awsResourceQueryer, IConsoleUtilities consoleUtilities)
```
ashovlin, line 20: Is this copied from existing `VpcCommand`, largely to set `CreateNew` set to `false`?

Is this a recurring need? Wondering if it'd be better to allow that configuration through the recipe, since this appears to be a fair amount of duplicated code.

Don't need to see changes here, more so curious.

ashovlin (changes requested):

PR #503, opened 2022-04-12
fix: remove --project-path argument from list-deployments command

96malhar:
*Description of changes:*
This PR removes the redundant `--project-path` CLI option from the `list-deployments` command.

### DIFF
--- a/src/AWS.Deploy.CLI/Commands/CommandFactory.cs
+++ b/src/AWS.Deploy.CLI/Commands/CommandFactory.cs
@@ -353,7 +353,6 @@ namespace AWS.Deploy.CLI.Commands
             {
                 listCommand.Add(_optionProfile);
                 listCommand.Add(_optionRegion);
-                listCommand.Add(_optionProjectPath);
                 listCommand.Add(_optionDiagnosticLogging);
             }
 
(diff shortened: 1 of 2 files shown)


test/AWS.Deploy.CLI.IntegrationTests/ServerMode/GetApplyOptionSettings.cs
```diff
@@ -73,7 +73,7 @@ private async Task<DeploymentStatus> WaitForDeployment(RestAPIClient restApiClie
             return (await restApiClient.GetDeploymentStatusAsync(sessionId)).Status;
         }
 
-        [Fact]
+        //[Fact]
```
ashovlin, line 76: I know we have https://github.com/aws/aws-dotnet-deploy/pull/504 working on this, but we shouldn't be commenting out tests just to sneak past CI.

ashovlin, line 76: I know we have https://github.com/aws/aws-dotnet-deploy/pull/504 working on this, but we shouldn't be commenting out tests just to sneak past CI.

philasmar:
I believe the project path switch was added to all commands early on. We then moved away from needing it as we were simply listing the CF stacks. That is why it is of no use in the current state of the command.

test/AWS.Deploy.CLI.IntegrationTests/ServerMode/GetApplyOptionSettings.cs
96malhar: Restored the fact attribute

ashovlin (approved):

PR #504, opened 2022-04-13
feat: Use CDK Diff to perform server mode integ tests

philasmar:
*Description of changes:*
* Add a StartDryRunDeployment operation in server mode that basically generates the CDK project and runs CDK Diff on that project
* In the integ test, we now check the generated CF template if it contains the correct settings.

### DIFF
--- a/src/AWS.Deploy.CLI/ServerMode/Controllers/DeploymentController.cs
+++ b/src/AWS.Deploy.CLI/ServerMode/Controllers/DeploymentController.cs
@@ -440,6 +440,43 @@ namespace AWS.Deploy.CLI.ServerMode.Controllers
         }
 
         /// <summary>
+        /// Creates the CloudFormation template that will be used by CDK for the deployment.
+        /// This operation returns the CloudFormation template that is created for this deployment.
+        /// </summary>
+        [HttpGet("session/<sessionId>/cftemplate")]
+        [SwaggerOperation(OperationId = "GenerateCloudFormationTemplate")]
+        [SwaggerResponse(200, type: typeof(GenerateCloudFormationTemplateOutput))]
+        [Authorize]
+        public async Task<IActionResult> GenerateCloudFormationTemplate(string sessionId)
+        {
+            var state = _stateServer.Get(sessionId);
+            if (state == null)
+            {
+                return NotFound($"Session ID {sessionId} not found.");
+            }
+
+            var serviceProvider = CreateSessionServiceProvider(state);
+
+            var orchestratorSession = CreateOrchestratorSession(state);
+
+            var orchestrator = CreateOrchestrator(state, serviceProvider);
+
+            var cdkProjectHandler = CreateCdkProjectHandler(state, serviceProvider);
+
+            if (state.SelectedRecommendation == null)
+                throw new SelectedRecommendationIsNullException("The selected recommendation is null or invalid.");
+
+            if (!state.SelectedRecommendation.Recipe.DeploymentType.Equals(Common.Recipes.DeploymentTypes.CdkProject))
+                throw new SelectedRecommendationIsIncompatibleException($"We cannot generate a CloudFormation template for the selected recommendation as it is not of type '{nameof(Models.DeploymentTypes.CloudFormationStack)}'.");
+
+            var task = new DeployRecommendationTask(orchestratorSession, orchestrator, state.ApplicationDetails, state.SelectedRecommendation);
+            var cloudFormationTemplate = await task.GenerateCloudFormationTemplate(cdkProjectHandler);
+            var output = new GenerateCloudFormationTemplateOutput(cloudFormationTemplate);
+
+            return Ok(output);
+        }
+
+        /// <summary>
         /// Begin execution of the deployment.
         /// </summary>
         [HttpPost("session/<sessionId>/execute")]
@@ -454,6 +491,8 @@ namespace AWS.Deploy.CLI.ServerMode.Controllers
             }
 
             var serviceProvider = CreateSessionServiceProvider(state);
+
+            var orchestratorSession = CreateOrchestratorSession(state);
 
             var orchestrator = CreateOrchestrator(state, serviceProvider);
 
@@ -473,7 +512,7 @@ namespace AWS.Deploy.CLI.ServerMode.Controllers
             if (capabilities.Any())
                 return Problem($"Unable to start deployment due to missing system capabilities.{Environment.NewLine}{missingCapabilitiesMessage}");
 
-            var task = new DeployRecommendationTask(orchestrator, state.ApplicationDetails, state.SelectedRecommendation);
+            var task = new DeployRecommendationTask(orchestratorSession, orchestrator, state.ApplicationDetails, state.SelectedRecommendation);
             state.DeploymentTask = task.Execute();
 
             return Ok();
@@ -601,6 +640,21 @@ namespace AWS.Deploy.CLI.ServerMode.Controllers
                     throw new FailedToRetrieveAWSCredentialsException("The tool was not able to retrieve the AWS Credentials."),
                 state.AWSRegion,
                 state.AWSAccountId);
+        }
+
+        private CdkProjectHandler CreateCdkProjectHandler(SessionState state, IServiceProvider? serviceProvider = null)
+        {
+            if (serviceProvider == null)
+            {
+                serviceProvider = CreateSessionServiceProvider(state);
+            }
+
+            return new CdkProjectHandler(
+                serviceProvider.GetRequiredService<IOrchestratorInteractiveService>(),
+                serviceProvider.GetRequiredService<ICommandLineWrapper>(),
+                serviceProvider.GetRequiredService<IAWSResourceQueryer>(),
+                serviceProvider.GetRequiredService<IFileManager>()
+                );
         }
 
         private Orchestrator CreateOrchestrator(SessionState state, IServiceProvider? serviceProvider = null, AWSCredentials? awsCredentials = null)
--- a/src/AWS.Deploy.CLI/ServerMode/Tasks/DeployRecommendationTask.cs
+++ b/src/AWS.Deploy.CLI/ServerMode/Tasks/DeployRecommendationTask.cs
@@ -15,10 +15,12 @@ namespace AWS.Deploy.CLI.ServerMode.Tasks
     {
         private readonly CloudApplication _cloudApplication;
         private readonly Orchestrator _orchestrator;
+        private readonly OrchestratorSession _orchestratorSession;
         private readonly Recommendation _selectedRecommendation;
 
-        public DeployRecommendationTask(Orchestrator orchestrator, CloudApplication cloudApplication, Recommendation selectedRecommendation)
+        public DeployRecommendationTask(OrchestratorSession orchestratorSession, Orchestrator orchestrator, CloudApplication cloudApplication, Recommendation selectedRecommendation)
         {
+            _orchestratorSession = orchestratorSession;
             _orchestrator = orchestrator;
             _cloudApplication = cloudApplication;
             _selectedRecommendation = selectedRecommendation;
@@ -28,6 +30,28 @@ namespace AWS.Deploy.CLI.ServerMode.Tasks
         {
             await CreateDeploymentBundle();
             await _orchestrator.DeployRecommendation(_cloudApplication, _selectedRecommendation);
+        }
+
+        /// <summary>
+        /// Generates the CloudFormation template that will be used by CDK for the deployment.
+        /// This involves creating a deployment bundle, generating the CDK project and running 'cdk diff' to get the CF template.
+        /// This operation returns the CloudFormation template that is created for this deployment.
+        /// </summary>
+        public async Task<string> GenerateCloudFormationTemplate(CdkProjectHandler cdkProjectHandler)
+        {
+            if (cdkProjectHandler == null)
+                throw new FailedToCreateCDKProjectException(DeployToolErrorCode.FailedToCreateCDKProject, $"We could not create a CDK deployment project due to a missing dependency '{nameof(cdkProjectHandler)}'.");
+
+            await CreateDeploymentBundle();
+            var cdkProject = await cdkProjectHandler.ConfigureCdkProject(_orchestratorSession, _cloudApplication, _selectedRecommendation);
+            try
+            {
+                return await cdkProjectHandler.PerformCdkDiff(cdkProject, _cloudApplication);
+            }
+            finally
+            {
+                cdkProjectHandler.DeleteTemporaryCdkProject(cdkProject);
+            }
         }
 
         private async Task CreateDeploymentBundle()
--- a/src/AWS.Deploy.Orchestration/CdkProjectHandler.cs
+++ b/src/AWS.Deploy.Orchestration/CdkProjectHandler.cs
@@ -19,7 +19,8 @@ namespace AWS.Deploy.Orchestration
         Task<string> ConfigureCdkProject(OrchestratorSession session, CloudApplication cloudApplication, Recommendation recommendation);
         string CreateCdkProject(Recommendation recommendation, OrchestratorSession session, string? saveDirectoryPath = null);
         Task DeployCdkProject(OrchestratorSession session, CloudApplication cloudApplication, string cdkProjectPath, Recommendation recommendation);
-        void DeleteTemporaryCdkProject(OrchestratorSession session, string cdkProjectPath);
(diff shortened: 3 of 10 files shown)


test/AWS.Deploy.CLI.IntegrationTests/ServerMode/GetApplyOptionSettings.cs
```diff
@@ -28,18 +33,17 @@ public class GetApplyOptionSettings : IDisposable
         private bool _isDisposed;
```
ashovlin, line 33: It won't let me comment on line 29 - but is `IntegrationTest` in the namespace still accurate?

If this is no longer relying on network calls and runs quickly (does it?), would `UnitTest` make it more likely that a developer working locally runs it?

test/AWS.Deploy.CLI.IntegrationTests/ServerMode/GetApplyOptionSettings.cs
```diff
@@ -53,24 +57,20 @@ public GetApplyOptionSettings()
...
-            await Task.Delay(TimeSpan.FromSeconds(3));
-
-            await WaitUntilHelper.WaitUntil(async () =>
-            {
-                DeploymentStatus status = (await restApiClient.GetDeploymentStatusAsync(sessionId)).Status; ;
-                return status != DeploymentStatus.Executing;
-            }, TimeSpan.FromSeconds(1), TimeSpan.FromMinutes(15));
-
-            return (await restApiClient.GetDeploymentStatusAsync(sessionId)).Status;
+            var testCredentials = FallbackCredentialsFactory.GetCredentials();
```
ashovlin, line 72: Do we still need to access the user's default credentials if we're not actually deploying anything?

src/AWS.Deploy.CLI/ServerMode/Controllers/DeploymentController.cs
```diff
@@ -439,6 +439,39 @@ public async Task<IActionResult> GetCompatibility(string sessionId)
             return Ok(output);
         }
 
+        /// <summary>
+        /// Start a dry run of the deployment.
+        /// </summary>
+        [HttpPost("session/<sessionId>/dryrun")]
```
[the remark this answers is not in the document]
philasmar: I looked into that approach but the dry run should be an optional parameter. However, Swagger does not support the optional parameter and ends up generating the RestClient with a required parameter which breaks backwards compatibility and forces the IDE to always provide the dry run flag. That's why I went with the approach of adding a new operation.

src/AWS.Deploy.CLI/ServerMode/Controllers/DeploymentController.cs
```diff
@@ -439,6 +439,39 @@ public async Task<IActionResult> GetCompatibility(string sessionId)
             return Ok(output);
         }
 
+        /// <summary>
+        /// Start a dry run of the deployment.
```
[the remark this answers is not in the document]
philasmar: A dry run of a deployment by our definition is generating the final CDK application along with the user properties, but not actually performing a "cdk deploy". I'll add that in the documentation.

test/AWS.Deploy.CLI.IntegrationTests/ServerMode/GetApplyOptionSettings.cs
```diff
@@ -144,48 +145,31 @@ await restClient.ApplyConfigSettingsAsync(sessionId, new ApplyConfigSettingsInpu
...
 
-                Assert.Equal(_stackName, existingDeployment.Name);
-                Assert.Equal(appRunnerRecommendation.RecipeId, existingDeployment.RecipeId);
-                Assert.Equal(appRunnerRecommendation.Name, existingDeployment.RecipeName);
-                Assert.Equal(appRunnerRecommendation.ShortDescription, existingDeployment.ShortDescription);
-                Assert.Equal(appRunnerRecommendation.Description, existingDeployment.Description);
-                Assert.Equal(appRunnerRecommendation.TargetService, existingDeployment.TargetService);
-                Assert.Equal(DeploymentTypes.CloudFormationStack, existingDeployment.DeploymentType);
+                Assert.True(metadata.Settings.ContainsKey("VPCConnector"));
+                var vpcConnector = JsonConvert.DeserializeObject<VPCConnectorTypeHintResponse>(metadata.Settings["VPCConnector"].ToString());
```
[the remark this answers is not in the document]
philasmar: In simple terms yes, but the writing to JSON portion is done as part of the process of generating a CDK project and translating the Recommendation and setting configuration to a way the CDK project will understand. Then we are using CDK to compile the project and generate the CF template as it would during a normal deployment. Then we are inspecting the settings as part of the template. 
For the CDK assertion module you linked, I'm not quite sure how it would fit into this. Maybe we can have an offline discussion about it?

test/AWS.Deploy.CLI.IntegrationTests/ServerMode/GetApplyOptionSettings.cs
philasmar: I'd still want to keep it under `IntergationTest` namespace as it is still technically an Integration Test even if it is not accessing the network. Developers could still run individual classes from the project without running the ones that require network calls. There are some tests in there that are similar.

src/AWS.Deploy.CLI/ServerMode/Controllers/DeploymentController.cs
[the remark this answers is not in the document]
philasmar: Done

src/AWS.Deploy.CLI/ServerMode/Models/StartDryRunDeploymentOutput.cs
```diff
@@ -0,0 +1,15 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+namespace AWS.Deploy.CLI.ServerMode.Models
+{
+    public class StartDryRunDeploymentOutput
+    {
+        public string CDKProject { get; set; }
+
+        public StartDryRunDeploymentOutput(string cdkProject)
```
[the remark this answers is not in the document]
philasmar: Done

96malhar (approved):

test/AWS.Deploy.CLI.IntegrationTests/ServerMode/GetApplyOptionSettings.cs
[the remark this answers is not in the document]
ashovlin: Okay. Revisiting those docs, CDK's testing looks good for when you're in working in within the actual CDK project, but agree we're removed from that here.

src/AWS.Deploy.CLI/ServerMode/Controllers/DeploymentController.cs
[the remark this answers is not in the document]
ashovlin: Ah, and we don't have a request shape for `StartDeployment` so that may be breaking anyway - though if there was a time to do it, we're still in preview.

I'll defer to others who have been involved with the project longer, but still a little wary that we're expanding our public surface area this much to fix a testing issue.

Interested in @normj 's opinion if you get a chance.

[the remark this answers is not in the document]
normj: I don't think you can have an optional resource path parameter. To make it optional you would need to make dryrun be a query string parameter. Can you try that and see if that works with the code generator?

src/AWS.Deploy.CLI/ServerMode/Controllers/DeploymentController.cs
```diff
@@ -439,6 +439,40 @@ public async Task<IActionResult> GetCompatibility(string sessionId)
...
 
+        /// <summary>
+        /// Start a dry run of the deployment, which performs all the steps in the process of performing a deployment except for running 'cdk bootstrap' and 'cdk deploy'.
+        /// This includes, but is not limited to, creating a deployment bundle and generating a CDK project.
+        /// </summary>
+        [HttpPost("session/<sessionId>/dryrun")]
+        [SwaggerOperation(OperationId = "StartDryRunDeployment")]
+        [SwaggerResponse(200, type: typeof(StartDryRunDeploymentOutput))]
+        [Authorize]
+        public async Task<IActionResult> StartDryRunDeployment(string sessionId)
```
normj, line 450: This doesn't really seem like a dry run operation. It just returns back the path to the CDK project that is generated leaving to testing code to do anything meaningful. If we are going to alter our public API we should make it meaningful and something we can expand on.

This whole operation could to generate the CloudFormation template and return back the template to the user. That can be useful later in the CLI experience as well for users that just want the template and then have the template carried forward in their pipeline to later be deployed. Also the toolkit might want to have a button to view the template before deploying it.

src/AWS.Deploy.CLI/ServerMode/Controllers/DeploymentController.cs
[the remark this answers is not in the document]
normj: Actually based on my later comment I don't think we should try and squeeze this into the StartDeployment and instead have a specific method for returning back the CF template that would be deployed.

src/AWS.Deploy.CLI/ServerMode/Controllers/DeploymentController.cs
```diff
@@ -439,6 +439,40 @@ public async Task<IActionResult> GetCompatibility(string sessionId)
...
+
+            var serviceProvider = CreateSessionServiceProvider(state);
+
+            var orchestratorSession = CreateOrchestratorSession(state);
+
+            var orchestrator = CreateOrchestrator(state, serviceProvider);
+
+            var cdkProjectHandler = CreateCdkProjectHandler(state, serviceProvider);
+
+            if (state.SelectedRecommendation == null)
```
normj, line 466: There should be another error if the selected recommendation is not CDK based.

src/AWS.Deploy.CLI/ServerMode/Tasks/DeployRecommendationTask.cs
```diff
@@ -30,6 +32,14 @@ public async Task Execute()
             await _orchestrator.DeployRecommendation(_cloudApplication, _selectedRecommendation);
         }
 
+        public async Task<string> CreateCDKDeploymentProject(CdkProjectHandler cdkProjectHandler)
+        {
+            await CreateDeploymentBundle();
```
normj, line 37: We should just be forwarding into the `Orchestrator` which already has access to the `cdkprojecthandler` and we also don't have to add in `OrchestratorSession`.

src/AWS.Deploy.CLI/ServerMode/Tasks/DeployRecommendationTask.cs
```diff
@@ -30,6 +32,14 @@ public async Task Execute()
...
+            var cdkProject = await cdkProjectHandler.ConfigureCdkProject(_orchestratorSession, _cloudApplication, _selectedRecommendation);
+            await cdkProjectHandler.PerformCdkDiff(cdkProject);
```
normj, line 39: A method called `CreateCDKDeploymentProject` should also need to run `PerformCdkDiff`. I understand why you are doing it but we should be clear with our names and intents.

normj (changes requested):

src/AWS.Deploy.CLI/ServerMode/Controllers/DeploymentController.cs
philasmar: Done

src/AWS.Deploy.CLI/ServerMode/Controllers/DeploymentController.cs
philasmar: done

src/AWS.Deploy.CLI/ServerMode/Tasks/DeployRecommendationTask.cs
philasmar: Done

src/AWS.Deploy.CLI/ServerMode/Tasks/DeployRecommendationTask.cs
philasmar: Done

src/AWS.Deploy.Orchestration/CdkProjectHandler.cs
```diff
@@ -69,6 +73,21 @@ public async Task<string> ConfigureCdkProject(OrchestratorSession session, Cloud
             return cdkProjectPath;
         }
 
+        public async Task<string> PerformCdkDiff(string cdkProjectPath, CloudApplication cloudApplication)
```
ashovlin, line 76: Throughout this entire call stack - can you document what the return type is? Based on the function name assumed it'd be the `cdk diff` output, but looks like it's the full body of the CloudFormation template instead?

src/AWS.Deploy.CLI/AWS.Deploy.CLI.csproj
```diff
@@ -43,6 +43,4 @@
     <None Include="..\..\icon.png" Pack="true" PackagePath="" />
   </ItemGroup>
 
-  <Import Project="..\AWS.Deploy.Constants\AWS.Deploy.Constants.projitems" Label="Shared" />
```
ashovlin, line 46: Why the change here and in the sln? I'm unsure how it's related.

src/AWS.Deploy.Orchestration/CdkProjectHandler.cs
philasmar: I added documentation on the call stack

src/AWS.Deploy.CLI/AWS.Deploy.CLI.csproj
philasmar: This is in response to Norm's comment about using the orchestrator to get to cdkprojecthandler. cdkprojecthandler is marked internal in the orchestrator, so i had to mark its internals visible to the CLI. By doing that I cause a conflict related to the constants project, so I had to remove it's reference from the CLI project as it can access them through the orchestrator.

ashovlin (approved):

src/AWS.Deploy.Orchestration/Properties/AssemblyInfo.cs
```diff
@@ -0,0 +1,7 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+using System.Runtime.CompilerServices;
+
+[assembly: InternalsVisibleTo("AWS.Deploy.CLI")]
```
normj, line 6: CLI should not need `InternalsVisibleTo`

src/AWS.Deploy.CLI/ServerMode/Tasks/DeployRecommendationTask.cs
```diff
@@ -30,6 +30,30 @@ public async Task Execute()
...
+            if (_orchestrator._session == null)
+                throw new FailedToCreateCDKProjectException(DeployToolErrorCode.FailedToCreateCDKProject, $"We could not create a CDK deployment project due to a missing dependency '{nameof(_orchestrator._session)}'.");
+            if (_orchestrator._cdkProjectHandler == null)
+                throw new FailedToCreateCDKProjectException(DeployToolErrorCode.FailedToCreateCDKProject, $"We could not create a CDK deployment project due to a missing dependency '{nameof(_orchestrator._cdkProjectHandler)}'.");
+
+            await CreateDeploymentBundle();
+            var cdkProject = await _orchestrator._cdkProjectHandler.ConfigureCdkProject(_orchestrator._session, _cloudApplication, _selectedRecommendation);
+            try
+            {
+                return await _orchestrator._cdkProjectHandler.PerformCdkDiff(cdkProject, _cloudApplication);
```
normj, line 49: From the CLI project we should not be accessing internal members of the `Orchestrator`. I actually missed that we did in a previous PR made these fields internal. That was probably shouldn't have done.

normj (changes requested):

src/AWS.Deploy.Orchestration/Properties/AssemblyInfo.cs
philasmar: Fixed

src/AWS.Deploy.CLI/ServerMode/Tasks/DeployRecommendationTask.cs
philasmar: Fixed

normj (approved):

Issue #155, opened 2021-03-16
The filename for Dockefile used should be configurable.


normj:
**Describe the bug**
Currently the tool is hardcoded to use `Dockerfile` for the Dockerfile to pass to docker build. Although that is most common it is not uncommon to use a different file name for different scenarios. Imagine a project having `Dockerfile.local` and Dockerfile.aws`. 

The tool should allow customizing the name of the Dockerfile passed into docker build.
---

This is a :bug: bug report

ericis:
@ganeshnj and @normj suggest that this allow for an entire path override and not just the file name. Our project stored this file under "./docker/Dockerfile". This alternative path outside of the root was necessary for other tooling.

ashovlin:
This was addressed by https://github.com/aws/aws-dotnet-deploy/pull/509, which was released in [0.46.16](https://github.com/aws/aws-dotnet-deploy/releases/tag/0.46.16).

PR #509, opened 2022-04-15
feat: Add support for custom Dockerfile paths

ashovlin:
*Issue #, if available:* https://github.com/aws/aws-dotnet-deploy/issues/155, DOTNET-5573

*Description of changes:* Adds support for `DockerfilePath` to the container deployments. This provides an alternative to the default location we check: `{ProjectPath}.Dockerfile`.
* ~~This takes an absolute path. I had briefly explored supporting either absolute or relative to the project directory, but started simpler.~~ _Updated on 4/27 to take either an absolute path or one relative to the project directory._
* This does not influence `DockerExecutionDirectory`
* Added a typehint. It doesn't suggest any values, but would signal any UIs that wrap this to expose a filepicker.

### DIFF
--- a/src/AWS.Deploy.CLI/Commands/TypeHints/DockerExecutionDirectoryCommand.cs
+++ b/src/AWS.Deploy.CLI/Commands/TypeHints/DockerExecutionDirectoryCommand.cs
@@ -35,7 +35,7 @@ namespace AWS.Deploy.CLI.Commands.TypeHints
                     _optionSettingHandler.GetOptionSettingValue<string>(recommendation, optionSetting),
                     allowEmpty: true,
                     resetValue: _optionSettingHandler.GetOptionSettingDefaultValue<string>(recommendation, optionSetting) ?? "",
-                    validators: async executionDirectory => await ValidateExecutionDirectory(executionDirectory));
+                    validators: async executionDirectory => await ValidateExecutionDirectory(executionDirectory, recommendation));
 
             recommendation.DeploymentBundle.DockerExecutionDirectory = settingValue;
             return Task.FromResult<object>(settingValue);
@@ -45,19 +45,26 @@ namespace AWS.Deploy.CLI.Commands.TypeHints
         /// This method will be invoked to set the Docker execution directory in the deployment bundle
         /// when it is specified as part of the user provided configuration file.
         /// </summary>
-        /// <param name="recommendation">The selected recommendation settings used for deployment <see cref="Recommendation"/></param>
+        /// <param name="recommendation">The selected recommendation used for deployment <see cref="Recommendation"/></param>
         /// <param name="executionDirectory">The directory specified for Docker execution.</param>
         public async Task OverrideValue(Recommendation recommendation, string executionDirectory)
         {
-            var resultString = await ValidateExecutionDirectory(executionDirectory);
+            var resultString = await ValidateExecutionDirectory(executionDirectory, recommendation);
             if (!string.IsNullOrEmpty(resultString))
                 throw new InvalidOverrideValueException(DeployToolErrorCode.InvalidDockerExecutionDirectory, resultString);
             recommendation.DeploymentBundle.DockerExecutionDirectory = executionDirectory;
         }
 
-        private async Task<string> ValidateExecutionDirectory(string executionDirectory)
+        /// <summary>
+        /// Validates that the Docker execution directory exists as either an
+        /// absolute path or a path relative to the project directory.
+        /// </summary>
+        /// <param name="executionDirectory">Proposed Docker execution directory</param>
+        /// <param name="recommendation">The selected recommendation settings used for deployment</param>
+        /// <returns>Empty string if the directory is valid, an error message if not</returns>
+        private async Task<string> ValidateExecutionDirectory(string executionDirectory, Recommendation recommendation)
         {
-            var validationResult = await new DirectoryExistsValidator(_directoryManager).Validate(executionDirectory);
+            var validationResult = await new DirectoryExistsValidator(_directoryManager).Validate(executionDirectory, recommendation);
 
             if (validationResult.IsValid)
             {
--- a/src/AWS.Deploy.CLI/Commands/TypeHints/FilePathCommand.cs
+++ b/src/AWS.Deploy.CLI/Commands/TypeHints/FilePathCommand.cs
@@ -1,0 +1,69 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+using System.Collections.Generic;
+using System.Threading.Tasks;
+using AWS.Deploy.Common;
+using AWS.Deploy.Common.IO;
+using AWS.Deploy.Common.Recipes;
+using AWS.Deploy.Common.Recipes.Validation;
+using AWS.Deploy.Common.TypeHintData;
+
+namespace AWS.Deploy.CLI.Commands.TypeHints
+{
+    /// <summary>
+    /// Typehint that lets the user specify a path to a file.
+    /// This can either be an absolute path to the file or relative to the project path.
+    /// </summary>
+    public class FilePathCommand : ITypeHintCommand
+    {
+        private readonly IConsoleUtilities _consoleUtilities;
+        private readonly IOptionSettingHandler _optionSettingHandler;
+        private readonly IFileManager _fileManager;
+
+        public FilePathCommand(IConsoleUtilities consoleUtilities , IOptionSettingHandler optionSettingHandler, IFileManager fileManager)
+        {
+            _consoleUtilities = consoleUtilities;
+            _optionSettingHandler = optionSettingHandler;
+            _fileManager = fileManager;
+        }
+
+        /// <summary>
+        /// Not implemented, specific files are not suggested to the user
+        /// </summary>
+        /// <returns>Empty list</returns>
+        public Task<List<TypeHintResource>?> GetResources(Recommendation recommendation, OptionSettingItem optionSetting) => Task.FromResult<List<TypeHintResource>?>(null);
+
+        /// <summary>
+        /// Prompts the user to enter a path to a file
+        /// </summary>
+        public Task<object> Execute(Recommendation recommendation, OptionSettingItem optionSetting)
+        {
+            var typeHintData = optionSetting.GetTypeHintData<FilePathTypeHintData>();
+
+            var userFilePath = _consoleUtilities
+               .AskUserForValue(
+                   string.Empty,
+                   _optionSettingHandler.GetOptionSettingValue<string>(recommendation, optionSetting),
+                   allowEmpty: typeHintData?.AllowEmpty ?? true,
+                   resetValue: _optionSettingHandler.GetOptionSettingDefaultValue<string>(recommendation, optionSetting) ?? "") ;
+
+            return Task.FromResult<object>(userFilePath);
+        }
+
+        /// <summary>
+        /// This method will be invoked to set a file path setting in the deployment bundle
+        /// when it is specified as part of the user provided configuration file.
+        /// </summary>
+        /// <param name="recommendation">The selected recommendation settings used for deployment <see cref="Recommendation"/></param>
+        /// <param name="filePath">File path entered by the user</param>
+        public async Task OverrideValue(Recommendation recommendation, string filePath)
+        {
+            var validator = new FileExistsValidator(_fileManager);
+            var validationResult = await (validator as IOptionSettingItemValidator).Validate(filePath, recommendation);
+
+            if (!validationResult.IsValid)
+                throw new InvalidOverrideValueException(DeployToolErrorCode.InvalidFilePath, validationResult.ValidationFailedMessage ?? validator.ValidationFailedMessage);
+        }
+    }
+}
--- a/src/AWS.Deploy.Common/DeploymentBundles/DeploymentBundle.cs
+++ b/src/AWS.Deploy.Common/DeploymentBundles/DeploymentBundle.cs
@@ -1,6 +1,8 @@
 // Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
 // SPDX-License-Identifier: Apache-2.0
 
+using System.IO;
+
 namespace AWS.Deploy.Common
 {
     /// <summary>
@@ -17,6 +19,11 @@ namespace AWS.Deploy.Common
         /// The list of additional dotnet publish args passed to the target application.
         /// </summary>
         public string DockerBuildArgs { get; set; } = "";
+
+        /// <summary>
+        /// The path to the Dockerfile. This can either be an absolute path or relative to the project directory.
+        /// </summary>
+        public string DockerfilePath { get; set; } = "";
 
         /// <summary>
         /// The ECR Repository Name where the docker image will be pushed to.
--- a/src/AWS.Deploy.Common/Recipes/OptionSettingItem.ValueOverride.cs
+++ b/src/AWS.Deploy.Common/Recipes/OptionSettingItem.ValueOverride.cs
@@ -89,6 +89,8 @@ namespace AWS.Deploy.Common.Recipes
         /// <summary>
         /// Assigns a value to the OptionSettingItem.
         /// </summary>
+        /// <param name="valueOverride">Value to assign</param>
+        /// <param name="recommendation">Current deployment recommendation, may be used if the validator needs to consider properties other than itself</param>
         /// <exception cref="ValidationFailedException">
         /// Thrown if one or more <see cref="Validators"/> determine
         /// <paramref name="valueOverride"/> is not valid.
@@ -99,7 +101,7 @@ namespace AWS.Deploy.Common.Recipes
             {
                 foreach (var validator in validators)
                 {
-                    var result = await validator.Validate(valueOverride);
+                    var result = await validator.Validate(valueOverride, recommendation);
                     if (!result.IsValid)
                     {
                         throw new ValidationFailedException(DeployToolErrorCode.OptionSettingItemValueValidationFailed,
--- a/src/AWS.Deploy.Common/Recipes/Validation/IOptionSettingItemValidator.cs
+++ b/src/AWS.Deploy.Common/Recipes/Validation/IOptionSettingItemValidator.cs
@@ -6,12 +6,18 @@ using System.Threading.Tasks;
 namespace AWS.Deploy.Common.Recipes.Validation
 {
     /// <summary>
-    /// This interface outlines the framework for OptionSettingItem validators.
+    /// This interface outlines the framework for <see cref="OptionSettingItem"/> validators.
     /// Validators such as <see cref="RegexValidator"/> implement this interface and provide custom validation logic
     /// on OptionSettingItems
     /// </summary>
     public interface IOptionSettingItemValidator
     {
-        Task<ValidationResult> Validate(object input);
+        /// <summary>
+        /// Validates an override value for an <see cref="OptionSettingItem"/>
+        /// </summary>
+        /// <param name="input">Raw input for an option</param>
+        /// <param name="recommendation">Selected recommendation, which may be used if the validator needs to consider properties other than itself</param>
+        /// <returns>Whether or not the input is valid</returns>
+        Task<ValidationResult> Validate(object input, Recommendation recommendation);
     }
 }
--- a/src/AWS.Deploy.Common/Recipes/Validation/OptionSettingItemValidatorList.cs
+++ b/src/AWS.Deploy.Common/Recipes/Validation/OptionSettingItemValidatorList.cs
@@ -32,6 +32,10 @@ namespace AWS.Deploy.Common.Recipes.Validation
         /// <summary>
         /// Must be paired with <see cref="ExistingResourceValidator"/>
         /// </summary>
-        ExistingResource
+        ExistingResource,
+        /// <summary>
+        /// Must be paired with <see cref="FileExistsValidator"/>
+        /// </summary>
+        FileExists
     }
 }
--- a/src/AWS.Deploy.Common/Recipes/Validation/OptionSettingItemValidators/FileExistsValidator.cs
+++ b/src/AWS.Deploy.Common/Recipes/Validation/OptionSettingItemValidators/FileExistsValidator.cs
@@ -1,0 +1,56 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+using System.Threading.Tasks;
+using AWS.Deploy.Common.IO;
+
+namespace AWS.Deploy.Common.Recipes.Validation
+{
+    /// <summary>
+    /// Validates that a recipe or deployment bundle option with a FilePath typehint points to an actual file.
+    /// This can either be an absolute path to the file or relative to the project path
+    /// </summary>
+    public class FileExistsValidator : IOptionSettingItemValidator
+    {
+        private readonly IFileManager _fileManager;
+
+        public FileExistsValidator(IFileManager fileManager)
+        {
+            _fileManager = fileManager;
+        }
+
+        public string ValidationFailedMessage { get; set; } = "The specified file does not exist";
+
+        /// <summary>
+        /// Whether or not an empty filepath is valid (essentially whether this option is required)
+        /// </summary>
+        public bool AllowEmptyString { get; set; } = true;
+
+        public Task<ValidationResult> Validate(object input, Recommendation recommendation)
+        {
+            var inputFilePath = input?.ToString() ?? string.Empty;
+
+            if (string.IsNullOrEmpty(inputFilePath))
+            {
+                if (AllowEmptyString)
+                {
+                    return ValidationResult.ValidAsync();
+                }
+                else
+                {
+                    return ValidationResult.FailedAsync("A file must be specified");
+                }
+            }
+
+            // Otherwise if there is a value, verify that it points to an actual file
+            if (_fileManager.Exists(inputFilePath, recommendation.GetProjectDirectory()))
+            {
+                return ValidationResult.ValidAsync();
+            }
+            else
+            {
+                return ValidationResult.FailedAsync($"The specified file {inputFilePath} does not exist");
+            }
+        }
+    }
+}
--- a/src/AWS.Deploy.Common/Recipes/Validation/RecipeValidators/DockerfilePathValidator.cs
+++ b/src/AWS.Deploy.Common/Recipes/Validation/RecipeValidators/DockerfilePathValidator.cs
@@ -1,0 +1,58 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+using System.IO;
+using System.Threading.Tasks;
+using AWS.Deploy.Common.IO;
+using AWS.Deploy.Common.Utilities;
+
+namespace AWS.Deploy.Common.Recipes.Validation
+{
+    /// <summary>
+    /// This validates that the Dockerfile is within the build context.
+    ///
+    /// Per https://docs.docker.com/engine/reference/commandline/build/#text-files
+    /// "The path must be to a file within the build context."
+    /// </summary>
+    public class DockerfilePathValidator : IRecipeValidator
+    {
+        private readonly IDirectoryManager _directoryManager;
+        private readonly IFileManager _fileManager;
+
+        public DockerfilePathValidator(IDirectoryManager directoryManager, IFileManager fileManager)
+        {
+            _directoryManager = directoryManager;
+            _fileManager = fileManager;
+        }
+
+        public Task<ValidationResult> Validate(Recommendation recommendation, IDeployToolValidationContext deployValidationContext)
+        {
+            DockerUtilities.TryGetAbsoluteDockerfile(recommendation, _fileManager, _directoryManager, out var absoluteDockerfilePath);
+
+            // Docker execution directory has its own typehint, which sets the value here
+            var dockerExecutionDirectory = recommendation.DeploymentBundle.DockerExecutionDirectory;
+
+            // We're only checking the interaction here against a user-specified file and execution directory,
+            // it's still possible that we generate a dockerfile and/or compute the execution directory later.
+            if (absoluteDockerfilePath == string.Empty || dockerExecutionDirectory == string.Empty)
+            {
+                return ValidationResult.ValidAsync();
+            }
+
+            // Convert both to absolute paths in case they were specified relative to the project directory
+            var projectPath = recommendation.GetProjectDirectory();
+
+            var absoluteDockerExecutionDirectory = Path.IsPathRooted(dockerExecutionDirectory)
+                            ? dockerExecutionDirectory
+                            : _directoryManager.GetAbsolutePath(projectPath, dockerExecutionDirectory);
+
+            if (!_directoryManager.ExistsInsideDirectory(absoluteDockerExecutionDirectory, absoluteDockerfilePath))
+            {
+                return ValidationResult.FailedAsync($"The specified Dockerfile \"{absoluteDockerfilePath}\" is not located within " +
+                    $"the specified Docker execution directory \"{dockerExecutionDirectory}\"");
+            }
+
+            return ValidationResult.ValidAsync();
+        }
+    }
+}
--- a/src/AWS.Deploy.Common/TypeHintData/FilePathTypeHintData.cs
+++ b/src/AWS.Deploy.Common/TypeHintData/FilePathTypeHintData.cs
@@ -1,0 +1,42 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+namespace AWS.Deploy.Common.TypeHintData
+{
+    /// <summary>
+    /// Additional typehint data for file options
+    /// </summary>
+    public class FilePathTypeHintData
+    {
+        /// <summary>
+        /// Corresponds to a Filter property for a System.Windows.Forms.FileDialog
+        /// to determine the choices that would appear in the dialog box if a wrapping tool prompted the user for a file path via a UI.
+        public string Filter { get; set; } = "All files (*.*)|*.*";
+
+        /// <summary>
+        /// Corresponds to the DefaultExt property for a System.Windows.Forms.FileDialog
+        /// to specify the default extension used if the user specifies a file name
+        /// without an extension.
+        /// </summary>
+        public string DefaultExtension { get; set; } = "";
+
+        /// <summary>
+        /// Corresponds to the Title property for a System.Windows.Forms.FileDialog
+        /// to specify the title of the file dialog box.
+        /// </summary>
+        public string Title { get; set; } = "Open";
+
+        /// <summary>
+        /// Corresponds to the CheckFileExists property for a System.Windows.Forms.FileDialog
+        /// to indicate whether the dialog box should display a warning if the user specifies a file that does not exist.
+        /// </summary>
+        public bool CheckFileExists { get; set; } = true;
+
+        /// <summary>
+        /// Corresponds to the the AllowEmpty parameter for ConsoleUtilities.AskUserForValue
+        /// This lets a recipe option that uses the FilePathCommand typehint
+        /// control whether an empty value is allowed during CLI mode
+        /// </summary>
+        public bool AllowEmpty { get; set; } = true;
+    }
+}
--- a/src/AWS.Deploy.Common/Utilities/DockerUtilities.cs
+++ b/src/AWS.Deploy.Common/Utilities/DockerUtilities.cs
@@ -1,0 +1,99 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+using System.IO;
+using AWS.Deploy.Common.IO;
+using AWS.Deploy.Common.Recipes;
+
+namespace AWS.Deploy.Common.Utilities
+{
+    /// <summary>
+    ///  Utility methods for working with a recommendation's Docker configuration
+    /// </summary>
+    public static class DockerUtilities
+    {
+        /// <summary>
+        /// Gets the path of a Dockerfile if it exists at the default location: "{ProjectPath}/Dockerfile"
+        /// </summary>
+        /// <param name="recommendation">The selected recommendation settings used for deployment</param>
+        /// <param name="fileManager">File manager, used for validating that the Dockerfile exists</param>
+        /// <param name="dockerfilePath">Path to the Dockerfile, relative to the recommendation's project directory</param>
+        /// <returns>True if the Dockerfile exists at the default location, false otherwise</returns>
+        public static bool TryGetDefaultDockerfile(Recommendation recommendation, IFileManager? fileManager, out string dockerfilePath)
+        {
+            if (fileManager == null)
+            {
+                fileManager = new FileManager();
+            }
+
+            if (fileManager.Exists(Constants.Docker.DefaultDockerfileName, recommendation.GetProjectDirectory()))
+            {
+                // Set the default value to the OS-specific ".\Dockerfile"
+                dockerfilePath = Path.Combine(".", Constants.Docker.DefaultDockerfileName);
+                return true;
+            }
+            else
+            {
+                dockerfilePath = string.Empty;
+                return false;
+            }
+        }
+
+        /// <summary>
+        /// Gets the path of a the project's Dockerfile if it exists, from either a user-specified or the default location
+        /// </summary>
+        /// <param name="recommendation">The selected recommendation settings used for deployment</param>
+        /// <param name="fileManager">File manager, used for validating that the Dockerfile exists</param>
+        /// <param name="dockerfilePath">Path to a Dockerfile,which may be absolute or relative</param>
+        /// <returns>True if a Dockerfile is specified for this deployment, false otherwise</returns>
+        public static bool TryGetDockerfile(Recommendation recommendation, IFileManager fileManager, out string dockerfilePath)
+        {
+            dockerfilePath = recommendation.DeploymentBundle.DockerfilePath;
+
+            if (!string.IsNullOrEmpty(dockerfilePath))
+            {
+                // Double-check that it still exists in case it was move/deleted after being specified.
+                if (fileManager.Exists(dockerfilePath, recommendation.GetProjectDirectory()))
+                {
+                    return true;
+                }
+                else
+                {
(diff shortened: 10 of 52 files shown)


src/AWS.Deploy.Recipes/DeploymentBundleDefinitions/Container.deploymentbundle
```diff
@@ -11,6 +11,16 @@
             "AdvancedSetting": true,
             "Updatable": true
         },
+        {
+            "Id": "DockerfilePath",
+            "Name": "Dockerfile Path",
+            "Description": "Specifies the absolute path to the Dockerfile as an alternative to the default location.",
+            "Type": "String",
+            "TypeHint": "DockerfilePath",
```
normj, line 19: Can we have the `TypeHint` be generic like `FilePath` and then put in the `TypeHintData` additional information for the type of file to use. (i.e. File extensions, File Type). That way the VS toolkit can make a generic type hint that adds a browse button for anything with a `FilePath` TypeHint and optional add some filtering mechanism.

normj (changes requested):

src/AWS.Deploy.CLI/Commands/TypeHints/DockerfilePathCommand.cs
```diff
@@ -0,0 +1,81 @@
...
+            return Task.FromResult<object>(settingValue);
+        }
+
+        /// <summary>
+        /// This method will be invoked to set the Docker execution directory in the deployment bundle
+        /// when it is specified as part of the user provided configuration file.
+        /// </summary>
+        /// <param name="recommendation">The selected recommendation settings used for deployment <see cref="Recommendation"/></param>
+        /// <param name="path">Dockerfile path entered by the user</param>
+        public void OverrideValue(Recommendation recommendation, string path)
```
96malhar, line 56: This method needs to be invoked from https://github.com/aws/aws-dotnet-deploy/blob/main/src/AWS.Deploy.CLI/Commands/DeployCommand.cs#L475. 

Add a new case "DockerfilePath"

src/AWS.Deploy.CLI/Commands/TypeHints/DockerfilePathCommand.cs
```diff
@@ -0,0 +1,81 @@
...
+                   allowEmpty: true,
+                   resetValue: recommendation.GetOptionSettingDefaultValue<string>(optionSetting) ?? "",
+                   validators: dockerfilePath => ValidateDockerfilePath(dockerfilePath));
+
+            recommendation.DeploymentBundle.DockerfilePath = settingValue;
+            return Task.FromResult<object>(settingValue);
+        }
+
+        /// <summary>
+        /// This method will be invoked to set the Docker execution directory in the deployment bundle
```
96malhar, line 51: ...invoked to set the **Docker file path**.

96malhar (dismissed):

src/AWS.Deploy.Recipes/DeploymentBundleDefinitions/Container.deploymentbundle
```diff
@@ -11,6 +11,16 @@
...
+            "DefaultValue": "",
+            "AdvancedSetting": true,
+            "Updatable": true
+        },
```
philasmar, line 23: You could also add a regex validator to make sure that this is a FilePath. Or create a new validator that checks file paths.

philasmar (changes requested):
I also agree with the comments on making the current type hint a generic one.

src/AWS.Deploy.Recipes/DeploymentBundleDefinitions/Container.deploymentbundle
```diff
@@ -11,6 +11,16 @@
...
         {
             "Id": "DockerExecutionDirectory",
```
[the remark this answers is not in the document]
96malhar: If the `DockerExecutionDirectory` always needs to be the parent directory of `DockerfilePath`, then we shouldn't be exposing it to the user as an option setting. We can set it ourselves based on whatever the `DockerfilePath` is

src/AWS.Deploy.Recipes/DeploymentBundleDefinitions/Container.deploymentbundle
96malhar: IMO,  a new `FilePathExists` validators would suffice since it would verify that the path was not only structurally valid but also exists on disk.

src/AWS.Deploy.Recipes/DeploymentBundleDefinitions/Container.deploymentbundle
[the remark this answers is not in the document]
ashovlin: 1. Added a recipe-level validator to validate their relationship.
    * I had originally tried to list the validator in `Container.deploymentbundle` as opposed to in every Docker-based recipe, but looks like we're not set up to load validators from the deployment bundle files, only the list of options. Opted to add to the recipes for now, though I imagine we could consolidate in the future.
3. I don't believe it's always the direct parent, per https://docs.docker.com/engine/reference/commandline/build/#text-files _"The path must be to a file within the build context."_

Contrived example:
```
sln
    csproj <- we may want to build here
        /BuildTools
            /Docker
                /Dev
                    Dockerfile <- but this can be in any child
                /Prod
```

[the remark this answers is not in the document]
philasmar: We don't necessarily know what the `DockerExecutionDirectory` is. We still need to allow users to set it.

src/AWS.Deploy.CLI/Commands/TypeHints/FilePathCommand.cs
```diff
@@ -0,0 +1,65 @@
...
+        public Task<object> Execute(Recommendation recommendation, OptionSettingItem optionSetting)
+        {
+            var userFilePath = _consoleUtilities
+               .AskUserForValue(
+                   string.Empty,
+                   recommendation.GetOptionSettingValue<string>(optionSetting),
+                   allowEmpty: true,
+                   resetValue: recommendation.GetOptionSettingDefaultValue<string>(optionSetting) ?? "");
+
+            return Task.FromResult<object>(userFilePath);
```
normj, line 43: I think we should validate the filepath points to an existing file. If it does not tell the user and prompt again. Assuming filepath is not empty.

src/AWS.Deploy.Common/Recipes/Validation/OptionSettingItemValidators/FileValidator.cs
```diff
@@ -0,0 +1,39 @@
...
+            var userFilePath = input?.ToString() ?? string.Empty;
+
+            // Allow the user to clear the option intentionally
+            if (userFilePath == string.Empty)
+            {
+                return ValidationResult.Valid();
+            }
+
+            // Otherwise if there is a value, verify that it points to an actual file
+            if (fileManager.Exists(userFilePath))
```
normj, line 29: If `userFilePath` is relative the absolute path should be computed based on the current project directory set by either `--project-path` switch or during server mode. This will fail in this case because it will compute based on the current working directory which is not necessary the project directory.

src/AWS.Deploy.Common/Recipes/Validation/RecipeValidators/DockerfilePathValidator.cs
```diff
@@ -0,0 +1,42 @@
...
+            var dockerExecutionDirectory = recommendation.DeploymentBundle.DockerExecutionDirectory;
+
+            // We're only checking the interaction here against a user-specified file and execution directory,
+            // it's still possible that we generate a dockerfile and/or compute the execution directory later.
+            if (dockerfilePath == string.Empty || dockerExecutionDirectory == string.Empty)
+            {
+                return ValidationResult.Valid();
+            }
+
+            if (!directoryManager.ExistsInsideDirectory(dockerExecutionDirectory, dockerfilePath))
```
normj, line 34: Same comment as above make sure directory path look up takes into account project path.

src/AWS.Deploy.Common/TypeHintData/FilePathTypeHintData.cs
```diff
@@ -0,0 +1,35 @@
...
+        /// Corresponds to the Title property for a System.Windows.Forms.FileDialog
+        /// to specify the title of the file dialog box.
+        /// </summary>
+        public string Title { get; set; } = "Open";
+
+        /// <summary>
+        /// Corresponds to the CheckFileExists property for a System.Windows.Forms.FileDialog
+        /// to indicate whether the dialog box should display a warning if the user specifies a file that does not exist.
+        /// </summary>
+        public bool CheckFileExists { get; set; } = false;
```
normj, line 33: Wouldn't we want the default to be `true`?

src/AWS.Deploy.Common/Utilities/DockerUtilities.cs
```diff
@@ -0,0 +1,86 @@
...
+
+            if (!string.IsNullOrEmpty(dockerfilePath))
+            {
+                if (fileManager == null)
+                {
+                    fileManager = new FileManager();
+                }
+
+                // Double-check that it still exists in case it was move/deleted after being specified.
+                if (fileManager.Exists(dockerfilePath))
```
normj, line 67: Relative to project path.

src/AWS.Deploy.Recipes/DeploymentBundleDefinitions/Container.deploymentbundle
```diff
@@ -11,6 +11,26 @@
             "AdvancedSetting": true,
             "Updatable": true
         },
+        {
+            "Id": "DockerfilePath",
+            "Name": "Dockerfile Path",
+            "Description": "Specifies the absolute path to the Dockerfile as an alternative to the default location.",
```
normj, line 17: Why does it have to be an absolute path. Eventually we want to persist build settings like we do with the other recipe settings and when we do that we will want to persist the file path relative to the project path. Otherwise we deployment would get messed up as projects are deployed from different machines.

normj (changes requested):

src/AWS.Deploy.Recipes/DeploymentBundleDefinitions/Container.deploymentbundle
ashovlin: 1. Moved to generic `FilePathCommand` typehint
2. Added `FilePathTypeHintData`, to provide VS Toolkit with common properties that they can pass into a `System.Windows.Forms.FileDialog` to customize the file picker.

src/AWS.Deploy.CLI/Commands/TypeHints/DockerfilePathCommand.cs
```diff
@@ -0,0 +1,81 @@
...
+using AWS.Deploy.Common.IO;
+using AWS.Deploy.Common.Recipes;
+using AWS.Deploy.Common.TypeHintData;
+
+namespace AWS.Deploy.CLI.Commands.TypeHints
+{
+    /// <summary>
+    /// Typehint that lets the user specify a Dockerfile outside of the default location.
+    /// </summary>
+    public class DockerfilePathCommand : ITypeHintCommand
```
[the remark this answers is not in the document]
ashovlin: Moved to generic `FilePathCommand` typehint

src/AWS.Deploy.CLI/Commands/TypeHints/DockerfilePathCommand.cs
```diff
@@ -0,0 +1,81 @@
...
+        /// </summary>
+        public Task<object> Execute(Recommendation recommendation, OptionSettingItem optionSetting)
+        {
+            var settingValue = _consoleUtilities
+               .AskUserForValue(
+                   string.Empty,
+                   recommendation.GetOptionSettingValue<string>(optionSetting),
+                   allowEmpty: true,
+                   resetValue: recommendation.GetOptionSettingDefaultValue<string>(optionSetting) ?? "",
+                   validators: dockerfilePath => ValidateDockerfilePath(dockerfilePath));
```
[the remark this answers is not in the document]
ashovlin: Moved to `FileValidator.cs`

src/AWS.Deploy.CLI/Commands/TypeHints/DockerfilePathCommand.cs
```diff
@@ -0,0 +1,81 @@
...
+        {
+            if (!string.IsNullOrEmpty(filePath) && !_fileManager.Exists(filePath))
+            {
+                return "The specified Dockerfile does not exist.";
+            }
+            
+            return string.Empty;
+        }
+    }
+}
```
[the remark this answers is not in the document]
ashovlin: Moved to `FileValidator.cs`

src/AWS.Deploy.CLI/Commands/TypeHints/DockerfilePathCommand.cs
ashovlin: Added, thanks.

src/AWS.Deploy.CLI/Commands/TypeHints/DockerfilePathCommand.cs
ashovlin: This whole file was dropped on the move to a generic `FIlePathCommand`

src/AWS.Deploy.Orchestration/DeploymentBundleHandler.cs
```diff
@@ -157,13 +166,21 @@ private string GetDockerExecutionDirectory(Recommendation recommendation)
...
         }
 
-        private string GetDockerFilePath(Recommendation recommendation)
+        public string GetDockerfilePath(Recommendation recommendation)
         {
-            var dockerFileDirectory = new FileInfo(recommendation.ProjectPath).Directory?.FullName;
-            if (dockerFileDirectory == null)
-                throw new InvalidProjectPathException(DeployToolErrorCode.ProjectPathNotFound, "The project path is invalid.");
+            // User specified a path to a dockerfile 
+            if (!string.IsNullOrEmpty(recommendation.DeploymentBundle.DockerfilePath))
```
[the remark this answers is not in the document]
ashovlin: This is now calling ` DockerUtilities.TryGetDockerfile`, which does check if the file still exists.

Though, currently it'll still pass an empty string into `docker build` which would then fail. I wasn't sure if we should add another exception at this point, since either way the build isn't going to be successful.

src/AWS.Deploy.Orchestration/DeploymentBundleHandler.cs
```diff
@@ -157,13 +166,21 @@ private string GetDockerExecutionDirectory(Recommendation recommendation)
...
+                return recommendation.DeploymentBundle.DockerfilePath;
+            }
+            else // we assume it's at "{ProjectPath}/Dockerfile"
+            {
+                var dockerFileDirectory = new FileInfo(recommendation.ProjectPath).Directory?.FullName;
+                if (dockerFileDirectory == null)
+                    throw new InvalidProjectPathException(DeployToolErrorCode.ProjectPathNotFound, "The project path is invalid.");
 
-            return Path.Combine(dockerFileDirectory, "Dockerfile");
+                return Path.Combine(dockerFileDirectory, "Dockerfile");
```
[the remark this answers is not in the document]
ashovlin: This is now calling `DockerUtilities.TryGetDockerfile`, which does check the `OptionSetting`, then the default location.

src/AWS.Deploy.Recipes/DeploymentBundleDefinitions/Container.deploymentbundle
```diff
@@ -11,6 +11,16 @@
...
+            "Type": "String",
+            "TypeHint": "DockerfilePath",
+            "DefaultValue": "",
```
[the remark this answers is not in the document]
ashovlin: Added

src/AWS.Deploy.Recipes/DeploymentBundleDefinitions/Container.deploymentbundle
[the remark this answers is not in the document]
ashovlin: Added a recipe-level `DockerfilePathValidator` to check the relationship between the Dockerfile path and the Docker execution directory.

src/AWS.Deploy.Recipes/DeploymentBundleDefinitions/Container.deploymentbundle
ashovlin: Added `FileValidator.cs` to check for existence.

src/AWS.Deploy.CLI/Commands/TypeHints/FilePathCommand.cs
ashovlin: We do, it's done via an `IOptionSettingItemValidator` `FileValidator.cs`.

```
Dockerfile Path:
Specifies the absolute path to the Dockerfile as an alternative to the default location.
Enter value (default C:\Users\shovlia\source\repos\ASPWebAPI\ASPWebAPI\Dockerfile. Type <reset> to reset.):
asdf
The specified file asdf does not exist

Dockerfile Path:
Specifies the absolute path to the Dockerfile as an alternative to the default location.
Enter value (default C:\Users\shovlia\source\repos\ASPWebAPI\ASPWebAPI\Dockerfile. Type <reset> to reset.):
```

src/AWS.Deploy.Recipes/DeploymentBundleDefinitions/Container.deploymentbundle
ashovlin: I brought this up at a internal stand-up this past week and nobody raised concerns.

I agree supporting absolute or relative would be nice. The biggest obstacle is that a `IOptionSettingItemValidator` doesn't have access to the project path per https://github.com/aws/aws-dotnet-deploy/blob/main/docs/OptionSettingsItems-input-validation.md#dependencies. I'll see if I can extend that to be aware of the project path.

src/AWS.Deploy.Orchestration/DeploymentBundleHandler.cs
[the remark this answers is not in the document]
philasmar: It is a good idea to add an exception is the file doesn't exist as we know for sure the `docker build` would fail. If we leave it till that point, the exception we get will be a generic docker build exception which will not track well in telemetry.

src/AWS.Deploy.Common/Recipes/Validation/RecipeValidators/DockerfilePathValidator.cs
ashovlin: Throughout, switched to either an absolute or a relative path to the project directory.

src/AWS.Deploy.Common/TypeHintData/FilePathTypeHintData.cs
ashovlin: Updated to `true`. I was initially just matching the `FileDialog` defaults, but agree this is a safer default

src/AWS.Deploy.Common/Utilities/DockerUtilities.cs
ashovlin: Throughout, switched to either an absolute or a relative path to the project directory.

src/AWS.Deploy.Common/Recipes/Validation/IOptionSettingItemValidator.cs
```diff
@@ -10,6 +10,12 @@ namespace AWS.Deploy.Common.Recipes.Validation
     /// </summary>
     public interface IOptionSettingItemValidator
     {
-        ValidationResult Validate(object input);
+        /// <summary>
+        /// Validates an override value for an OptionSetting
```
philasmar, line 14: nit: add a ref to OptionSettingItem object. i.e <see cref="OptionSettingItem"/>

src/AWS.Deploy.Common/Recipes/Validation/OptionSettingItemValidators/FileValidator.cs
```diff
@@ -0,0 +1,41 @@
...
+// SPDX-License-Identifier: Apache-2.0
+
+using System.IO;
+using AWS.Deploy.Common.IO;
+
+namespace AWS.Deploy.Common.Recipes.Validation
+{
+    /// <summary>
+    /// Validates that a recipe or deployment bundle option with a FilePath typehint points to an actual file.
+    /// This can either be an absolute path to the file or relative to the project path for the current recommendation.
```
philasmar, line 11: nit: im not sure what you mean by `for the current recommendation` but i propose removing it.

philasmar, line 11: nit: im not sure what you mean by `for the current recommendation` but i propose removing it.

src/AWS.Deploy.Common/Recipes/Validation/OptionSettingItemValidators/FileValidator.cs
```diff
@@ -0,0 +1,41 @@
...
+    /// </summary>
+    public class FileValidator : IOptionSettingItemValidator
+    {
+        public static readonly string defaultValidationFailedMessage = "The specified file does not exist";
+
+        public string ValidationFailedMessage { get; set; } = defaultValidationFailedMessage;
+
+        ValidationResult IOptionSettingItemValidator.Validate(object input, Recommendation recommendation)
+        {
+            IFileManager fileManager = new FileManager();
```
philasmar, line 21: can you pass in IFileManager using dependency injection instead of instantiating it here?

src/AWS.Deploy.Common/Recipes/Validation/OptionSettingItemValidators/FileValidator.cs
```diff
@@ -0,0 +1,41 @@
...
+            var inputFilePath = input?.ToString() ?? string.Empty;
+
+            // Allow the user to clear the option intentionally
+            if (inputFilePath == string.Empty)
+            {
+                return ValidationResult.Valid();
```
philasmar, line 27: i'm curious why an empty value is considered valid?

src/AWS.Deploy.Common/Recipes/Validation/OptionSettingItemValidators/FileValidator.cs
```diff
@@ -0,0 +1,41 @@
...
+            }
+
+            // Otherwise if there is a value, verify that it points to an actual file
+            if (fileManager.Exists(inputFilePath, recommendation.GetProjectDirectory()))
```
philasmar, line 31: If the only reason we need the recommendation object is to get the project directory, which in turn is used to check the path relative to the project directory, then why don't we add the project directory to the file and directory managers as the base directory and set that at startup. That way through dependency injection, you wouldn't need the recommendation object here.

src/AWS.Deploy.Common/Recipes/Validation/RecipeValidators/DockerfilePathValidator.cs
```diff
@@ -0,0 +1,53 @@
...
+    /// This validates that the Dockerfile is within the build context.
+    ///
+    /// Per https://docs.docker.com/engine/reference/commandline/build/#text-files
+    /// "The path must be to a file within the build context."
+    /// </summary>
+    public class DockerfilePathValidator : IRecipeValidator
+    {
+        public ValidationResult Validate(Recommendation recommendation, IDeployToolValidationContext deployValidationContext)
+        {
+            IDirectoryManager directoryManager = new DirectoryManager();
```
philasmar, line 19: IDirectoryManager should be passed using dependency injection and not instantiated here.

src/AWS.Deploy.Common/Recipes/Validation/RecipeValidators/DockerfilePathValidator.cs
```diff
@@ -0,0 +1,53 @@
...
+
+            // Dockerfile uses the generic FilePath typehint, so we must load the latest value from the option setting.
+            // We also assume that if we detected a Dockerfile automatically it was set here via the replacement token
+            var dockerfilePath = recommendation.GetOptionSettingValue<string>(recommendation.GetOptionSetting("DockerfilePath"));
```
philasmar, line 23: I don't like how we are getting the DockerfilePath, we should be able to distinguish DeploymentBundle settings from regular Option settings

philasmar (changes requested):

96malhar (commented):
Possible alternate approach to handle file paths relative to the docker file:
* Set the project path as a **process-specific** environment variable in the [CommandFactory](https://github.com/aws/aws-dotnet-deploy/blob/main/src/AWS.Deploy.CLI/Commands/CommandFactory.cs#L174)
* Retrieve this environment variable in the `FilePathCommand` type hint and resolve the `userFilePath` to an absolute path by calling [Path.GetFullPath(userFilePath, projectPath)](https://docs.microsoft.com/en-us/dotnet/api/system.io.path.getfullpath?view=net-6.0#system-io-path-getfullpath(system-string-system-string))
* Always pass in the absolute path to `FileValidator`

This approach will avoid (unnecessarily) passing the `Recommendation` object to all the validate calls.

src/AWS.Deploy.Common/Recipes/Validation/RecipeValidators/DockerfilePathValidator.cs
ashovlin: Why don't you like this? Or why is it important to distinguish where the setting came from in the validator?

My understanding of here is that we're treating all OptionSettings the same, whether they came from the recipe or the deployment bundle.
https://github.com/aws/aws-dotnet-deploy/blob/f2e5db538a032be70de1dc3953b297528a62afeb/src/AWS.Deploy.Common/Recommendation.cs#L107

That same OptionSetting will be in this property, but it doesn't have the helper to find the one we're looking for.
https://github.com/aws/aws-dotnet-deploy/blob/f2e5db538a032be70de1dc3953b297528a62afeb/src/AWS.Deploy.Common/Recommendation.cs#L33

src/AWS.Deploy.Common/Recipes/Validation/OptionSettingItemValidators/FileValidator.cs
ashovlin: Based on this it doesn't seem like we're set up to inject into the `IOptionSettingItemValidator`:
https://github.com/aws/aws-dotnet-deploy/blob/f2e5db538a032be70de1dc3953b297528a62afeb/docs/OptionSettingsItems-input-validation.md?plain=1#L142-L146

src/AWS.Deploy.Common/Recipes/Validation/RecipeValidators/DockerfilePathValidator.cs
ashovlin: I think similar to `IOptionSettingItemValidator`, we're not set up to inject into recipe validators because they're designed to be constructed without any parameters.

https://github.com/aws/aws-dotnet-deploy/blob/f2e5db538a032be70de1dc3953b297528a62afeb/docs/OptionSettingsItems-input-validation.md?plain=1#L142-L146

src/AWS.Deploy.Common/TypeHintData/FilePathTypeHintData.cs
```diff
@@ -0,0 +1,35 @@
...
+        /// Corresponds to a Filter property for a System.Windows.Forms.FileDialog
+        /// to determine the choices that would appear in the dialog box if a wrapping tool prompted the user for a file path via a UI.
+        public string Filter { get; set; } = "All files (*.*)|*.*";
+
+        /// <summary>
+        /// Corresponds to the DefaultExt property for a System.Windows.Forms.FileDialog
+        /// to specify the default extension used if the user specifies a file name
+        /// without an extension.
+        /// </summary>
+        public string DefaultExtension { get; set; } = "";
```
floralph, line 21: This property doesn't seem to be defined in TypeHintData for FilePath in src/AWS.Deploy.Recipes/DeploymentBundleDefinitions/Container.deploymentbundle in this CR.  Should it be?

src/AWS.Deploy.CLI/Commands/TypeHints/FilePathCommand.cs
ashovlin: At the time I didn't realize that `AskUserForValue` has a `validators` parameter.

`FileExistsValidator : `IOptionSettingItemValidator` implements the file check. I've wrapped that in `ValidateFileExists` here to match the signature for `AskUserForValue`

src/AWS.Deploy.CLI/Commands/DeployCommand.cs
```diff
@@ -443,7 +443,7 @@ private void ConfigureDeploymentFromConfigFile(Recommendation recommendation, Us
                         throw new InvalidOverrideValueException(DeployToolErrorCode.InvalidValueForOptionSettingItem, $"Invalid value {optionSettingValue} for option setting item {optionSettingJsonPath}");
                     }
 
-                    optionSetting.SetValueOverride(settingValue);
+                    optionSetting.SetValueOverride(settingValue, recommendation);
```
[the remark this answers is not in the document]
ashovlin: I believe this was supplanted by `OptionSettingHandler` via https://github.com/aws/aws-dotnet-deploy/pull/548, but let me know otherwise.

src/AWS.Deploy.CLI/Commands/TypeHints/DockerExecutionDirectoryCommand.cs
```diff
@@ -46,16 +46,23 @@ public Task<object> Execute(Recommendation recommendation, OptionSettingItem opt
...
             recommendation.DeploymentBundle.DockerExecutionDirectory = executionDirectory;
         }
 
-        private string ValidateExecutionDirectory(string executionDirectory)
+        /// <summary>
+        /// Validates that the Docker execution directory exists as either an
+        /// absolute path or a path relative to the project directory.
+        /// </summary>
+        /// <param name="executionDirectory">Proposed Docker execution directory</param>
+        /// <param name="recommendation">The selected recommendation settings used for deployment</param>
```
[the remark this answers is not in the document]
ashovlin: Done.

src/AWS.Deploy.CLI/Commands/TypeHints/DockerExecutionDirectoryCommand.cs
```diff
@@ -46,16 +46,23 @@ public Task<object> Execute(Recommendation recommendation, OptionSettingItem opt
...
+        /// <returns>Empty string if the directory is valid, and error message if not</returns>
+        private string ValidateExecutionDirectory(string executionDirectory, Recommendation recommendation)
```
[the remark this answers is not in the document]
ashovlin: It's `string` to match the expected signature for the `validators` param on `AskUserForValue` above, but let me know if you want to rework all of that.

src/AWS.Deploy.CLI/Commands/TypeHints/FilePathCommand.cs
```diff
@@ -0,0 +1,66 @@
...
+        }
+
+        /// <summary>
+        /// This method will be invoked to set a file path setting in the deployment bundle
+        /// when it is specified as part of the user provided configuration file.
+        /// </summary>
+        /// <param name="recommendation">The selected recommendation settings used for deployment <see cref="Recommendation"/></param>
+        /// <param name="filePath">File path path entered by the user</param>
```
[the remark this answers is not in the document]
ashovlin: Fixed.

src/AWS.Deploy.Common/DeploymentBundles/DeploymentBundle.cs
```diff
@@ -18,6 +18,11 @@ public class DeploymentBundle
         /// </summary>
         public string DockerBuildArgs { get; set; } = "";
 
+        /// <summary>
+        /// The path to the Dockerfile. This can either be an absolute path or relative to the project directory.
+        /// </summary>
+        public string DockerfilePath { get; set; } = "";
```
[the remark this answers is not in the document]
ashovlin: Added

src/AWS.Deploy.Common/Recipes/OptionSettingItem.ValueOverride.cs
```diff
@@ -150,7 +152,7 @@ public void SetValueOverride(object valueOverride)
                 {
                     if (deserialized.TryGetValue(childOptionSetting.Id, out var childValueOverride))
                     {
-                        childOptionSetting.SetValueOverride(childValueOverride);
+                        childOptionSetting.SetValueOverride(childValueOverride, recommendation);
                     }
                 }
             }
```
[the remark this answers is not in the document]
ashovlin: I believe this was supplanted by `OptionSettingHandler` via https://github.com/aws/aws-dotnet-deploy/pull/548, but let me know otherwise. 

I still needed to pass `recommendation` into all of the `Validate` calls, which we discussed at the 5/5/22 internal meeting.

src/AWS.Deploy.Common/Recipes/Validation/IOptionSettingItemValidator.cs
ashovlin: Done

src/AWS.Deploy.Common/Recipes/Validation/OptionSettingItemValidatorList.cs
```diff
@@ -16,6 +16,10 @@ public enum OptionSettingItemValidatorList
         /// <summary>
         /// Must be paired with <see cref="RequiredValidator"/>
         /// </summary>
-        Required
+        Required,
+        /// <summary>
+        /// Must be paired with <see cref="FileValidator"/>
+        /// </summary>
+        File
```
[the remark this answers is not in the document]
ashovlin: I ended up switching to `FileExists`, since I like having it suggest what scenario is being validated as opposed to what type of resource it's validating, but let me know if you prefer something else still.

src/AWS.Deploy.Common/Recipes/Validation/OptionSettingItemValidators/FileValidator.cs
ashovlin: Dropped. Early on I was trying to clarify that it's the recommendation that the user has selected from the list, but having worked on this project longer it seems like that's implied most places.

ashovlin: Dropped. Early on I was trying to clarify that it's the recommendation that the user has selected from the list, but having worked on this project longer it seems like that's implied most places.

src/AWS.Deploy.Common/Recipes/Validation/OptionSettingItemValidators/FileValidator.cs
ashovlin: Switched to injection now that it's possible after https://github.com/aws/aws-dotnet-deploy/pull/560 has been merged.

src/AWS.Deploy.Common/Recipes/Validation/OptionSettingItemValidators/FileValidator.cs
ashovlin: Updated the comment above this check. My thinking was that since this is generic, it's possible that not every file option should be required, so we should let users clear the value here. If it is required, that can be added with a separate validator.

src/AWS.Deploy.Common/Recipes/Validation/RecipeValidators/DockerfilePathValidator.cs
ashovlin: Switched to injection now that it's possible after https://github.com/aws/aws-dotnet-deploy/pull/560 has been merged.

src/AWS.Deploy.Common/Recipes/Validation/RecipeValidators/DockerfilePathValidator.cs
ashovlin: We discussed this at the 5/5/22 internal meeting, switched to reading from `recommendation.DeploymentBundle` now that https://github.com/aws/aws-dotnet-deploy/pull/564 has been merged.

src/AWS.Deploy.Recipes/DeploymentBundleDefinitions/Container.deploymentbundle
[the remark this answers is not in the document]
ashovlin: Updated.

src/AWS.Deploy.Common/Recipes/Validation/OptionSettingItemValidators/FileValidator.cs
ashovlin: At the 5/5/22 internal meeting, we had agreed upon passing in `recommendation`  to the `Validate` calls. 

I didn't explore fully, but my hunch is we set up the service collection relatively early compared to when we know the project directory, and generally prefer to avoid global state like this.

src/AWS.Deploy.Common/Recipes/Validation/IOptionSettingItemValidator.cs
```diff
@@ -10,6 +10,12 @@ namespace AWS.Deploy.Common.Recipes.Validation
...
+        /// </summary>
+        /// <param name="input">Raw input for the current OptionSetting</param>
+        /// <param name="recommendation">Current deployment recommendation, may be used if the validator needs to consider properties other than itself</param>
+        /// <returns>Whether or not the input is valid</returns>
+        ValidationResult Validate(object input, Recommendation recommendation);
```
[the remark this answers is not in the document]
ashovlin: Left as-is for now - prefer to require it everywhere since we're going to have to always pass it in anyway from the actual source since we're generally looping over `IOptionSettingItemValidator[]`.

Would rather have a compile-time error rather than a runtime error when a validator asserts that it was passed in. 

Though a lot has shifted in a month and I know you're working on option dependencies, so let me know if you'd still like to see changes.

src/AWS.Deploy.CLI/Commands/TypeHints/FilePathCommand.cs
```diff
@@ -0,0 +1,81 @@
...
+        /// <summary>
+        /// Prompts the user to enter a path to a file
+        /// </summary>
+        public Task<object> Execute(Recommendation recommendation, OptionSettingItem optionSetting)
+        {
+            var userFilePath = _consoleUtilities
+               .AskUserForValue(
+                   string.Empty,
+                   _optionSettingHandler.GetOptionSettingValue<string>(recommendation, optionSetting),
+                   allowEmpty: true,
```
philasmar, line 46: Could you pass in the allow empty from the recipe, using the type hint data. This would make this type hint more generic since some file paths can be empty and some cannot.

src/AWS.Deploy.Common/Recipes/Validation/OptionSettingItemValidators/FileExistsValidator.cs
```diff
@@ -0,0 +1,46 @@
...
+        public FileExistsValidator(IFileManager fileManager)
+        {
+            _fileManager = fileManager;
+        }
+
+        public string ValidationFailedMessage { get; set; } = "The specified file does not exist";
+
+        public Task<ValidationResult> Validate(object input, Recommendation recommendation)
+        {
+            var inputFilePath = input?.ToString() ?? string.Empty;
```
philasmar, line 26: Similar to my other comment, there should be an AllowEmpty attribute which determines if the path can be empty or not.

philasmar, line 26: Similar to my other comment, there should be an AllowEmpty attribute which determines if the path can be empty or not.

src/AWS.Deploy.Orchestration/OptionSettingHandler.cs
```diff
@@ -81,6 +81,9 @@ private void SetDeploymentBundleProperty(Recommendation recommendation, OptionSe
                 case "DockerExecutionDirectory":
                     recommendation.DeploymentBundle.DockerExecutionDirectory = value.ToString() ?? string.Empty;
                     break;
+                case Constants.Docker.DockerfileOptionId:
```
philasmar, line 84: for consistency, could you update the reference of the other deployment bundle Ids to reference a const?

philasmar, line 84: for consistency, could you update the reference of the other deployment bundle Ids to reference a const?

src/AWS.Deploy.CLI/Commands/TypeHints/FilePathCommand.cs
```diff
@@ -0,0 +1,81 @@
...
+                   resetValue: _optionSettingHandler.GetOptionSettingDefaultValue<string>(recommendation, optionSetting) ?? "",
+                   validators: async filePath => await ValidateFileExists(filePath, recommendation));
```
philasmar, line 48: why do we need to specify a validator here if there exists a validator on the option setting defined in the recipe? Aren't we running the validator twice in this case? Once in the console utilities when choosing a value, and another time after the value has been set?

philasmar, line 48: why do we need to specify a validator here if there exists a validator on the option setting defined in the recipe? Aren't we running the validator twice in this case? Once in the console utilities when choosing a value, and another time after the value has been set?

philasmar (changes requested):

src/AWS.Deploy.CLI/Commands/TypeHints/FilePathCommand.cs
ashovlin: I agree with all of that, but I thought you and @normj asked to add this in https://github.com/aws/aws-dotnet-deploy/pull/509#discussion_r855801436 unless I misunderstood that comment.

src/AWS.Deploy.CLI/Commands/TypeHints/FilePathCommand.cs
ashovlin: Done, though I set the default to `true` so didn't specify it explicitly here since we will generate the Dockerfile if it isn't specified.

src/AWS.Deploy.Common/Recipes/Validation/OptionSettingItemValidators/FileExistsValidator.cs
ashovlin: Done via `FileExistsValidator.AllowEmptyString`, though similar to above I defaulted it to `true` so did not specify in the recipe file here.

I also intentionally went with `AllowEmptyString` for consistency with some of the other validators.

src/AWS.Deploy.Orchestration/OptionSettingHandler.cs
ashovlin: Done

src/AWS.Deploy.CLI/Commands/TypeHints/FilePathCommand.cs
ashovlin: Per our discussion yesterday, I did remove the second validation from the type hint in favor of the new option setting validator

image

philasmar (approved):

PR #519, opened 2022-04-22
fix: Amend strategy to find the source control root directory

96malhar:
*Description of changes:*
Currently, we are performing 2 passes of the file-system to determine the source control root of the customer's target application.
Furthermore, incase of nested `git` directories, we are traversing till the very top of the source control chain.

This PR amend this behavior in the following ways: 
* Only performing a single pass of the file-system to determine the source control root.
* Stop the search on the first ancestor that contains a `.git` folder.
* If the source control root is equal to the file-system root, then don't use it to locate custom recipe paths (default to solution directory path)

### DIFF
--- a/src/AWS.Deploy.Orchestration/CustomRecipeLocator.cs
+++ b/src/AWS.Deploy.Orchestration/CustomRecipeLocator.cs
@@ -23,20 +23,14 @@ namespace AWS.Deploy.Orchestration
     /// </summary>
     public class CustomRecipeLocator : ICustomRecipeLocator
     {
-        private const string GIT_STATUS_COMMAND = "git worktree list";
-        private const string SVN_STATUS_COMMAND = "svn status";
-
         private readonly string _ignorePathSubstring = Path.DirectorySeparatorChar + "bin" + Path.DirectorySeparatorChar;
         private readonly IOrchestratorInteractiveService _orchestratorInteractiveService;
-        private readonly ICommandLineWrapper _commandLineWrapper;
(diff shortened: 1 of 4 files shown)


src/AWS.Deploy.Orchestration/CustomRecipeLocator.cs
```diff
@@ -88,26 +88,21 @@ private async Task<List<string>> LocateRecipePathsFromManifestFile(string target
...
-            {
-                rootDirectoryPath = GetSourceControlRootDirectory(targetApplicationDirectoryPath);
-            }
-            else
-            {
+            if (string.IsNullOrEmpty(rootDirectoryPath) || string.Equals(rootDirectoryPath, fileSystemRootPath))
                 rootDirectoryPath = solutionDirectoryPath;
-            }
 
             return GetRecipePathsFromRootDirectory(rootDirectoryPath);
```
normj, line 107: When `GetRecipePathsFromRootDirectory` calls `_directoryManager.GetFiles`. If that `GetFiles` method fails can you report a logging message and then return an empty collection of paths. This would have allowed Bo during his testing to keep working even though he was having permissioning issues.

normj (changes requested):

src/AWS.Deploy.Orchestration/CustomRecipeLocator.cs
96malhar: Fixed

normj (approved):

philasmar (approved):

PR #530, opened 2022-04-26
feat: Rework deployment logging to have section headers

normj:
*Description of changes:*

For the log output during a deployment write section headers along with descriptions that explain what is happening at this stage of deployment.

Here is what the headers look like using the CLI.
LoggingHeaders3

For the VS toolkit there is a new message sent through the SignalR hub that the toolkit can use to get notified when new sections are started. The toolkit can either add the section headers in the output window or create category based UI control.

Notes for PR
* Renamed the logging methods on `IOrchestratorInteractiveService` and the SignalR client `IDeploymentCommunicationHub` to make the names consisted. I have discussed this with the IDE team so and they are fine adapting to the change when the adopt the new version.
* On SignalR client removed the `ReceiveLogMessageLineAction` callback. This also meant the toolkit was displaying debug messages and it didn't make sense to add the new header names in this callback. Toolkit will update to add callback to the Info and Error messages.
* Fix up the deploy to existing beanstalk integration test to make sure the testing stack is deployed to `us-west-2` since the integration test that look for the existing environment was failing on my machine which didn't have `us-west-2` as the default region.

### DIFF
--- a/src/AWS.Deploy.CLI/Commands/DeployCommand.cs
+++ b/src/AWS.Deploy.CLI/Commands/DeployCommand.cs
@@ -121,8 +121,7 @@ namespace AWS.Deploy.CLI.Commands
 
         private void DisplayOutputResources(List<DisplayedResourceItem> displayedResourceItems)
         {
-            _toolInteractiveService.WriteLine("Resources");
-            _toolInteractiveService.WriteLine("---------");
+            _orchestratorInteractiveService.LogSectionStart("AWS Resource Details", null);
             foreach (var resource in displayedResourceItems)
             {
                 _toolInteractiveService.WriteLine($"{resource.Description}:");
@@ -677,6 +676,9 @@ namespace AWS.Deploy.CLI.Commands
         {
             if (selectedRecommendation.Recipe.DeploymentBundle == DeploymentBundleTypes.Container)
             {
+                _orchestratorInteractiveService.LogSectionStart("Creating deployment image",
+                    "Using the docker CLI to perform a docker build to create a container image.");
+
                 while (!await orchestrator.CreateContainerDeploymentBundle(cloudApplication, selectedRecommendation))
                 {
                     if (_toolInteractiveService.DisableInteractive)
@@ -710,6 +712,9 @@ namespace AWS.Deploy.CLI.Commands
             }
             else if (selectedRecommendation.Recipe.DeploymentBundle == DeploymentBundleTypes.DotnetPublishZipFile)
             {
+                _orchestratorInteractiveService.LogSectionStart("Creating deployment zip bundle",
+                    "Using the dotnet CLI build the project and zip the publish artifacts.");
+
                 var dotnetPublishDeploymentBundleResult = await orchestrator.CreateDotnetPublishDeploymentBundle(selectedRecommendation);
                 if (!dotnetPublishDeploymentBundleResult)
                     throw new FailedToCreateDeploymentBundleException(DeployToolErrorCode.FailedToCreateDotnetPublishDeploymentBundle, "Failed to create a deployment bundle");
--- a/src/AWS.Deploy.CLI/ConsoleOrchestratorLogger.cs
+++ b/src/AWS.Deploy.CLI/ConsoleOrchestratorLogger.cs
@@ -14,17 +14,32 @@ namespace AWS.Deploy.CLI
             _interactiveService = interactiveService;
         }
 
-        public void LogErrorMessageLine(string? message)
+        public void LogSectionStart(string message, string? description)
+        {
+            var sectionBreak = new string('*', message.Length);
+            _interactiveService.WriteLine(string.Empty);
+            _interactiveService.WriteLine(sectionBreak);
+            _interactiveService.WriteLine(message);
+            if(description != null)
+            {
+                _interactiveService.WriteLine(new string('-', message.Length));
+                _interactiveService.WriteLine(description);
+            }
+            _interactiveService.WriteLine(sectionBreak);
+            _interactiveService.WriteLine(string.Empty);
+        }
+
+        public void LogErrorMessage(string? message)
         {
             _interactiveService.WriteErrorLine(message);
         }
 
-        public void LogMessageLine(string? message)
+        public void LogInfoMessage(string? message)
         {
             _interactiveService.WriteLine(message);
         }
 
-        public void LogDebugLine(string? message)
+        public void LogDebugMessage(string? message)
         {
             _interactiveService.WriteDebugLine(message);
         }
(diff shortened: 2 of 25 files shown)


src/AWS.Deploy.CLI/Commands/DeployCommand.cs
```diff
@@ -677,6 +676,9 @@ private async Task CreateDeploymentBundle(Orchestrator orchestrator, Recommendat
         {
             if (selectedRecommendation.Recipe.DeploymentBundle == DeploymentBundleTypes.Container)
             {
+                _orchestratorInteractiveService.LogSectionStart("Creating deployment image",
+                    "Using the docker CLI perform a docker build to create a container image.");
```
ashovlin, line 680: nit: Grammar here took me a few reads to parse this sentence
* "Using the docker CLI **to** perform...
* Using the docker CLI, perform a... (add the comma?)

(same comment down on line 716)

test/AWS.Deploy.CLI.UnitTests/Utilities/TestToolOrchestratorInteractiveService.cs
```diff
@@ -8,12 +8,14 @@ namespace AWS.Deploy.CLI.UnitTests.Utilities
...
     {
+        public IList<string> SectionStartMessages { get; } = new List<string>();
         public IList<string> DebugMessages { get; } = new List<string>();
         public IList<string> OutputMessages { get; } = new List<string>();
         public IList<string> ErrorMessages { get; } = new List<string>();
 
-        public void LogDebugLine(string message) => DebugMessages.Add(message);
-        public void LogErrorMessageLine(string message) => ErrorMessages.Add(message);
-        public void LogMessageLine(string message) => OutputMessages.Add(message);
+        public void LogSectionStart(string message, string? description) => SectionStartMessages.Add(message);
```
ashovlin, line 16: Is the warning on GitHub here addressable - `The annotation for nullable reference types should only be used in code within a '#nullable' annotations context.`?

If so, think it'd be nice to keep build warnings clean, especially if this will appear on all future PRs.

src/AWS.Deploy.Orchestration/Orchestrator.cs
```diff
@@ -209,14 +209,16 @@ public async Task<bool> CreateContainerDeploymentBundle(CloudApplication cloudAp
                 }
 
                 await _deploymentBundleHandler.BuildDockerImage(cloudApplication, recommendation, imageTag);
+
+                _interactiveService.LogSectionStart("Pushing container image to Elastic Container Registry (ECR)", "Using the docker CLI log on to ECR and push the local image to ECR.");
```
ashovlin, line 213: Same nit as on `DeployCommand` on adding either "to" or a comma after "CLI"

src/AWS.Deploy.Orchestration/DeploymentCommands/BeanstalkEnvironmentDeploymentCommand.cs
```diff
@@ -37,19 +37,21 @@ public async Task ExecuteAsync(Orchestrator orchestrator, CloudApplication cloud
                 throw new AWSResourceNotFoundException(DeployToolErrorCode.FailedToFindElasticBeanstalkApplication, message);
             }
 
-            orchestrator._interactiveService.LogMessageLine($"Initiating deployment to {cloudApplication.DisplayName}...");
+            orchestrator._interactiveService.LogSectionStart($"Create application version", "Uploading deployment bundle to S3 and create an Elastic Beanstalk application version");
```
ashovlin, line 40: nit: Most other `LogSectionStart` calls are using `-ing` on the verb, but here it's just "Create" and "Deploy".

I think it'd be nice polish if we were consistent throughout.

normj: > Current behavior will stream the headers as they come in so in terms of UI, the toolkit will add new sections adhoc which wouldn't give the user a sense of what the overall process is going to be like. This doesn't have to be part of this PR, but it's worth thinking of a way to outline major sections like Configuration, Deployment Bundle, Deployment (things that we know will happen for sure regardless of recipe) and the rest of the recipe dependent headers will be streamed in.

Yes, Chris and I have discussed this. But since our system is so dynamic we don't currently have a way in our system to precompute all of the stages. That can be something we possibly do in the future.

src/AWS.Deploy.CLI/ConsoleOrchestratorLogger.cs
```diff
@@ -14,17 +14,31 @@ public ConsoleOrchestratorLogger(IToolInteractiveService interactiveService)
...
+            var sectionBreak = new string('*', message.Length);
+            _interactiveService.WriteLine(string.Empty);
+            _interactiveService.WriteLine(sectionBreak);
+            _interactiveService.WriteLine(message);
+            if(description != null)
+            {
+                _interactiveService.WriteLine(new string('-', message.Length));
+                _interactiveService.WriteLine(description);
+            }
+            _interactiveService.WriteLine(sectionBreak);
```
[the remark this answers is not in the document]
normj: I went back and forth on this but I added a newline. Curious what others will think.

normj:
> Current behavior will stream the headers as they come in so in terms of UI, the toolkit will add new sections adhoc which wouldn't give the user a sense of what the overall process is going to be like. This doesn't have to be part of this PR, but it's worth thinking of a way to outline major sections like Configuration, Deployment Bundle, Deployment (things that we know will happen for sure regardless of recipe) and the rest of the recipe dependent headers will be streamed in.

Yes, Chris and I have discussed this. But since our system is so dynamic we don't currently have a way in our system to precompute all of the stages. That can be something we possibly do in the future.

test/AWS.Deploy.CLI.UnitTests/Utilities/TestToolOrchestratorInteractiveService.cs
normj: Removed. Didn't realize the test project didn't have nullablity turned on.

src/AWS.Deploy.Orchestration/Orchestrator.cs
normj: Added `to`

src/AWS.Deploy.CLI/Commands/DeployCommand.cs
ashovlin: It looks like `Orchestrator` was updated in the force-push but not `DeployCommand`

ashovlin (approved):

philasmar (approved):

PR #537, opened 2022-04-28
fix: Consolidate implementations of CreateDeploymentBundle

normj:
*Description of changes:*
The CLI and ServerMode had separate `CreateDeploymentBundle` method implementations. This caused issues with the recent new change to add new logging section headers because only the CLI version had the section header.

This PR pushes the `CreateDeploymentBundle` into the Orchestrator so both the CLI and ServerMode can use the same implementation. The CLI implementation did have the interactive part to ask the user if they want to change the Docker execution directory if the build failed. I kept that logic in the CLI but it only prompts if `CreateDeploymentBundle` in the Orchestrator throws an exception with the container build error code.

### DIFF
--- a/src/AWS.Deploy.CLI/Commands/DeployCommand.cs
+++ b/src/AWS.Deploy.CLI/Commands/DeployCommand.cs
@@ -674,59 +674,41 @@ namespace AWS.Deploy.CLI.Commands
 
         private async Task CreateDeploymentBundle(Orchestrator orchestrator, Recommendation selectedRecommendation, CloudApplication cloudApplication)
         {
-            if (selectedRecommendation.Recipe.DeploymentBundle == DeploymentBundleTypes.Container)
+            try
             {
-                _orchestratorInteractiveService.LogSectionStart("Creating deployment image",
-                    "Using the docker CLI to perform a docker build to create a container image.");
-
-                while (!await orchestrator.CreateContainerDeploymentBundle(cloudApplication, selectedRecommendation))
+                await orchestrator.CreateDeploymentBundle(cloudApplication, selectedRecommendation);
+            }
+            catch(FailedToCreateDeploymentBundleException ex) when (ex.ErrorCode == DeployToolErrorCode.FailedToCreateContainerDeploymentBundle)
+            {
+                if (_toolInteractiveService.DisableInteractive)
                 {
-                    if (_toolInteractiveService.DisableInteractive)
-                    {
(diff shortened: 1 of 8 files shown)


src/AWS.Deploy.CLI/Commands/DeployCommand.cs
```diff
@@ -674,51 +674,43 @@ private bool ConfirmDeployment(Recommendation recommendation)
...
                             selectedRecommendation.DeploymentBundle.DockerExecutionDirectory,
                             allowEmpty: true);
 
-                        if (!_directoryManager.Exists(dockerExecutionDirectory))
-                            continue;
+                        if (_directoryManager.Exists(dockerExecutionDirectory))
+                        {
+                            _toolInteractiveService.WriteLine($"Error, directory does not exist \"{dockerExecutionDirectory}\"");
+                            break;
+                        }
```
96malhar, line 707: I believe this should be the following:
```suggestion
                        if (!_directoryManager.Exists(dockerExecutionDirectory))
                        {
                            _toolInteractiveService.WriteErrorLine($"Error, directory does not exist \"{dockerExecutionDirectory}\"");
                        }

```

normj: Done

src/AWS.Deploy.CLI/Commands/DeployCommand.cs
```diff
@@ -674,51 +674,43 @@ private bool ConfirmDeployment(Recommendation recommendation)
...
                             allowEmpty: true);
 
-                        if (!_directoryManager.Exists(dockerExecutionDirectory))
-                            continue;
+                        if (_directoryManager.Exists(dockerExecutionDirectory))
+                        {
+                            _toolInteractiveService.WriteErrorLine($"Error, directory does not exist \"{dockerExecutionDirectory}\"");
+                            break;
+                        }
+                    } while (!_directoryManager.Exists(dockerExecutionDirectory));
```
96malhar, line 708: You are missing the "not" operator in the if condition. The break statement also shouldn't be there.

It should be:
```
if (!_directoryManager.Exists(dockerExecutionDirectory))
                        {
                            _toolInteractiveService.WriteErrorLine($"Error, directory does not exist \"{dockerExecutionDirectory}\"");
                        }
```

normj: Done again

96malhar (approved):

philasmar (approved):

PR #538, opened 2022-04-29
fix: Fixed issue with Beanstalk Application and Environment name

normj:
*Description of changes:*
The Beanstalk recipe had 2 issues with the settings.

* **Application Name** - This is a object setting with a **CreateNew** child setting. In the original version the **ApplicationName** setting was incorrectly being used for either new application names or existing application names. This has been fixed to have a separate **ExistingApplicationName** for existing application names. The **ExistingApplicationName** has the type hint telling and the **ApplicationName** is just a pure text entry. 
* **Environment Name** - This object setting incorrectly had a **CreateNew** setting that would allow selecting an existing Beanstalk environment name. The CDK project doesn't support using a Beanstalk environment that doesn't belong to the stack.


96malhar (changes requested):
Need to amend the [config file](https://github.com/aws/aws-dotnet-deploy/blob/main/testapps/WebAppNoDockerFile/ElasticBeanStalkConfigFile.json) to fix the integ test

philasmar (approved):

96malhar (approved):

PR #540, opened 2022-05-02
fix: Add new error code to indicate deployment bundle creation failure from a generate docker file

96malhar:
*Description of changes:*
This PR adds a new error code to distinguish `FailedToCreateDeploymentBundleException` when using a generated Docker file.

### DIFF
--- a/src/AWS.Deploy.CLI/Commands/DeployCommand.cs
+++ b/src/AWS.Deploy.CLI/Commands/DeployCommand.cs
@@ -706,6 +706,15 @@ namespace AWS.Deploy.CLI.Commands
                     }
                     else
                     {
+                        if (!selectedRecommendation.ProjectDefinition.HasDockerFile)
+                        {
+                            var projectDirectory = _directoryManager.GetDirectoryInfo(selectedRecommendation.ProjectPath).Parent.FullName;
+                            var dockerfilePath = Path.Combine(projectDirectory, "Dockerfile");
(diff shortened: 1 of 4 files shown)


src/AWS.Deploy.CLI/Commands/DeployCommand.cs
```diff
@@ -706,6 +706,10 @@ private async Task CreateDeploymentBundle(Orchestrator orchestrator, Recommendat
                     }
                     else
                     {
+                        if (!selectedRecommendation.ProjectDefinition.HasDockerFile)
+                        {
+                            throw new FailedToCreateDeploymentBundleException(DeployToolErrorCode.FailedToCreateContainerDeploymentBundleFromGeneratedDockerFile, "Failed to create a deployment bundle from generated Docker file");
```
normj, line 711: For error message what about something like to give the user some direction on what to do to fix the issue.
```
Failed to create a container image from generated Docker file. Please edit the Dockerfile at {DockerfilePath} to correct the required build steps for the project. Common errors are missing project dependencies not included in the Dockerfile.
```

normj (changes requested):

src/AWS.Deploy.CLI/Commands/DeployCommand.cs
96malhar: Added the error message.

normj (approved):

philasmar (approved):

PR #541, opened 2022-05-02
feat: Skip executing "cdk deploy" if CDKToolkit stack already exists

normj:
*Description of changes:*
Before executing "cdk bootstrap" command to make create the CDK bootstrap CloudFormation stack check to see if the stack already exists. If it does exist and its version checked in parameter store is at least the version that the deploy uses then skip calling out to the CDK.

### DIFF
--- a/src/AWS.Deploy.Orchestration/AWS.Deploy.Orchestration.csproj
+++ b/src/AWS.Deploy.Orchestration/AWS.Deploy.Orchestration.csproj
@@ -22,6 +22,7 @@
     <PackageReference Include="AWSSDK.CloudFormation" Version="3.7.7.14" />
     <PackageReference Include="AWSSDK.S3" Version="3.7.1.17" />
     <PackageReference Include="AWSSDK.AppRunner" Version="3.7.3.11" />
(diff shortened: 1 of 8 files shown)


philasmar (approved):

src/AWS.Deploy.Orchestration/AWS.Deploy.Orchestration.csproj
```diff
@@ -22,6 +22,7 @@
     <PackageReference Include="AWSSDK.CloudFormation" Version="3.7.7.14" />
     <PackageReference Include="AWSSDK.S3" Version="3.7.1.17" />
     <PackageReference Include="AWSSDK.AppRunner" Version="3.7.3.11" />
+    <PackageReference Include="AWSSDK.SimpleSystemsManagement" Version="3.7.16" />
```
ashovlin, line 25: Could you update our internal license tooling and the THIRD_PARTY_NOTICES file with the new dependency?

normj: Done

96malhar (approved):

PR #545, opened 2022-05-04
chore: improve error handling for failed first time deployments

philasmar:
*Issue #, if available:*
DOTNET-5899
DOTNET-5896

*Description of changes:*
* Added missing CloudFormation event statuses to capture all failed events
* Added ability to retrieve unique stack id to retrieve stack events even after deployment finishes (whether the stack is deleted or not)
* Added metric to check deployment time which will help in analysis of errors in telemetry

### DIFF
--- a/src/AWS.Deploy.Orchestration/CdkProjectHandler.cs
+++ b/src/AWS.Deploy.Orchestration/CdkProjectHandler.cs
@@ -3,6 +3,7 @@ using System.Collections;
 using System.Collections.Generic;
 using System.IO;
 using System.Linq;
+using System.Threading;
 using System.Threading.Tasks;
 using Amazon.CloudFormation;
 using AWS.Deploy.Common;
@@ -12,6 +13,7 @@ using AWS.Deploy.Orchestration.CDK;
 using AWS.Deploy.Orchestration.Data;
 using AWS.Deploy.Orchestration.Utilities;
 using AWS.Deploy.Recipes.CDK.Common;
+using Stack = Amazon.CloudFormation.Model.Stack;
 
 namespace AWS.Deploy.Orchestration
 {
@@ -124,20 +126,48 @@ namespace AWS.Deploy.Orchestration
             _interactiveService.LogSectionStart("Deploying AWS CDK project",
                 "Use the CDK project to create or update the AWS CloudFormation stack and deploy the project to the AWS resources in the stack.");
 
-            var deploymentStartDate = DateTime.Now;
             // Handover to CDK command line tool
             // Use a CDK Context parameter to specify the settings file that has been serialized.
-            var cdkDeploy = await _commandLineWrapper.TryRunWithResult($"npx cdk deploy --require-approval never -c {Constants.CloudFormationIdentifier.SETTINGS_PATH_CDK_CONTEXT_PARAMETER}=\"{appSettingsFilePath}\"",
+            var cdkDeployTask = _commandLineWrapper.TryRunWithResult( $"npx cdk deploy --require-approval never -c {Constants.CloudFormationIdentifier.SETTINGS_PATH_CDK_CONTEXT_PARAMETER}=\"{appSettingsFilePath}\"",
                 workingDirectory: cdkProjectPath,
                 environmentVariables: environmentVariables,
                 needAwsCredentials: true,
                 redirectIO: true,
                 streamOutputToInteractiveService: true);
 
-            await CheckCdkDeploymentFailure(cloudApplication, deploymentStartDate);
+            var cancellationTokenSource = new CancellationTokenSource();
+            var retrieveStackIdTask = RetrieveStackId(cloudApplication, cancellationTokenSource.Token);
 
+            var deploymentStartDate = DateTime.UtcNow;
+            var firstCompletedTask = await Task.WhenAny(cdkDeployTask, retrieveStackIdTask);
+            // Deployment end date is captured at this point after 1 of the 2 running tasks yields.
+            var deploymentEndDate = DateTime.UtcNow;
+
+            TryRunResult? cdkDeploy = null;
+            if (firstCompletedTask == retrieveStackIdTask)
+            {
+                // If retrieveStackIdTask completes first, that means a stack was created and exists in CloudFormation.
+                // We can proceed with checking for deployment failures.
+                var stackId = cloudApplication.Name;
+                if (!retrieveStackIdTask.IsFaulted)
+                    stackId = retrieveStackIdTask.Result;
+                cdkDeploy = await cdkDeployTask;
+                // We recapture the deployment end date at this point after the deployment task completes.
+                deploymentEndDate = DateTime.UtcNow;
+                await CheckCdkDeploymentFailure(stackId, deploymentStartDate, deploymentEndDate);
+            }
+            else
+            {
+                // If cdkDeployTask completes first, that means 'cdk deploy' failed before creating a stack in CloudFormation.
+                // In this case, we skip checking for deployment failures since a stack does not exist.
+
+                cdkDeploy = cdkDeployTask.Result;
+                cancellationTokenSource.Cancel();
+            }
+
+            var deploymentTotalTime = Math.Round((deploymentEndDate - deploymentStartDate).TotalSeconds, 2);
             if (cdkDeploy.ExitCode != 0)
-                throw new FailedToDeployCDKAppException(DeployToolErrorCode.FailedToDeployCdkApplication, "We had an issue deploying your application to AWS. Check the deployment output for more details.");
+                throw new FailedToDeployCDKAppException(DeployToolErrorCode.FailedToDeployCdkApplication, $"We had an issue deploying your application to AWS. Check the deployment output for more details. Deployment took {deploymentTotalTime}s.");
         }
 
         public async Task<bool> DetermineIfCDKBootstrapShouldRun()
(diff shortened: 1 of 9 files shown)


src/AWS.Deploy.Orchestration/CdkProjectHandler.cs
```diff
@@ -124,20 +125,47 @@ public async Task DeployCdkProject(OrchestratorSession session, CloudApplication
...
                 environmentVariables: environmentVariables,
                 needAwsCredentials: true,
                 redirectIO: true,
                 streamOutputToInteractiveService: true);
 
-            await CheckCdkDeploymentFailure(cloudApplication, deploymentStartDate);
+            var cancellationTokenSource = new CancellationTokenSource();
+            var retrieveStackIdTask = RetreiveStackId(cloudApplication, cancellationTokenSource.Token);
+
+            var deploymentStartDate = DateTime.Now;
```
normj, line 140: To enforce good habits lets use `UtcNow`.

src/AWS.Deploy.Orchestration/CdkProjectHandler.cs
```diff
@@ -124,20 +125,47 @@ public async Task DeployCdkProject(OrchestratorSession session, CloudApplication
...
+            var firstCompletedTask = await Task.WhenAny(cdkDeployTask, retrieveStackIdTask);
+            // Deployment end date is captured at this point after 1 of the 2 running tasks yields.
+            var deploymentEndDate = DateTime.Now;
```
normj, line 143: `UtcNow`

src/AWS.Deploy.Orchestration/CdkProjectHandler.cs
```diff
@@ -168,27 +196,55 @@ public async Task<bool> DetermineIfCDKBootstrapShouldRun()
             return false;
         }
 
-        private async Task CheckCdkDeploymentFailure(CloudApplication cloudApplication, DateTime deploymentStartDate)
+        private async Task<string> RetreiveStackId(CloudApplication cloudApplication, CancellationToken cancellationToken)
+        {
+            // Do an initial delay to avoid a race condition of the status being checked before the deployment has kicked off.
+            await Task.Delay(TimeSpan.FromSeconds(3));
```
normj, line 202: This isn't safe logic. If the CDK project involves uploading lots of data to S3 for example like the deployment bundle for Beanstalk or S3 files for Blazor this could take an undetermined amount of time.

src/AWS.Deploy.Orchestration/CdkProjectHandler.cs
```diff
@@ -168,27 +196,55 @@ public async Task<bool> DetermineIfCDKBootstrapShouldRun()
...
+                try
+                {
+                    var stack = await _awsResourceQueryer.GetCloudFormationStack(cloudApplication.Name);
+                    return stack != null;
+                }
+                catch (ResourceQueryException exception) when (exception.InnerException != null && exception.InnerException.Message.Equals($"Stack with id {cloudApplication.Name} does not exist"))
+                {
+                    return false;
+                }
+            }, TimeSpan.FromSeconds(1), TimeSpan.FromMinutes(5), cancellationToken);
```
normj, line 215: To reduce throttling errors between us and the CDK ping 3 seconds instead of 1.

src/AWS.Deploy.Orchestration/CdkProjectHandler.cs
```diff
@@ -168,27 +196,55 @@ public async Task<bool> DetermineIfCDKBootstrapShouldRun()
...
+        private async Task<string> RetreiveStackId(CloudApplication cloudApplication, CancellationToken cancellationToken)
+        {
+            // Do an initial delay to avoid a race condition of the status being checked before the deployment has kicked off.
+            await Task.Delay(TimeSpan.FromSeconds(3));
+
+            await WaitUntilHelper.WaitUntil(async () =>
+            {
+                try
+                {
+                    var stack = await _awsResourceQueryer.GetCloudFormationStack(cloudApplication.Name);
```
normj, line 208: Declare the `stack` variable outside if the `WaitUtil`. Then you don't have to an extra call to `GetCloudFormationStack` after the wait call.

normj (changes requested):

src/AWS.Deploy.Orchestration/CdkProjectHandler.cs
philasmar: Done

src/AWS.Deploy.Orchestration/CdkProjectHandler.cs
philasmar: Removed

src/AWS.Deploy.Orchestration/CdkProjectHandler.cs
philasmar: Done

src/AWS.Deploy.Orchestration/CdkProjectHandler.cs
philasmar: Done

src/AWS.Deploy.Orchestration/CdkProjectHandler.cs
```diff
@@ -124,20 +126,47 @@ public async Task DeployCdkProject(OrchestratorSession session, CloudApplication
...
+            // Deployment end date is captured at this point after 1 of the 2 running tasks yields.
+            var deploymentEndDate = DateTime.UtcNow;
+
+            TryRunResult? cdkDeploy = null;
+            if (firstCompletedTask == retrieveStackIdTask)
+            {
+                // If retrieveStackIdTask completes first, that means a stack was created and exists in CloudFormation.
+                // We can proceed with checking for deployment failures.
+
+                var stackId = retrieveStackIdTask.Result;
```
normj, line 152: If the `RetreiveStackId` method throws an unhandled exception the `.Result` call here will throw an exception which I think will skip all of the error error checking and logging here. For example I assume we still want to get the CDK exit code. In particular if we trigger a CloudFormation throttling error or a transient error on the SDK calls I don't want this to trigger a deployment failure.

src/AWS.Deploy.Orchestration/CdkProjectHandler.cs
philasmar: I reverted to Now because we are filtering the CloufFormation stack events based on the timestamp which causes a mismatch.

src/AWS.Deploy.Orchestration/CdkProjectHandler.cs
philasmar: I reverted to Now because we are filtering the CloufFormation stack events based on the timestamp which causes a mismatch.

normj (approved):

ashovlin (changes requested):

src/AWS.Deploy.Orchestration/CdkProjectHandler.cs
philasmar: The CloudFormation SDK is still returning in local time, which is why I switched back to local time.

src/AWS.Deploy.Orchestration/CdkProjectHandler.cs
```diff
@@ -124,20 +126,48 @@ public async Task DeployCdkProject(OrchestratorSession session, CloudApplication
...
+            {
+                // If cdkDeployTask completes first, that means 'cdk deploy' failed before creating a stack in CloudFormation.
+                // In this case, we skip checking for deployment failures since a stack does not exist.
+
+                cdkDeploy = cdkDeployTask.Result;
+                cancellationTokenSource.Cancel();
+            }
 
+            var deploymentTotalTime = Math.Round((deploymentEndDate - deploymentStartDate).TotalSeconds, 2);
             if (cdkDeploy.ExitCode != 0)
```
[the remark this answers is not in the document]
philasmar: We are blocking the toolkit release until we merge this PR. Even if we add the exit code, the toolkit still needs to implement that. So it wouldn't make a lot of sense for us to tackle it here.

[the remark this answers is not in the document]
philasmar: We are blocking the toolkit release until we merge this PR. Even if we add the exit code, the toolkit still needs to implement that. So it wouldn't make a lot of sense for us to tackle it here.

src/AWS.Deploy.Orchestration/CdkProjectHandler.cs
```diff
@@ -168,27 +198,52 @@ public async Task<bool> DetermineIfCDKBootstrapShouldRun()
             return false;
         }
 
-        private async Task CheckCdkDeploymentFailure(CloudApplication cloudApplication, DateTime deploymentStartDate)
+        private async Task<string> RetreiveStackId(CloudApplication cloudApplication, CancellationToken cancellationToken)
```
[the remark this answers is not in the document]
philasmar: fixed

src/AWS.Deploy.Orchestration/CdkProjectHandler.cs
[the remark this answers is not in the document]
ashovlin: Got it, didn't realize it would require work on their side. Okay pushing until later.

src/AWS.Deploy.Orchestration/CdkProjectHandler.cs
ashovlin: Thanks for moving to UTC with the `.ToUniversalTime`

ashovlin (approved):

PR #547, opened 2022-05-05
fix: Add stack name validation in server mode

96malhar:
*Issue #, if available:*
DOTNET-5895

*Description of changes:*
This PR adds the stack name validation logic to server mode.

### DIFF
--- a/src/AWS.Deploy.CLI/ServerMode/Controllers/DeploymentController.cs
+++ b/src/AWS.Deploy.CLI/ServerMode/Controllers/DeploymentController.cs
@@ -358,21 +358,33 @@ namespace AWS.Deploy.CLI.ServerMode.Controllers
 
             var serviceProvider = CreateSessionServiceProvider(state);
             var orchestrator = CreateOrchestrator(state, serviceProvider);
+            var cloudApplicationNameGenerator = serviceProvider.GetRequiredService<ICloudApplicationNameGenerator>();
 
-            if(!string.IsNullOrEmpty(input.NewDeploymentRecipeId) &&
-               !string.IsNullOrEmpty(input.NewDeploymentName))
+            if (!string.IsNullOrEmpty(input.NewDeploymentRecipeId))
             {
+                var newDeploymentName = input.NewDeploymentName ?? string.Empty;
+
                 state.SelectedRecommendation = state.NewRecommendations?.FirstOrDefault(x => string.Equals(input.NewDeploymentRecipeId, x.Recipe.Id));
-                if(state.SelectedRecommendation == null)
+                if (state.SelectedRecommendation == null)
                 {
                     return NotFound($"Recommendation {input.NewDeploymentRecipeId} not found.");
                 }
 
-                state.ApplicationDetails.Name = input.NewDeploymentName;
+                if (state.SelectedRecommendation.Recipe.DeploymentType != Common.Recipes.DeploymentTypes.ElasticContainerRegistryImage)
+                {
+                    // We only validate the name when the recipe deployment type is not ElasticContainerRegistryImage.
+                    // This is because pushing images to ECR does not need a cloud application name.
+                    if (!cloudApplicationNameGenerator.IsValidName(newDeploymentName))
+                    {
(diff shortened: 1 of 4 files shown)


src/AWS.Deploy.CLI/ServerMode/Controllers/DeploymentController.cs
```diff
@@ -358,15 +358,19 @@ public async Task<IActionResult> SetDeploymentTarget(string sessionId, [FromBody
 
             var serviceProvider = CreateSessionServiceProvider(state);
             var orchestrator = CreateOrchestrator(state, serviceProvider);
+            var cloudApplicationNameGenerator = serviceProvider.GetRequiredService<ICloudApplicationNameGenerator>();
 
-            if(!string.IsNullOrEmpty(input.NewDeploymentRecipeId) &&
-               !string.IsNullOrEmpty(input.NewDeploymentName))
+            if (!string.IsNullOrEmpty(input.NewDeploymentRecipeId) && !string.IsNullOrEmpty(input.NewDeploymentName))
```
normj, line 363: What do we do in the push image case where name is not required? I worry we are getting lucky that the toolkit is accidently sending us a name that we ignore.

normj (changes requested):
We need test coverage to ensure we don't lose this validation in the future.

src/AWS.Deploy.CLI/ServerMode/Controllers/DeploymentController.cs
96malhar: Added logic to exclude the stack name validation in the push image case

normj (approved):

src/AWS.Deploy.CLI/ServerMode/Controllers/DeploymentController.cs
```diff
@@ -358,21 +358,33 @@ public async Task<IActionResult> SetDeploymentTarget(string sessionId, [FromBody
...
+                if (state.SelectedRecommendation == null)
                 {
                     return NotFound($"Recommendation {input.NewDeploymentRecipeId} not found.");
                 }
 
-                state.ApplicationDetails.Name = input.NewDeploymentName;
+                if (state.SelectedRecommendation.Recipe.DeploymentType != Common.Recipes.DeploymentTypes.ElasticContainerRegistryImage)
+                {
+                    // We only validate the name when the recipe deployment type is not ElasticContainerRegistryImage.
+                    // This is because pushing images to ECR does not need a cloud application name.
```
ashovlin, line 376: Doesn't `ApplyAllReplacementTokens` reuse the `cloudApplicationName` as the ECR Repository Name?

Is it possible by skipping the validation here, we're setting an invalid ECR Repository Name?

96malhar: ECR Repository names have their own [validators](https://github.com/aws/aws-dotnet-deploy/blob/main/src/AWS.Deploy.Recipes/DeploymentBundleDefinitions/Container.deploymentbundle#L33). Even if an invalid value is set, it will be flagged by the option setting validator.

ashovlin: Is that being run for server mode? In the current public build of the VS Toolkit, I am able to initiate a deployment with an invalid ECR Repository Name that I set via the _"Application name"_ field, that results in a deployment failure:

image

96malhar: I cannot set any application name while selecting the "Push Container Images to Amazon Elastic Container Registry" deployment recipe. That field is blocked for me.

ashovlin: We discussed internally that we are running the option validator in this workflow, but the regex it's validating does not match what the service is actually enforcing.

Submitted this separately as https://github.com/aws/aws-dotnet-deploy/issues/549, but okay proceeding here.

philasmar (approved):

PR #548, opened 2022-05-09
chore: add a service to handle Recommendations and Option Setting items

philasmar:
*Description of changes:*
Added a new service to handle setting and getting values from Option Setting Items as well as remove any unnecessary functionality from the Recommendation and OptionSettingItem classes into the Service. This allows for added extensibility for upcoming work.

### DIFF
--- a/src/AWS.Deploy.CLI/Commands/TypeHints/TypeHintCommandFactory.cs
+++ b/src/AWS.Deploy.CLI/Commands/TypeHints/TypeHintCommandFactory.cs
@@ -9,6 +9,7 @@ using AWS.Deploy.Common.IO;
 using AWS.Deploy.Common.Recipes;
 using AWS.Deploy.Common.TypeHintData;
 using AWS.Deploy.Orchestration.Data;
+using Microsoft.Extensions.DependencyInjection;
 
 namespace AWS.Deploy.CLI.Commands.TypeHints
 {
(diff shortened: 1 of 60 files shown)


src/AWS.Deploy.CLI/Commands/TypeHints/TypeHintCommandFactory.cs
```diff
@@ -33,37 +33,37 @@ public class TypeHintCommandFactory : ITypeHintCommandFactory
...
             _commands = new Dictionary<OptionSettingTypeHint, ITypeHintCommand>
             {
-                { OptionSettingTypeHint.BeanstalkApplication, new BeanstalkApplicationCommand(awsResourceQueryer, consoleUtilities) },
-                { OptionSettingTypeHint.ExistingBeanstalkApplication, new BeanstalkApplicationCommand(awsResourceQueryer, consoleUtilities) },
-                { OptionSettingTypeHint.BeanstalkEnvironment, new BeanstalkEnvironmentCommand(awsResourceQueryer, consoleUtilities) },
-                { OptionSettingTypeHint.DotnetBeanstalkPlatformArn, new DotnetBeanstalkPlatformArnCommand(awsResourceQueryer, consoleUtilities) },
-                { OptionSettingTypeHint.EC2KeyPair, new EC2KeyPairCommand(toolInteractiveService, awsResourceQueryer, consoleUtilities) },
-                { OptionSettingTypeHint.IAMRole, new IAMRoleCommand(awsResourceQueryer, consoleUtilities) },
-                { OptionSettingTypeHint.ExistingIAMRole, new IAMRoleCommand(awsResourceQueryer, consoleUtilities) },
+                { OptionSettingTypeHint.BeanstalkApplication, new BeanstalkApplicationCommand(awsResourceQueryer, consoleUtilities, optionSettingHandler) },
```
normj, line 40: Use `ActivatorUtilities.CreateInstance` here again. Which probably means all `TypeHintCommandFactory` needs to take in is the `IServiceProvider`.

normj (changes requested):
I had originally thought this feature was going to allow a different type of IOptionSettingHandler per setting type (data type/ type hint). In Alex's case which wants to add a setting with a filepath type hint he could create a subclass of the default option setting handler that would do file path validation and conversion to relative path.

Instead we are doing a single instance of `IOptionSettingHandler` for the process. So I assume that means the `SetOptionSettingValue` method will grow for each special validation we need to do. Or am I missing something.

src/AWS.Deploy.CLI/Commands/TypeHints/TypeHintCommandFactory.cs
philasmar: Done

philasmar:
> I had originally thought this feature was going to allow a different type of IOptionSettingHandler per setting type (data type/ type hint). In Alex's case which wants to add a setting with a filepath type hint he could create a subclass of the default option setting handler that would do file path validation and conversion to relative path.
> 
> Instead we are doing a single instance of `IOptionSettingHandler` for the process. So I assume that means the `SetOptionSettingValue` method will grow for each special validation we need to do. Or am I missing something. If I'm not missing something is the trade off we want to make?

The follow up work that I am doing will add a new field to the recipe to define which values are not allowed.

For this work, you are correct. `SetOptionSettingValue` method will grow for each special validation we need to do. It will handle the additional services needed as part of my follow up work and Alex's work.

src/AWS.Deploy.Common/Recipes/OptionSettingItem.ValueOverride.cs
```diff
@@ -150,7 +148,7 @@ public void SetValueOverride(object valueOverride)
                 {
                     if (deserialized.TryGetValue(childOptionSetting.Id, out var childValueOverride))
                     {
-                        childOptionSetting.SetValueOverride(childValueOverride);
+                        childOptionSetting.SetValue(childValueOverride);
```
normj, line 151: Something to think about. Having the set call calling directly to the child skips any validation that could be happening in the `IOptionSettingHandler`. I assume at this point if a setting is being done at the object level we are assuming the whole object has been validated. This might not always be true and could be the source of future problems.

I'm fine revising this later to get this initial PR to unblock others but maybe making a comment here for future work we need to do.

src/AWS.Deploy.Common/Recipes/OptionSettingItem.cs
```diff
@@ -10,10 +10,19 @@
 
 namespace AWS.Deploy.Common.Recipes
 {
+    public interface IOptionSettingItem
```
normj, line 13: Add docs

normj (approved):

src/AWS.Deploy.Common/Recipes/OptionSettingItem.ValueOverride.cs
normj: Perfect, thanks

normj (approved):

96malhar (approved):

PR #553, opened 2022-05-12
fix: Add ProducesResponseType for the possible error codes in ServerMode API

normj:
*Description of changes:*
Add the `ProducesResponseType` attribute to all of the REST APIs in server mode. This allows the client code generator to generate a `ProblemDetails` type that the toolkit can use to get the error message for the bad error code.

### DIFF
--- a/src/AWS.Deploy.CLI/ServerMode/Controllers/DeploymentController.cs
+++ b/src/AWS.Deploy.CLI/ServerMode/Controllers/DeploymentController.cs
@@ -121,6 +121,7 @@ namespace AWS.Deploy.CLI.ServerMode.Controllers
         [HttpGet("session/<sessionId>/recommendations")]
         [SwaggerOperation(OperationId = "GetRecommendations")]
         [SwaggerResponse(200, type: typeof(GetRecommendationsOutput))]
+        [ProducesResponseType(Microsoft.AspNetCore.Http.StatusCodes.Status404NotFound)]
         [Authorize]
         public async Task<IActionResult> GetRecommendations(string sessionId)
         {
@@ -162,6 +163,7 @@ namespace AWS.Deploy.CLI.ServerMode.Controllers
         [HttpGet("session/<sessionId>/settings")]
         [SwaggerOperation(OperationId = "GetConfigSettings")]
         [SwaggerResponse(200, type: typeof(GetOptionSettingsOutput))]
+        [ProducesResponseType(Microsoft.AspNetCore.Http.StatusCodes.Status404NotFound)]
         [Authorize]
         public IActionResult GetConfigSettings(string sessionId)
         {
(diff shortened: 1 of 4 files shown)


src/AWS.Deploy.CLI/ServerMode/Controllers/DeploymentController.cs
```diff
@@ -382,6 +395,7 @@ public async Task<IActionResult> SetDeploymentTarget(string sessionId, [FromBody
                     // This is because pushing images to ECR does not need a cloud application name.
                     if (!cloudApplicationNameGenerator.IsValidName(newDeploymentName))
                     {
+                        //throw new InvalidParameterException(cloudApplicationNameGenerator.InvalidNameMessage(newDeploymentName));
```
philasmar, line 398: This is commented out

philasmar (changes requested):

src/AWS.Deploy.CLI/ServerMode/Controllers/DeploymentController.cs
ashovlin: And if removing this comment, it looks like it would have been the only use of `InvalidParameterException` above so could we remove that too?

normj: Whoops. I was playing around with having custom exceptions but since you can only have a custom exception per Http status code it didn't seem like it was buying much. Removed the orphaned code.

src/AWS.Deploy.CLI/ServerMode/Controllers/DeploymentController.cs
```diff
@@ -505,6 +521,7 @@ public async Task<IActionResult> GenerateCloudFormationTemplate(string sessionId
         /// </summary>
         [HttpPost("session/<sessionId>/execute")]
         [SwaggerOperation(OperationId = "StartDeployment")]
+        [ProducesResponseType(Microsoft.AspNetCore.Http.StatusCodes.Status404NotFound)]
```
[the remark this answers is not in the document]
normj: Good catch, I have set the status code for the `Problem` to `424 Failed Dependency` and added it to the `ProducesResponseType`.

awschristou (changes requested):

test/AWS.Deploy.CLI.IntegrationTests/ServerModeTests.cs
```diff
@@ -356,13 +399,13 @@ public async Task InvalidStackName_ThrowsException(string invalidStackName)
                         NewDeploymentRecipeId = fargateRecommendation.RecipeId
                     });
                 }
-                catch (ApiException ex)
+                catch (ApiException<ProblemDetails> ex)
```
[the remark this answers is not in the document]
normj: Good catch, fixed.

ashovlin (approved):

philasmar (approved):

awschristou (approved):

PR #557, opened 2022-05-12
fix: cannot use ActivatorUtilities.CreateInstance with internal class

philasmar:
*Issue #, if available:*

*Description of changes:*
This PR fixes the reason behind the release failure.


ashovlin (approved):
I think okay given this is primarily a CLI tool and not a class library so we may less concerned with surface area than usual, but would defer to @normj.

normj (approved):

PR #560, opened 2022-05-13
chore: Add constructor injection support to option and recipe validators

ashovlin:
*Issue #, if available:* N/A, this is to support upcoming work to improve validation on deployment bundle settings. https://github.com/aws/aws-dotnet-deploy/pull/509 will also make use of this.

*Description of changes:*

Previously, `IOptionSettingItemValidator` and `IRecipeValidator` were always initialized through parameterless constructors. The current set of validators are self contained (required, ranges, regexes) that did not need any dependencies.

However upcoming validators around the Docker file and execution directory require an instance of `IDirectoryManager` and `IFileManager`.

This reworks `ValidatorFactory` to be able to inject services into validators when initializing them. 

---

### DIFF
--- a/src/AWS.Deploy.CLI/Commands/DeployCommand.cs
+++ b/src/AWS.Deploy.CLI/Commands/DeployCommand.cs
@@ -52,6 +52,7 @@ namespace AWS.Deploy.CLI.Commands
         private readonly ICDKVersionDetector _cdkVersionDetector;
         private readonly IAWSServiceHandler _awsServiceHandler;
         private readonly IOptionSettingHandler _optionSettingHandler;
+        private readonly IValidatorFactory _validatorFactory;
 
         public DeployCommand(
             IServiceProvider serviceProvider,
@@ -76,7 +77,8 @@ namespace AWS.Deploy.CLI.Commands
             IDirectoryManager directoryManager,
             IFileManager fileManager,
             IAWSServiceHandler awsServiceHandler,
-            IOptionSettingHandler optionSettingHandler)
+            IOptionSettingHandler optionSettingHandler,
+            IValidatorFactory validatorFactory)
         {
             _serviceProvider = serviceProvider;
             _toolInteractiveService = toolInteractiveService;
@@ -101,6 +103,7 @@ namespace AWS.Deploy.CLI.Commands
             _systemCapabilityEvaluator = systemCapabilityEvaluator;
             _awsServiceHandler = awsServiceHandler;
             _optionSettingHandler = optionSettingHandler;
+            _validatorFactory = validatorFactory;
         }
 
         public async Task ExecuteAsync(string applicationName, string deploymentProjectPath, UserDeploymentSettings? userDeploymentSettings = null)
@@ -458,9 +461,8 @@ namespace AWS.Deploy.CLI.Commands
             }
 
             var validatorFailedResults =
-                        recommendation.Recipe
-                            .BuildValidators()
-                            .Select(validator => validator.Validate(recommendation, _session, _optionSettingHandler))
+                _validatorFactory.BuildValidators(recommendation.Recipe)
+                            .Select(validator => validator.Validate(recommendation, _session))
                             .Where(x => !x.IsValid)
                             .ToList();
 
@@ -768,9 +770,8 @@ namespace AWS.Deploy.CLI.Commands
                 if (string.IsNullOrEmpty(input))
                 {
                     var validatorFailedResults =
-                        recommendation.Recipe
-                            .BuildValidators()
-                            .Select(validator => validator.Validate(recommendation, _session, _optionSettingHandler))
+                        _validatorFactory.BuildValidators(recommendation.Recipe)
+                            .Select(validator => validator.Validate(recommendation, _session))
                             .Where(x => !x.IsValid)
                             .ToList();
 
--- a/src/AWS.Deploy.Common/Recipes/OptionSettingItem.ValueOverride.cs
+++ b/src/AWS.Deploy.Common/Recipes/OptionSettingItem.ValueOverride.cs
@@ -92,11 +92,11 @@ namespace AWS.Deploy.Common.Recipes
         /// Thrown if one or more <see cref="Validators"/> determine
         /// <paramref name="valueOverride"/> is not valid.
         /// </exception>
-        public void SetValue(IOptionSettingHandler optionSettingHandler, object valueOverride)
+        public void SetValue(IOptionSettingHandler optionSettingHandler, object valueOverride, IOptionSettingItemValidator[] validators)
         {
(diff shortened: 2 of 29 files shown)


src/AWS.Deploy.Common/Recipes/Validation/ValidatorFactory.cs
```diff
@@ -4,15 +4,45 @@
...
+    /// Factory that builds the validators for a given option or recipe
+    /// </summary>
+    public interface IValidatorFactory
+    {
+        /// <summary>
+        /// Builds the validators that apply to the given option
+        /// </summary>
+        /// <param name="optionSettingItem">Option to validate</param>
+        /// <returns>Array of validators for the given option</returns>
+        public IOptionSettingItemValidator[] BuildValidators(OptionSettingItem optionSettingItem);
```
philasmar, line 24: nit: no need for `public`

src/AWS.Deploy.Common/Recipes/Validation/ValidatorFactory.cs
```diff
@@ -4,15 +4,45 @@
...
+
+        /// <summary>
+        /// Builds the validators that apply to the given recipe
+        /// </summary>
+        /// <param name="recipeDefinition">Recipe to validate</param>
+        /// <returns>Array of validators for the given recipe</returns>
+        public IRecipeValidator[] BuildValidators(RecipeDefinition recipeDefinition);
```
philasmar, line 31: nit: no need for `public`

src/AWS.Deploy.CLI/Commands/DeployCommand.cs
```diff
@@ -458,8 +461,7 @@ private void ConfigureDeploymentFromConfigFile(Recommendation recommendation, Us
             }
 
             var validatorFailedResults =
-                        recommendation.Recipe
-                            .BuildValidators()
+                _validatorFactory.BuildValidators(recommendation.Recipe)
                             .Select(validator => validator.Validate(recommendation, _session, _optionSettingHandler))
```
philasmar, line 465: could you inject `_optionSettingHandler` as a dependency of `_validatorFactory`? that way we wouldn't need to update the `validator.Validate` signature every time we need a new service added?

src/AWS.Deploy.CLI/Commands/DeployCommand.cs
```diff
@@ -768,8 +770,7 @@ private async Task ConfigureDeploymentFromCli(Recommendation recommendation, IEn
                 if (string.IsNullOrEmpty(input))
                 {
                     var validatorFailedResults =
-                        recommendation.Recipe
```
philasmar, line 771: same comment as above

src/AWS.Deploy.Common/Recipes/OptionSettingItem.ValueOverride.cs
```diff
@@ -92,11 +92,11 @@ public object GetValue(IDictionary<string, string> replacementTokens, IDictionar
         /// Thrown if one or more <see cref="Validators"/> determine
         /// <paramref name="valueOverride"/> is not valid.
         /// </exception>
-        public void SetValue(IOptionSettingHandler optionSettingHandler, object valueOverride)
+        public void SetValue(IOptionSettingHandler optionSettingHandler, IValidatorFactory validatorFactory, object valueOverride)
```
philasmar, line 95: could you remove the `IValidatorFactory` as a dependency to `SetValue` and instead pass in the list of validators, which is the result of `validatorFactory.BuildValidators(this)`? `SetOptionSettingValue` in `OptionSettingHandler` is meant to handle anything related to services and only pass down simplified objects to `SetValue` in `OptionSettingItem`.

src/AWS.Deploy.Common/Recipes/OptionSettingItem.cs
```diff
@@ -38,9 +38,10 @@ public interface IOptionSettingItem
         /// <summary>
         /// Set the value of an <see cref="OptionSettingItem"/> while validating the provided input.
         /// </summary>
-        /// <param name="optionSettingHandler"></param>
-        /// <param name="value"></param>
-        void SetValue(IOptionSettingHandler optionSettingHandler, object value);
+        /// <param name="optionSettingHandler">Handler use to set any child option settings</param>
+        /// <param name="validatorFactory">Source of validators for this item</param>
+        /// <param name="value">Value to set</param>
+        void SetValue(IOptionSettingHandler optionSettingHandler, IValidatorFactory validatorFactory, object value);
```
philasmar, line 44: could you remove the `IValidatorFactory` as a dependency to `SetValue` and instead pass in the list of validators, which is the result of `validatorFactory.BuildValidators(this)`? `SetOptionSettingValue` in `OptionSettingHandler` is meant to handle anything related to services and only pass down simplified objects to `SetValue` in `OptionSettingItem`.

philasmar (changes requested):

src/AWS.Deploy.Common/Recipes/Validation/ValidatorFactory.cs
ashovlin: Dropped from the interface.

src/AWS.Deploy.Common/Recipes/Validation/ValidatorFactory.cs
ashovlin: Dropped from the interface.

src/AWS.Deploy.CLI/Commands/DeployCommand.cs
ashovlin: Updated both existing recipe validators with a `IOptionSettingHandler` in their constructor, which is injected in the new `ValidatorFactory`.  Removed `_optionSettingHandler` from the `Validate` calls.

src/AWS.Deploy.CLI/Commands/DeployCommand.cs
ashovlin: Updated both existing recipe validators with a `IOptionSettingHandler` in their constructor, which is injected in the new `ValidatorFactory`.  Removed `_optionSettingHandler` from the `Validate` calls.

src/AWS.Deploy.Common/Recipes/OptionSettingItem.ValueOverride.cs
ashovlin: Done

src/AWS.Deploy.Common/Recipes/OptionSettingItem.cs
ashovlin: Done

philasmar (approved):

normj (approved):

PR #564, opened 2022-05-17
fix: Validate and respect deployment bundle settings for the Docker or dotnet builds during server mode

ashovlin:
*Issue #, if available:* DOTNET-5922

*Description of changes:* 

**Problem**:
Option settings that are defined in the `.deploymentbundle` files (as opposed to the `.recipe` files) have been backed by `DeploymentBundle` class in addition to the `OptionSettingItem` collection.

The CLI mode writes directly to `DeploymentBundle` via the TypeHints, but server mode (and thus the Visual Studio Toolkit) was only setting the values in the `OptionSettingItem` objects.

The orchestrator was largely reading these options from `DeploymentBundle`, effectively ignoring them during server mode.

**Now**:
1. We "write-through" the `OptionSettingItem` to the `DeploymentBundle` for these options.
     * We were previously doing this during `ConfigureDeploymentFromConfigFile`, but now we do it for every `SetValue`
1. Moved validation for these settings to three new `IOptionSettingItemValidator` instances out of the TypeHint, since those don't run during server mode.
     * Validating `DockerExecutionDirectory` requires access to the recommendation object, so that we can validate a relative path from the project directory. This is extracting some in-progress work from https://github.com/aws/aws-dotnet-deploy/pull/509

---

### DIFF
--- a/src/AWS.Deploy.Common/IO/DirectoryManager.cs
+++ b/src/AWS.Deploy.Common/IO/DirectoryManager.cs
@@ -13,7 +13,30 @@ namespace AWS.Deploy.Common.IO
     {
         DirectoryInfo CreateDirectory(string path);
         DirectoryInfo GetDirectoryInfo(string path);
+
+        /// <summary>
+        /// Determines whether the given path refers to an existing directory on disk.
+        /// This can either be an absolute path or relative to the current working directory.
+        /// </summary>
+        /// <param name="path">The path to test</param>
+        /// <returns>
+        /// true if path refers to an existing directory;
+        /// false if the directory does not exist or an error occurs when trying to determine if the specified directory exists
+        /// </returns>
         bool Exists(string path);
+
+        /// <summary>
+        /// Determines whether the given path refers to an existing directory on disk.
+        /// This can either be an absolute path or relative to the given directory.
+        /// </summary>
+        /// <param name="path">The path to test</param>
+        /// <param name="relativeTo">Directory to consider the path as relative to</param>
+        /// <returns>
+        /// true if path refers to an existing directory;
+        /// false if the directory does not exist or an error occurs when trying to determine if the specified directory exists
+        /// </returns>
+        bool Exists(string path, string relativeTo);
+
         string[] GetFiles(string path, string? searchPattern = null, SearchOption searchOption = SearchOption.TopDirectoryOnly);
         string[] GetDirectories(string path, string? searchPattern = null, SearchOption searchOption = SearchOption.TopDirectoryOnly);
         bool IsEmpty(string path);
@@ -37,6 +60,18 @@ namespace AWS.Deploy.Common.IO
         public DirectoryInfo GetDirectoryInfo(string path) => new DirectoryInfo(path);
 
         public bool Exists(string path) => IsDirectoryValid(path);
+
+        public bool Exists(string path, string relativeTo)
+        {
+            if (Path.IsPathRooted(path))
+            {
+                return Exists(path);
+            }
+            else
+            {
+                return Exists(Path.Combine(relativeTo, path));
+            }
+        }
 
         public string[] GetFiles(string path, string? searchPattern = null, SearchOption searchOption = SearchOption.TopDirectoryOnly)
             => Directory.GetFiles(path, searchPattern ?? "*", searchOption);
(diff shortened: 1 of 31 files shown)


src/AWS.Deploy.Orchestration/OptionSettingHandler.cs
```diff
@@ -26,9 +26,47 @@ public OptionSettingHandler(IValidatorFactory validatorFactory)
...
+                    recommendation.DeploymentBundle.ECRRepositoryName = value.ToString() ?? string.Empty;
+                    break;
+                case "DotnetBuildConfiguration":
+                    recommendation.DeploymentBundle.DotnetPublishBuildConfiguration = value.ToString() ?? string.Empty;
+                    break;
+                case "DotnetPublishArgs":
+                    recommendation.DeploymentBundle.DotnetPublishAdditionalBuildArguments = value.ToString() ?? string.Empty;
+                    break;
+                case "SelfContainedBuild":
+                    recommendation.DeploymentBundle.DotnetPublishSelfContainedBuild = (bool)value;
```
normj, line 65: Instead of casting can we do `System.Convert.ToBoolean(value)` to make the code more defensive.

src/AWS.Deploy.Common/Recipes/Validation/OptionSettingItemValidators/DockerBuildArgsValidator.cs
```diff
@@ -0,0 +1,41 @@
...
+// SPDX-License-Identifier: Apache-2.0
+
+using System;
+
+namespace AWS.Deploy.Common.Recipes.Validation
+{
+    /// <summary>
+    /// Validator for Docker build-time variables, passed via --build-arg
+    /// </summary>
+    public class DockerBuildArgsValidator : IOptionSettingItemValidator
```
normj, line 11: We should be support any command line argument passed into `docker build` and not just `--build-arg`. Basically make this like the `DotnetPublishArgsValidator` and just make sure they aren't passing in a docker file or tag.

normj (changes requested):

src/AWS.Deploy.Orchestration/OptionSettingHandler.cs
ashovlin: Done

src/AWS.Deploy.Common/Recipes/Validation/OptionSettingItemValidators/DockerBuildArgsValidator.cs
ashovlin: Fixed, and added a separate commit so we get a second entry in the release notes calling out the breaking change.

Here's how the next release look's so far via `git-chglog --next-tag TBD TBD`:
```
### Bug Fixes

* Rework the "Docker Build Args" option to support appending any options to the docker build command, instead of just `--build-arg`
* Validate and respect deployment bundle settings for the Docker or dotnet builds during server mode

### Build System or External Dependencies

* Version bump to 0.45

### Chore

* Add constructor injection support to option and recipe validators
* Add process exit code to improve telemetry of failed processes

### BREAKING CHANGE


If you were previously using "Docker Build Args" to set build-time variables, you must now prepend `--build-arg` for each argument yourself.

```

philasmar (approved):

src/AWS.Deploy.Common/IO/DirectoryManager.cs
```diff
@@ -13,7 +13,30 @@ public interface IDirectoryManager
...
+        /// Determines whether the given path refers to an existing directory on disk.
+        /// This can either be an absolute path or relative to the given directory.
+        /// </summary>
+        /// <param name="path">The path to test.</param>
+        /// <param name="directory">Directory to consider the path as relative to</param>
+        /// <returns>
+        /// true if path refers to an existing directory;
+        /// false if the directory does not exist or an error occurs when trying to determine if the specified directory exists
+        /// </returns>
+        bool Exists(string path, string directory);
```
normj, line 38: I would recommend better argument names as `path` and `directory` mean basically the same thing. Maybe `path` and `relativeFrom` or something like that.

src/AWS.Deploy.Common/IO/DirectoryManager.cs
```diff
@@ -38,6 +61,18 @@ public class DirectoryManager : IDirectoryManager
 
         public bool Exists(string path) => IsDirectoryValid(path);
 
+        public bool Exists(string path, string directory)
```
normj, line 64: Same comment as above about argument names.

normj (approved):
Just a minor comment on naming, otherwise looks good.

96malhar (approved):

src/AWS.Deploy.Common/IO/DirectoryManager.cs
ashovlin: Went with `relativeTo` inspired by `Path.GetRelativePath`

src/AWS.Deploy.Common/IO/DirectoryManager.cs
ashovlin: Updated with `relativeTo` inspired by `Path.GetRelativePath`

PR #565, opened 2022-05-19
chore: Improve error message for missing NodeJS dependency

96malhar:
*Issue #, if available:*
DOTNET-5889

*Description of changes:*
This PR makes the following changes:
* Improves the error message for missing NodeJS dependency by explicitly asking the user to restart their IDE/Shell after the installation process completes. This is done so that the dependencies can be recognized correctly.
* Be explicit on on how to add missing dependency if an `InstallationUrl` is available.
* Removes and refactors some redundant code.

---

### Current error message
image

### Proposed error message
image

### DIFF
--- a/src/AWS.Deploy.CLI/ServerMode/Models/SystemCapabilitySummary.cs
+++ b/src/AWS.Deploy.CLI/ServerMode/Models/SystemCapabilitySummary.cs
@@ -6,16 +6,14 @@ namespace AWS.Deploy.CLI.ServerMode.Models
     public class SystemCapabilitySummary
     {
         public string Name { get; set; }
-        public bool Installed { get; set; }
-        public bool Available { get; set; }
-        public string? Message { get; set; }
+        public string Message { get; set; }
         public string? InstallationUrl { get; set; }
 
-        public SystemCapabilitySummary(string name, bool installed, bool available)
+        public SystemCapabilitySummary(string name, string message, string? installationUrl = null)
         {
             Name = name;
-            Installed = installed;
-            Available = available;
+            Message = message;
+            InstallationUrl = installationUrl;
         }
     }
 }
--- a/src/AWS.Deploy.ServerMode.Client/RestAPI.cs
+++ b/src/AWS.Deploy.ServerMode.Client/RestAPI.cs
@@ -1789,7 +1789,7 @@ namespace AWS.Deploy.ServerMode.Client
         [Newtonsoft.Json.JsonProperty("displayName", Required = Newtonsoft.Json.Required.Default, NullValueHandling = Newtonsoft.Json.NullValueHandling.Ignore)]
         public string DisplayName { get; set; }
     
-        /// <summary>The order used to sort categories in UI screens. Categories will be showin in sorted descending order.</summary>
+        /// <summary>The order used to sort categories in UI screens. Categories will be shown in sorted descending order.</summary>
         [Newtonsoft.Json.JsonProperty("order", Required = Newtonsoft.Json.Required.DisallowNull, NullValueHandling = Newtonsoft.Json.NullValueHandling.Ignore)]
         public int Order { get; set; }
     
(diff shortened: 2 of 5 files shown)


src/AWS.Deploy.CLI/ServerMode/Models/SystemCapabilitySummary.cs
```diff
@@ -5,17 +5,15 @@ namespace AWS.Deploy.CLI.ServerMode.Models
 {
     public class SystemCapabilitySummary
     {
-        public string Name { get; set; }
-        public bool Installed { get; set; }
```
awschristou, line 9: The interface was changed, so the Rest API generator needs to be re-run

awschristou (changes requested):

src/AWS.Deploy.CLI/ServerMode/Models/SystemCapabilitySummary.cs
96malhar: Re-ran the generator and updated the PR.

src/AWS.Deploy.ServerMode.Client/RestAPI.cs
```diff
@@ -2190,21 +2190,6 @@ public partial class StartDeploymentSessionOutput
     [System.CodeDom.Compiler.GeneratedCode("NJsonSchema", "10.4.1.0 (Newtonsoft.Json v13.0.0.0)")]
     public partial class SystemCapabilitySummary 
     {
-        [Newtonsoft.Json.JsonProperty("name", Required = Newtonsoft.Json.Required.Default, NullValueHandling = Newtonsoft.Json.NullValueHandling.Ignore)]
-        public string Name { get; set; }
```
awschristou, line 2194: This might be due to the removal of "{ get; set; }"
Please ensure the Rest API is emitting the expected properties, and add an integration test to help guard against this kind of accident. This would break compilation if a release was integrated into a client.

awschristou (changes requested):

src/AWS.Deploy.ServerMode.Client/RestAPI.cs
96malhar: I could call the `GetCompatibility` API in one of our existing integrations tests but the result would always be an empty list as all the dependencies are satisfied.

I looked into modifying the environment variables to simulate the absence of Node.js or Docker. However, our CI system runs on a Linux machine and .NET implementation on UNIX based OS only supports manipulations on process-scoped environment variables.

For now, to ensure the API response is as expected, I performed a manual test by deleteing Node.js from my system

```
public async Task CheckCompatibilities()
        {
            var projectPath = _testAppManager.GetProjectPath(Path.Combine("testapps", "WebAppWithDockerFile", "WebAppWithDockerFile.csproj"));
            var portNumber = 4012;
            using var httpClient = ServerModeHttpClientFactory.ConstructHttpClient(ResolveCredentials);

            var serverCommand = new ServerModeCommand(_serviceProvider.GetRequiredService<IToolInteractiveService>(), portNumber, null, true);
            var cancelSource = new CancellationTokenSource();

            var serverTask = serverCommand.ExecuteAsync(cancelSource.Token);
            try
            {
                var baseUrl = $"http://localhost:{portNumber}/";
                var restClient = new RestAPIClient(baseUrl, httpClient);

                await WaitTillServerModeReady(restClient);

                var startSessionOutput = await restClient.StartDeploymentSessionAsync(new StartDeploymentSessionInput
                {
                    AwsRegion = _awsRegion,
                    ProjectPath = projectPath
                });

                var sessionId = startSessionOutput.SessionId;
                var getRecommendationOutput = await restClient.GetRecommendationsAsync(sessionId);
                var fargateRecommendation = getRecommendationOutput.Recommendations.FirstOrDefault(x => string.Equals(x.RecipeId, "AspNetAppEcsFargate"));

                await restClient.SetDeploymentTargetAsync(sessionId, new SetDeploymentTargetInput
                {
                    NewDeploymentRecipeId = fargateRecommendation.RecipeId,
                    NewDeploymentName = "MyAppStack"
                });

                var result = await restClient.GetCompatibilityAsync(sessionId);

                Assert.Equal(1, result.Capabilities.Count);

                var capability = result.Capabilities.First();

                Assert.Equal("Node.js", capability.Name);
                Assert.Equal("https://nodejs.org/en/download/", capability.InstallationUrl);
                Assert.False(string.IsNullOrEmpty(capability.Message));
            }
            finally
            {
                cancelSource.Cancel();
            }
        }
```

awschristou (approved):

ashovlin (approved):

philasmar (approved):

PR #570, opened 2022-05-23
chore: create an option setting dependency tree

philasmar:
*Issue #, if available:*
DOTNET-5920

*Description of changes:*
* Build an option setting dependency tree that allows us to easily determine which option settings need to be validated when a value update happens.
* Added Leaf ID and Parent fields to option settings to allow easy navigation of option settings

### DIFF
--- a/src/AWS.Deploy.Common/Recipes/OptionSettingItem.cs
+++ b/src/AWS.Deploy.Common/Recipes/OptionSettingItem.cs
@@ -58,6 +58,17 @@ namespace AWS.Deploy.Common.Recipes
         public string Id { get; set; }
 
         /// <summary>
+        /// The fully qualified id of the setting that includes the Id and all of the parent's Ids.
+        /// This helps easily reference the Option Setting without context of the parent setting.
+        /// </summary>
+        public string FullyQualifiedId { get; set; }
+
+        /// <summary>
+        /// The parent Option Setting Item that allows an option setting item to reference it's parent.
+        /// </summary>
+        public string? ParentId { get; set; }
+
+        /// <summary>
         /// The id of the parent option setting. This is used for tooling that wants to look up the existing resources for a setting based on the TypeHint but needs
         /// to know the parent AWS resource. For example if listing the available Beanstalk environments the listing should be for the environments of the Beanstalk application.
         /// </summary>
@@ -145,9 +156,15 @@ namespace AWS.Deploy.Common.Recipes
         /// </summary>
         public Dictionary<string, object> TypeHintData { get; set; } = new ();
 
-        public OptionSettingItem(string id, string name, string description)
+        /// <summary>
+        /// Indicates which option setting items need to be validated if a value update occurs.
+        /// </summary>
+        public List<string> Dependents { get; set; } = new List<string> ();
+
+        public OptionSettingItem(string id, string fullyQualifiedId, string name, string description)
         {
             Id = id;
+            FullyQualifiedId = fullyQualifiedId;
             Name = name;
             Description = description;
         }
--- a/src/AWS.Deploy.Orchestration/OptionSettingHandler.cs
+++ b/src/AWS.Deploy.Orchestration/OptionSettingHandler.cs
@@ -140,6 +140,44 @@ namespace AWS.Deploy.Orchestration
         }
 
         /// <summary>
+        /// Interactively traverses given json path and returns target option setting.
+        /// Throws exception if there is no <see cref="OptionSettingItem" /> that matches <paramref name="jsonPath"/> />
+        /// In case an option setting of type <see cref="OptionSettingValueType.KeyValue"/> is encountered,
+        /// that <paramref name="jsonPath"/> can have the key value pair name as the leaf node with the option setting Id as the node before that.
+        /// </summary>
+        /// <param name="jsonPath">
+        /// Dot (.) separated key values string pointing to an option setting.
+        /// Read more <see href="https://tools.ietf.org/id/draft-goessner-dispatch-jsonpath-00.html"/>
+        /// </param>
+        /// <returns>Option setting at the json path. Throws <see cref="OptionSettingItemDoesNotExistException"/> if there doesn't exist an option setting.</returns>
+        public OptionSettingItem GetOptionSetting(RecipeDefinition recipe, string? jsonPath)
+        {
+            if (string.IsNullOrEmpty(jsonPath))
+                throw new OptionSettingItemDoesNotExistException(DeployToolErrorCode.OptionSettingItemDoesNotExistInRecipe, $"An option setting item with the specified fully qualified Id '{jsonPath}' cannot be found in the" +
+                    $" '{recipe.Name}' recipe.");
+
+            var ids = jsonPath.Split('.');
+            OptionSettingItem? optionSetting = null;
+
+            for (int i = 0; i < ids.Length; i++)
+            {
+                var optionSettings = optionSetting?.ChildOptionSettings ?? recipe.OptionSettings;
+                optionSetting = optionSettings.FirstOrDefault(os => os.Id.Equals(ids[i]));
+                if (optionSetting == null)
+                {
+                    throw new OptionSettingItemDoesNotExistException(DeployToolErrorCode.OptionSettingItemDoesNotExistInRecipe, $"An option setting item with the specified fully qualified Id '{jsonPath}' cannot be found in the" +
+                    $" '{recipe.Name}' recipe.");
+                }
+                if (optionSetting.Type.Equals(OptionSettingValueType.KeyValue))
+                {
+                    return optionSetting;
+                }
+            }
+
+            return optionSetting!;
+        }
+
+        /// <summary>
         /// Retrieves the value of the Option Setting Item in a given recommendation.
         /// </summary>
         public T GetOptionSettingValue<T>(Recommendation recommendation, OptionSettingItem optionSetting)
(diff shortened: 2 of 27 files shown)


src/AWS.Deploy.Common/Recipes/OptionSettingItem.cs
```diff
@@ -57,6 +57,17 @@ public partial class OptionSettingItem : IOptionSettingItem
         /// </summary>
         public string Id { get; set; }
 
+        /// <summary>
+        /// The Id of the option setting item in the Recipe Option Setting tree.
+        /// This helps easily reference the Option Setting without context of the parent setting.
+        /// </summary>
+        public string LeafId { get; set; }
```
normj, line 64: `FullyQualifiedId` or something like that would make more sense to me then `LeafId`. When I think of leaf I expected it to be only the last part of the id which is the same as the id field.

src/AWS.Deploy.Orchestration/OptionSettingHandler.cs
```diff
@@ -139,6 +139,44 @@ public OptionSettingItem GetOptionSetting(Recommendation recommendation, string?
...
+        /// </summary>
+        /// <param name="jsonPath">
+        /// Dot (.) separated key values string pointing to an option setting.
+        /// Read more <see href="https://tools.ietf.org/id/draft-goessner-dispatch-jsonpath-00.html"/>
+        /// </param>
+        /// <returns>Option setting at the json path. Throws <see cref="OptionSettingItemDoesNotExistException"/> if there doesn't exist an option setting.</returns>
+        public OptionSettingItem GetOptionSetting(RecipeDefinition recipe, string? jsonPath)
+        {
+            if (string.IsNullOrEmpty(jsonPath))
+                throw new OptionSettingItemDoesNotExistException(DeployToolErrorCode.OptionSettingItemDoesNotExistInRecipe, $"The Option Setting Item {jsonPath} does not exist as part of the" +
```
normj, line 156: If `jsonPath` is null or empty the error message should say that not that the key can't be found.

src/AWS.Deploy.Orchestration/OptionSettingHandler.cs
```diff
@@ -139,6 +139,44 @@ public OptionSettingItem GetOptionSetting(Recommendation recommendation, string?
...
+                    $" {recipe.Name} recipe");
+
+            var ids = jsonPath.Split('.');
```
normj, line 159: We should probably update the recipe schema to put validation that doesn't allow `.` in the setting id field.

src/AWS.Deploy.Orchestration/RecipeHandler.cs
```diff
@@ -225,5 +234,31 @@ private bool ContainsRecipeFile(string directoryPath)
...
+        /// </summary>
+        private void BuildDependencyTree(RecipeDefinition recipe, List<OptionSettingItem> optionSettingItems, Dictionary<string, List<string>> dependencyTree, string parentLeafId = "")
+        {
+            foreach (var optionSettingItem in optionSettingItems)
+            {
+                optionSettingItem.LeafId = $"{parentLeafId}{optionSettingItem.Id}";
+                optionSettingItem.ParentId = parentLeafId.EndsWith(".") ? parentLeafId.Substring(0, parentLeafId.Length - 1) : parentLeafId;
+                foreach (var dependency in optionSettingItem.DependsOn)
+                {
+                    if (dependencyTree.ContainsKey(dependency.Id))
```
normj, line 251: Shouldn't the key in this dictionary be the settings full id? Meaning if a recipe had 2 `CreateNew` setting ids under 2 different object settings those ids would get grouped together here.

normj (changes requested):
Where are the tests that the `Dependents` collection is correctly filled. You also might need to create some dummy recipe files to test scenarios we don't have covered in our existing recipes. In particular I want to make sure that if somebody accidently creates a recipe with a recursive depends on we don't get stuck in an infinite loop and crash VS.

src/AWS.Deploy.Orchestration/RecipeHandler.cs
philasmar: https://github.com/aws/aws-dotnet-deploy/blob/aacb712b5be4f1816ad9cb6933927c1b4d93811c/src/AWS.Deploy.Common/Recipes/PropertyDependency.cs#L11

src/AWS.Deploy.Common/Recipes/OptionSettingItem.cs
philasmar: Done

src/AWS.Deploy.Orchestration/OptionSettingHandler.cs
philasmar: Done

src/AWS.Deploy.Orchestration/OptionSettingHandler.cs
philasmar: Done. I added a pattern that supports letters, numbers and - and _

philasmar:
> Where are the tests that the `Dependents` collection is correctly filled. You also might need to create some dummy recipe files to test scenarios we don't have covered in our existing recipes. In particular I want to make sure that if somebody accidently creates a recipe with a recursive depends on we don't get stuck in an infinite loop and crash VS.

I added a new test class that handles the happy path and cyclic dependencies.

src/AWS.Deploy.Orchestration/RecipeHandler.cs
```diff
@@ -225,5 +236,31 @@ private bool ContainsRecipeFile(string directoryPath)
...
+        /// Creates an option setting item dependency tree that indicates
+        /// which option setting items need to be validated if a value update occurs.
+        /// The function recursively goes through all the settings and their children to build this tree.
+        /// This method also creates a Leaf Id which will help reference <see cref="OptionSettingItem"/>.
+        /// </summary>
+        private void BuildDependencyTree(RecipeDefinition recipe, List<OptionSettingItem> optionSettingItems, Dictionary<string, List<string>> dependencyTree, string parentLeafId = "")
+        {
+            foreach (var optionSettingItem in optionSettingItems)
+            {
+                optionSettingItem.FullyQualifiedId = $"{parentLeafId}{optionSettingItem.Id}";
```
normj, line 249: Scan the code change to remove the term `Leaf`. Since I was already pushy to rename the property lets not have the artifacts of 2 different naming patterns.

normj (approved):
A few naming comments. After those are addressed I'm good with the change.

96malhar (changes requested):
All substring occurrences of `leafId` must be replaced with `fullyQualifiedId`

96malhar (changes requested):
Once we construct the dependency tree, we _"walk"_ on the tree and identify any cycles. If cycles are found then the appropriate exception must be thrown.

96malhar (approved):

src/AWS.Deploy.Common/Recipes/OptionSettingItem.cs
```diff
@@ -145,9 +156,15 @@ public partial class OptionSettingItem : IOptionSettingItem
...
-        public OptionSettingItem(string id, string name, string description)
+        /// <summary>
+        /// Indicates which option setting items need to be validated if a value update occurs.
+        /// </summary>
+        public List<string> Dependents { get; set; } = new List<string> ();
+
+        public OptionSettingItem(string id, string leafId, string name, string description)
         {
             Id = id;
+            FullyQualifiedId = leafId;
```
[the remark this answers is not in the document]
philasmar: Done

src/AWS.Deploy.Orchestration/RecipeHandler.cs
philasmar: Done

src/AWS.Deploy.Orchestration/RecipeHandler.cs
```diff
@@ -50,17 +54,24 @@ public async Task<List<RecipeDefinition>> GetRecipeDefinitions(ProjectDefinition
...
                             definition.RecipePath = recipeDefinitionFile;
                             if (!uniqueRecipeId.Contains(definition.Id))
                             {
+                                var dependencyTree = new Dictionary<string, List<string>>();
+                                BuildDependencyTree(definition, definition.OptionSettings, dependencyTree);
+                                foreach (var dependent in dependencyTree.Keys)
+                                {
+                                    var optionSetting = _optionSettingHandler.GetOptionSetting(definition, dependent);
+                                    optionSetting.Dependents = dependencyTree[dependent];
+                                }
```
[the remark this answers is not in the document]
philasmar: Done

src/AWS.Deploy.Orchestration/RecipeHandler.cs
```diff
@@ -225,5 +236,31 @@ private bool ContainsRecipeFile(string directoryPath)
...
+                        dependencyTree[dependency.Id].Add(optionSettingItem.FullyQualifiedId);
+                    }
+                    else
+                    {
+                        dependencyTree[dependency.Id] = new List<string> { optionSettingItem.FullyQualifiedId };
+                    }
+                }
+                BuildDependencyTree(recipe, optionSettingItem.ChildOptionSettings, dependencyTree, $"{optionSettingItem.FullyQualifiedId}.");
+            }
+        }
```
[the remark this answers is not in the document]
philasmar: Done

PR #571, opened 2022-05-23
chore: Alter option setting names in Elastic Beanstalk recipe

96malhar:
*Issue #, if available:*
DOTNET-5882

*Description of changes:*
This PR changes the top-level `BeanstalkApplication` option setting name to `Elastic Beanstalk Application`. This is done because the original option setting name (`Application Name`) was also being used to set the name of the CloudFormation stack. Using the same name to represent the different constructs can be confusing to users.

### DIFF
--- a/src/AWS.Deploy.Recipes/RecipeDefinitions/ASP.NETAppElasticBeanstalk.recipe
+++ b/src/AWS.Deploy.Recipes/RecipeDefinitions/ASP.NETAppElasticBeanstalk.recipe
@@ -128,9 +128,9 @@
     "OptionSettings": [
         {
             "Id": "BeanstalkApplication",
-            "Name": "Application Name",
+            "Name": "Elastic Beanstalk Application",
             "Category": "General",
-            "Description": "The Elastic Beanstalk application name.",
+            "Description": "The Elastic Beanstalk application.",
             "Type": "Object",
             "TypeHint": "BeanstalkApplication",
             "AdvancedSetting": false,


philasmar (changes requested):
The beanstalk environment is intentionally done this way to avoid a breaking change. Norm and I have discussed this before and we decided against any action on the current structure.

96malhar:
> The beanstalk environment is intentionally done this way to avoid a breaking change. Norm and I have discussed this before and we decided against any action on the current structure.

Removed the changes related to Beanstalk Environment.

normj (approved):

ashovlin (changes requested):
Like we discussed at stand-up yesterday, could you add a unit test asserting on the current structure of the relevant Beanstalk recipe options, with a note why we shouldn't update them?

96malhar:
> Like we discussed at stand-up yesterday, could you add a unit test asserting on the current structure of the relevant Beanstalk recipe options, with a note why we shouldn't update them?

We want to assert on the recipe structure of all the recipes and not only the Beanstalk one. This can be tackled as part of a separate story. I will go ahead and create a backlog item for it

ashovlin (approved):

96malhar:
> > Like we discussed at stand-up yesterday, could you add a unit test asserting on the current structure of the relevant Beanstalk recipe options, with a note why we shouldn't update them?
> 
> We want to assert on the recipe structure of all the recipes and not only the Beanstalk one. This can be tackled as part of a separate story. I will go ahead and create a backlog item for it

Create DOTNET-5998 to add unit tests that assert on all recipe structures and can detect breaking changes.

philasmar (approved):

PR #572, opened 2022-05-25
 feat: Add URI validator and set default value for resource path pattern in the Blazor recipe

96malhar:
*Issue #, if available:*
* DOTNET-5805
* DOTNET-5806

*Description of changes:*
* Add new `UriValidator` class
* Enforces the `UriValidator` and `RegexValidator` on `BackendApi.Uri`.
* Set default value for `BackendApi.ResourcePathPattern` and impose the `RegexValidator` on it.

image
image

### DIFF
--- a/src/AWS.Deploy.Common/Recipes/Validation/OptionSettingItemValidators/UriValidator.cs
+++ b/src/AWS.Deploy.Common/Recipes/Validation/OptionSettingItemValidators/UriValidator.cs
@@ -1,0 +1,44 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+using System;
+using System.Collections.Generic;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace AWS.Deploy.Common.Recipes.Validation
+{
+    /// <summary>
+    /// Checks if a URI is structurally valid
+    /// </summary>
+    public class UriValidator : IOptionSettingItemValidator
+    {
+        public string ValidationFailedMessage { get; set; } = "{{URI}} is not a valid URI.";
+
+        public Task<ValidationResult> Validate(object input, Recommendation recommendation, OptionSettingItem optionSetting)
+        {
+            var uri = input?.ToString() ?? string.Empty;
+
+            /// It is possible that a URI specific option setting item is optional and can be null or empty.
+            /// To enforce the presence of a non-null or non-empty value, you must combine this validator with a <see cref="RequiredValidator"/>
+            if (string.IsNullOrEmpty(uri))
+            {
+                return ValidationResult.ValidAsync();
+            }
+
+            var message = ValidationFailedMessage.Replace("{{URI}}", uri);
+            try
+            {
+                var uriResult = new Uri(uri);
+                if (!Uri.IsWellFormedUriString(uri, UriKind.Absolute))
+                    return ValidationResult.FailedAsync(message);
+            }
+            catch (UriFormatException)
+            {
+                return ValidationResult.FailedAsync(message);
+            }
+
+            return ValidationResult.ValidAsync();
+        }
+    }
+}
--- a/src/AWS.Deploy.Recipes/RecipeDefinitions/BlazorWasm.recipe
+++ b/src/AWS.Deploy.Recipes/RecipeDefinitions/BlazorWasm.recipe
@@ -174,6 +174,14 @@
                             "Id": "BackendApi.Enable",
                             "Value": true
                         }
+                    ],
+                    "Validators": [
+                        {
+                            "ValidatorType": "Uri"
+                        },
+                        {
+                            "ValidatorType": "Required"
+                        }
                     ]
                 },
                 {
@@ -181,13 +189,22 @@
                     "Name": "Resource Path Pattern",
                     "Description": "The resource path pattern to determine which request go to backend rest api. (i.e. \"/api/*\") ",
                     "Type": "String",
-                    "DefaultValue": "",
+                    "DefaultValue": "/api/*",
                     "AdvancedSetting": false,
                     "Updatable": true,
                     "DependsOn": [
                         {
                             "Id": "BackendApi.Enable",
                             "Value": true
+                        }
+                    ],
+                    "Validators": [
+                        {
+                            "ValidatorType": "Regex",
+                            "Configuration": {
+                                "Regex": "^/\\S+$",
+                                "ValidationFailedMessage": "Invalid resource path pattern. The resource path must start with a forward slash (/) followed by one or more unicode characters"
+                            }
                         }
                     ]
                 }
--- a/src/AWS.Deploy.CLI/Commands/DeployCommand.cs
+++ b/src/AWS.Deploy.CLI/Commands/DeployCommand.cs
@@ -831,7 +831,7 @@ namespace AWS.Deploy.CLI.Commands
                 }
             }
 
-            if (!Equals(settingValue, currentValue) && settingValue != null)
+            if (settingValue != null)
             {
                 try
                 {
--- a/src/AWS.Deploy.Common/Recipes/Validation/OptionSettingItemValidatorList.cs
+++ b/src/AWS.Deploy.Common/Recipes/Validation/OptionSettingItemValidatorList.cs
@@ -52,6 +52,9 @@ namespace AWS.Deploy.Common.Recipes.Validation
         /// <summary>
         /// Must be paired with <see cref="SecurityGroupsInVpcValidator"/>
         /// </summary>
-        SecurityGroupsInVpc
+        SecurityGroupsInVpc,
+        /// Must be paired with <see cref="UriValidator"/>
+        /// </summary>
+        Uri
     }
 }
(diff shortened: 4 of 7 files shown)


ashovlin (commented):
While we're in this area - we're not enforcing the requirement on `Uri` which is failing during CDK deploy if left blank. 

```
Unhandled exception. System.UriFormatException: Invalid URI: The URI is empty.
   at System.Uri.CreateThis(String uri, Boolean dontEscape, UriKind uriKind)
   at System.Uri..ctor(String uriString)
   at BlazorWasm50.Recipe.ConfigureCloudFrontDistribution(Configuration settings) in C:\Users\shovlia\.aws-dotnet-deploy\Projects\cv24keup\Generated\Recipe.cs:line 167
   at BlazorWasm50.Recipe..ctor(Construct scope, IRecipeProps`1 props) in C:\Users\shovlia\.aws-dotnet-deploy\Projects\cv24keup\Generated\Recipe.cs:line 48
   at BlazorWasm50.AppStack..ctor(Construct scope, IDeployToolStackProps`1 props) in C:\Users\shovlia\.aws-dotnet-deploy\Projects\cv24keup\AppStack.cs:line 31
   at BlazorWasm50.Program.Main(String[] args) in C:\Users\shovlia\.aws-dotnet-deploy\Projects\cv24keup\Program.cs:line 29

Subprocess exited with error 3762504530
We were unable to delete the temporary project that was created for this deployment. 
```

Could we add that why we're here? If not, is there a separate issue tracking that?

philasmar:
> While we're in this area - we're not enforcing the requirement on `Uri` which is failing during CDK deploy if left blank.
> 
> I'm allowed to publish this: image
> 
> But it results in:
> 
> ```
> Unhandled exception. System.UriFormatException: Invalid URI: The URI is empty.
>    at System.Uri.CreateThis(String uri, Boolean dontEscape, UriKind uriKind)
>    at System.Uri..ctor(String uriString)
>    at BlazorWasm50.Recipe.ConfigureCloudFrontDistribution(Configuration settings) in C:\Users\shovlia\.aws-dotnet-deploy\Projects\cv24keup\Generated\Recipe.cs:line 167
>    at BlazorWasm50.Recipe..ctor(Construct scope, IRecipeProps`1 props) in C:\Users\shovlia\.aws-dotnet-deploy\Projects\cv24keup\Generated\Recipe.cs:line 48
>    at BlazorWasm50.AppStack..ctor(Construct scope, IDeployToolStackProps`1 props) in C:\Users\shovlia\.aws-dotnet-deploy\Projects\cv24keup\AppStack.cs:line 31
>    at BlazorWasm50.Program.Main(String[] args) in C:\Users\shovlia\.aws-dotnet-deploy\Projects\cv24keup\Program.cs:line 29
> 
> Subprocess exited with error 3762504530
> We were unable to delete the temporary project that was created for this deployment. 
> ```
> 
> Could we add that why we're here? If not, is there a separate issue tracking that?

The story that i am working on will help in this area. So when you hit enable Backend rest api, that'll run the validators of the dependent settings. Only issue here is that the URI does not have a regex validator. So I think part of this story, Malhar should add a regex validator, which will be automatically executed once I finish my feature.

philasmar (changes requested):
https://github.com/aws/aws-dotnet-deploy/pull/572#issuecomment-1137800290

96malhar:
> While we're in this area - we're not enforcing the requirement on `Uri` which is failing during CDK deploy if left blank.
> 
> I'm allowed to publish this: image
> 
> But it results in:
> 
> ```
> Unhandled exception. System.UriFormatException: Invalid URI: The URI is empty.
>    at System.Uri.CreateThis(String uri, Boolean dontEscape, UriKind uriKind)
>    at System.Uri..ctor(String uriString)
>    at BlazorWasm50.Recipe.ConfigureCloudFrontDistribution(Configuration settings) in C:\Users\shovlia\.aws-dotnet-deploy\Projects\cv24keup\Generated\Recipe.cs:line 167
>    at BlazorWasm50.Recipe..ctor(Construct scope, IRecipeProps`1 props) in C:\Users\shovlia\.aws-dotnet-deploy\Projects\cv24keup\Generated\Recipe.cs:line 48
>    at BlazorWasm50.AppStack..ctor(Construct scope, IDeployToolStackProps`1 props) in C:\Users\shovlia\.aws-dotnet-deploy\Projects\cv24keup\AppStack.cs:line 31
>    at BlazorWasm50.Program.Main(String[] args) in C:\Users\shovlia\.aws-dotnet-deploy\Projects\cv24keup\Program.cs:line 29
> 
> Subprocess exited with error 3762504530
> We were unable to delete the temporary project that was created for this deployment. 
> ```
> 
> Could we add that why we're here? If not, is there a separate issue tracking that?

Added a default value for the URI field and also added the appropriate validators to the URI and the resource path pattern.

src/AWS.Deploy.Recipes/RecipeDefinitions/BlazorWasm.recipe
```diff
@@ -166,29 +166,43 @@
                     "Name": "Uri",
                     "Description": "Uri to the backend rest api",
                     "Type": "String",
-                    "DefaultValue": "",
+                    "DefaultValue": "http://api-uri",
```
normj, line 169: I think we should have no default value for this field so the user has to enter a value. We want the toolkit to flag you must put a value here before publishing.

normj (changes requested):

src/AWS.Deploy.Recipes/RecipeDefinitions/BlazorWasm.recipe
philasmar: To Norm's point, even if a user tries to deploy with an invalid value, the validators will all be run before we kick off deployments which should catch this https://github.com/aws/aws-dotnet-deploy/blob/e2ad9b5a026869aa8854182dd5df1f4cedc6fa46/src/AWS.Deploy.CLI/Commands/DeployCommand.cs#L723

philasmar: In general, for this https://github.com/aws/aws-dotnet-deploy/blob/main/src/AWS.Deploy.CLI/Commands/DeployCommand.cs#L834, I vote to remove that check. any objections @normj ?

normj (approved):

src/AWS.Deploy.Recipes/RecipeDefinitions/BlazorWasm.recipe
96malhar: We can remove this check (If we decide to) as part of a separate PR.

96malhar: Also, setting invalid default values can easily trip customers as they could be performing deployments via a config file and may not specify an explicit value for all option settings

philasmar (approved):

src/AWS.Deploy.Common/Recipes/Validation/OptionSettingItemValidators/UriValidator.cs
```diff
@@ -0,0 +1,42 @@
...
+    {
+        public string ValidationFailedMessage { get; set; } = "{{URI}} is not a valid URI.";
+
+        public Task<ValidationResult> Validate(object input, Recommendation recommendation)
+        {
+            var uri = input?.ToString() ?? string.Empty;
+
+            if (string.IsNullOrEmpty(uri))
+            {
+                return ValidationResult.FailedAsync("Invalid URI. The URI cannot be null or empty");
```
ashovlin, line 24: Is true of all future `Uri` options? 

If not and we may conceivably add an option Uri option in the future, I'd rather use a `Required` validator here to enforce that a value is provided.

src/AWS.Deploy.Recipes/RecipeDefinitions/BlazorWasm.recipe
ashovlin: But `"http://api-uri"` is likely invalid for the user's application that they're deploying, right? 

I don't think we should be defaulting a value just to get past client side validation, then deploy an application that likely won't work.

It feels like we have to have a concept/experience for _"this field is required even though there isn't a reasonable default value so you must provide one"_

src/AWS.Deploy.Common/Recipes/Validation/OptionSettingItemValidators/UriValidator.cs
96malhar: I cannot foresee such a use case right now. But even if we encounter it in the future than it should be easy to remove the
`string.IsNullOrEmpty(uri)`  check. Also, this removal would not be a breaking change as we would be making the validator more permissive.

src/AWS.Deploy.Recipes/RecipeDefinitions/BlazorWasm.recipe
96malhar: This needs a larger discussion.

src/AWS.Deploy.Common/Recipes/Validation/OptionSettingItemValidators/UriValidator.cs
ashovlin: What if a user is using a custom recipe with a Uri option? Could they modify the source of the validators?

In general I would prefer validators as self-contained and composable as possible.

src/AWS.Deploy.Recipes/RecipeDefinitions/BlazorWasm.recipe
96malhar: After syncing up with @normj, removed the default value on `BackendApi.Uri` and also removed  https://github.com/aws/aws-dotnet-deploy/blob/main/src/AWS.Deploy.CLI/Commands/DeployCommand.cs#L834

philasmar (approved):

normj (approved):

PR #573, opened 2022-05-26
feat: run dependent settings validators when an option setting value is set

philasmar:
*Issue #, if available:*
DOTNET-5920

*Description of changes:*
* Added a validation status and validation message to option setting items
* The validators of dependent settings are now run when that setting is set
* Updated server mode to return the validation object

### DIFF
--- a/src/AWS.Deploy.CLI/ServerMode/Models/OptionSettingItemSummary.cs
+++ b/src/AWS.Deploy.CLI/ServerMode/Models/OptionSettingItemSummary.cs
@@ -3,47 +3,111 @@
 
 using System;
 using System.Collections.Generic;
+using AWS.Deploy.Common.Recipes;
 
 namespace AWS.Deploy.CLI.ServerMode.Models
 {
     public class OptionSettingItemSummary
     {
+        /// <summary>
+        /// The unique id of setting. This value will be persisted in other config files so its value should never change once a recipe is released.
+        /// </summary>
         public string Id { get; set; }
 
+        /// <summary>
+        /// The fully qualified id of the setting that includes the Id and all of the parent's Ids.
+        /// This helps easily reference the Option Setting without context of the parent setting.
+        /// </summary>
+        public string FullyQualifiedId { get; set; }
+
+        /// <summary>
+        /// The display friendly name of the setting.
+        /// </summary>
         public string Name { get; set; }
 
+        /// <summary>
+        /// The category for the setting. This value must match an id field in the list of categories.
+        /// </summary>
         public string? Category { get; set; }
 
+        /// <summary>
+        /// The description of what the setting is used for.
+        /// </summary>
         public string Description { get; set; }
 
+        /// <summary>
+        /// The value used for the recipe if it is set by the user.
+        /// </summary>
         public object? Value { get; set; }
 
+        /// <summary>
+        /// The type of primitive value expected for this setting.
+        /// For example String, Int
+        /// </summary>
         public string Type { get; set; }
 
+        /// <summary>
+        /// Hint the the UI what type of setting this is optionally add additional UI features to select a value.
+        /// For example a value of BeanstalkApplication tells the UI it can display the list of available Beanstalk applications for the user to pick from.
+        /// </summary>
         public string? TypeHint { get; set; }
 
+        /// <summary>
+        /// Type hint additional data required to facilitate handling of the option setting.
+        /// </summary>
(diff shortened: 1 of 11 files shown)


normj (changes requested):

src/AWS.Deploy.Common/Recipes/OptionSettingItem.ValueOverride.cs
```diff
@@ -109,7 +109,16 @@ public async Task SetValue(IOptionSettingHandler optionSettingHandler, object va
                     }
                 }
                 if (!isValid)
-                    throw new ValidationFailedException(DeployToolErrorCode.OptionSettingItemValueValidationFailed, validationFailedMessage.Trim());
+                {
+                    Validation.ValidationStatus = ValidationStatus.Invalid;
```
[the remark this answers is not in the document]
philasmar: Fixed

src/AWS.Deploy.CLI/ServerMode/Models/OptionSettingItemSummary.cs
```diff
@@ -3,13 +3,16 @@
...
 using System.Collections.Generic;
+using AWS.Deploy.Common.Recipes;
 
 namespace AWS.Deploy.CLI.ServerMode.Models
 {
     public class OptionSettingItemSummary
     {
         public string Id { get; set; }
 
+        public string FullyQualifiedId { get; set; }
```
ashovlin, line 14: Could you document this new property? Unsure how it's different from `Id`?

Given other Toolkits besides Visual Studio could potentially wrap server mode in the future, think it'd be nice to document as much as possible while we're here.

src/AWS.Deploy.Common/Recipes/OptionSettingItem.cs
```diff
@@ -160,13 +160,16 @@ public partial class OptionSettingItem : IOptionSettingItem
         /// Indicates which option setting items need to be validated if a value update occurs.
         /// </summary>
         public List<string> Dependents { get; set; } = new List<string> ();
+        
+        public  OptionSettingValidation Validation { get; set; }
```
ashovlin, line 164: Given this class is already in pretty good shape - could you document this?

test/AWS.Deploy.CLI.IntegrationTests/ServerMode/DependencyValidationOptionSettings.cs
```diff
@@ -0,0 +1,207 @@
...
+                Assert.True(string.IsNullOrEmpty(validRoleArn.Validation.ValidationMessage));
+            }
+            finally
+            {
+                cancelSource.Cancel();
+                _stackName = null;
+            }
+        }
+
+        private async Task WaitTillServerModeReady(RestAPIClient restApiClient)
```
ashovlin, line 133: This is already duplicated several times, could we move to something shared?

More generally, I think it'd be nice if server mode tests had less boilerplate so it'd be easier to see what the test is actually doing. Though not necessarily introduced by this PR, so okay if we move to a new issue.

ashovlin (changes requested):

src/AWS.Deploy.Recipes/RecipeDefinitions/ASP.NETAppElasticBeanstalk.recipe
```diff
@@ -340,7 +340,6 @@
                             "ValidatorType": "Regex",
                             "Configuration": {
                                 "Regex": "arn:.+:iam::[0-9]{12}:.+",
-                                "AllowEmptyString": true,
```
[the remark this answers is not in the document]
philasmar: This is not a breaking change as anyone that had this as empty had a deployment failure and thus their settings were not persisted.

src/AWS.Deploy.CLI/ServerMode/Models/OptionSettingItemSummary.cs
philasmar: Done. I've updated the entire class based on OptionSettingItem

src/AWS.Deploy.Common/Recipes/OptionSettingItem.cs
philasmar: Done

test/AWS.Deploy.CLI.IntegrationTests/ServerMode/DependencyValidationOptionSettings.cs
philasmar: Done

normj (approved):

PR #585, opened 2022-06-01
feat: persist invalid value in validation object of option setting item

philasmar:
*Description of changes:*
Persist invalid value set by the user in the validation object of the option setting item. This is used mainly in the toolkit from server mode where invalid values can remain on the UI until users replace them with valid settings.

### DIFF
--- a/src/AWS.Deploy.Common/Recipes/OptionSettingItem.ValueOverride.cs
+++ b/src/AWS.Deploy.Common/Recipes/OptionSettingItem.ValueOverride.cs
@@ -102,18 +102,26 @@ namespace AWS.Deploy.Common.Recipes
                     var result = await validator.Validate(valueOverride);
                     if (!result.IsValid)
                     {
-                        throw new ValidationFailedException(DeployToolErrorCode.OptionSettingItemValueValidationFailed,
-                            result.ValidationFailedMessage?.Trim() ?? $"The value '{valueOverride}' is invalid for option setting '{Name}'.");
+                        Validation.ValidationStatus = ValidationStatus.Invalid;
+                        Validation.ValidationMessage = result.ValidationFailedMessage?.Trim() ?? $"The value '{valueOverride}' is invalid for option setting '{Name}'.";
+                        Validation.InvalidValue = valueOverride;
+                        throw new ValidationFailedException(DeployToolErrorCode.OptionSettingItemValueValidationFailed, Validation.ValidationMessage);
                     }
                 }
-                
-                Validation.ValidationStatus = ValidationStatus.Valid;
-                Validation.ValidationMessage = string.Empty;
             }
 
             if (AllowedValues != null && AllowedValues.Count > 0 && valueOverride != null &&
                 !AllowedValues.Contains(valueOverride.ToString() ?? ""))
-                throw new InvalidOverrideValueException(DeployToolErrorCode.InvalidValueForOptionSettingItem, $"Invalid value for option setting item '{Name}'");
+            {
+                Validation.ValidationStatus = ValidationStatus.Invalid;
+                Validation.ValidationMessage = $"Invalid value for option setting item '{Name}'";
+                Validation.InvalidValue = valueOverride;
+                throw new InvalidOverrideValueException(DeployToolErrorCode.InvalidValueForOptionSettingItem, Validation.ValidationMessage);
+            }
+
+            Validation.ValidationStatus = ValidationStatus.Valid;
+            Validation.ValidationMessage = string.Empty;
+            Validation.InvalidValue = null;
 
             if (valueOverride is bool || valueOverride is int || valueOverride is long || valueOverride is double || valueOverride is Dictionary<string, string> || valueOverride is SortedSet<string>)
             {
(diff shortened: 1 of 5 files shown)


normj (changes requested):
There need to be test coverage updates to handle this new field. Test that new field is set and cleared for either the SetValue and the "DependsOn" changes.

src/AWS.Deploy.Common/Recipes/OptionSettingItem.ValueOverride.cs
```diff
@@ -102,13 +102,16 @@ public async Task SetValue(IOptionSettingHandler optionSettingHandler, object va
                     var result = await validator.Validate(valueOverride);
                     if (!result.IsValid)
                     {
-                        throw new ValidationFailedException(DeployToolErrorCode.OptionSettingItemValueValidationFailed,
-                            result.ValidationFailedMessage?.Trim() ?? $"The value '{valueOverride}' is invalid for option setting '{Name}'.");
+                        Validation.ValidationStatus = ValidationStatus.Invalid;
```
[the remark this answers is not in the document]
philasmar: The validation object is basically maintaining the state of the last set value attempt. Since we are not blocking the deployment based on the validation object status then we should be fine. Though for the Toolkit, the Validation object can be considered as the source of truth since it will be mirroring what the user has on the UI.

src/AWS.Deploy.Common/Recipes/OptionSettingValidation.cs
```diff
@@ -11,8 +11,21 @@ public enum ValidationStatus
...
+        /// The validation message in the case where the value set by the user is an invalid one.
+        /// This is empty in the case where the value set by the user is a valid one.
+        /// </summary>
         public string ValidationMessage { get; set; } = string.Empty;
+
+        /// <summary>
+        /// The value last attempted to be set by the user in the case where that value is invalid.
+        /// This is null in the case where the value is valid.
+        /// </summary>
+        public object? InvalidValue { get; set; }
```
ashovlin, line 29: Curious on @awschristou's take on this if he gets a chance - is it awkward that any wrapper tools will have to do something like this?
```
if (optionSettingItemSummary.Validation.ValidationStatus == Valid)
   Display optionSettingItemSummary.Value
else
  Display optionSettingItemSummary.Validation.InvalidValue

Should the server be more opinionated about _"if you are showing a form to the user, here is the value to show"_?

File: src/AWS.Deploy.Common/Recipes/OptionSettingItem.ValueOverride.cs [hunk 3]
```diff
@@ -102,13 +102,16 @@ public async Task SetValue(IOptionSettingHandler optionSettingHandler, object va
...
+                        throw new ValidationFailedException(DeployToolErrorCode.OptionSettingItemValueValidationFailed, Validation.ValidationMessage);
                     }
                 }
                 
                 Validation.ValidationStatus = ValidationStatus.Valid;
                 Validation.ValidationMessage = string.Empty;
+                Validation.InvalidValue = null;
             }
 
             if (AllowedValues != null && AllowedValues.Count > 0 && valueOverride != null &&
```
ashovlin, line 117: Should an invalid value due to the `AllowedValues` list also set the new property?

And it's not this code path, but similarly wondering about recipe validators...

src/AWS.Deploy.Orchestration/OptionSettingHandler.cs
```diff
@@ -48,15 +49,19 @@ public List<ValidationResult> RunOptionSettingValidators(Recommendation recommen
...
+                    if (settingValidatorFailedResults.Any())
+                    {
+                        optionSetting.Validation.ValidationStatus = ValidationStatus.Invalid;
+                        optionSetting.Validation.ValidationMessage = string.Join(Environment.NewLine, settingValidatorFailedResults.Select(x => x.ValidationFailedMessage)).Trim();
+                        optionSetting.Validation.InvalidValue = optionSettingValue;
+                    }
+                    else
+                    {
+                        optionSetting.Validation.ValidationStatus = ValidationStatus.Valid;
+                        optionSetting.Validation.ValidationMessage = string.Empty;
```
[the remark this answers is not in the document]
ashovlin: Line 52 isn't clear to me - when will be be here with `InvalidValue` already set?

----
If this is doing something like _"only update the validation status if it wasn't previously already invalid"_, I'd rather check the previous `ValidationStatus` instead of the presence of an `InvalidValue` - what if the reason it was invalid is because `null` is the invalid value?

src/AWS.Deploy.Common/Recipes/OptionSettingItem.ValueOverride.cs
philasmar: Fixed

src/AWS.Deploy.Common/Recipes/OptionSettingValidation.cs
awschristou: I agree it is good for clients to have access to both values (last-valid and current-invalid). In the current Toolkit implementation, the "last good" value will be ignored, but this PR's design allows clients to stay in sync with the backend. This is why I don't mind the conditional check mentioned by @ashovlin .

Both the clients and the backend shouldn't allow a deployment if there are invalid values as well.

ashovlin (approved):

PR #586, opened 2022-06-01
fix: Add validation for EC2 instance types for Elastic Beanstalk recipes

ashovlin:
*Issue #, if available:* DOTNET-5894

*Description of changes:* This adds an OptionSettingItemValidator for the EC2 Instance Type on the Elastic Beanstalk recipe.

### Before
A user in the VS Toolkit could skip the modal dialog to select from the full list, they are able to enter freetext which may be invalid. This would attempt to deploy and then fail.
image

### After

image

This also validates that a seemingly well-formed instance type that is not available in the deployment region (here ap-northeast-3):
image

### DIFF
--- a/src/AWS.Deploy.Common/Recipes/Validation/OptionSettingItemValidators/InstanceTypeValidator.cs
+++ b/src/AWS.Deploy.Common/Recipes/Validation/OptionSettingItemValidators/InstanceTypeValidator.cs
@@ -1,0 +1,61 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+using System;
+using System.Threading.Tasks;
+using Amazon.EC2;
+using Amazon.EC2.Model;
+using AWS.Deploy.Common.Data;
+
+namespace AWS.Deploy.Common.Recipes.Validation
+{
+    /// <summary>
+    /// Validates that a given EC2 instance is valid for the deployment region
+    /// </summary>
+    public class InstanceTypeValidator : IOptionSettingItemValidator
+    {
+        private readonly IAWSResourceQueryer _awsResourceQueryer;
+
+        public InstanceTypeValidator(IAWSResourceQueryer awsResourceQueryer)
+        {
+            _awsResourceQueryer = awsResourceQueryer;
+        }
+
+        public async Task<ValidationResult> Validate(object input, Recommendation recommendation)
+        {
+            var rawInstanceType = Convert.ToString(input);
+            InstanceTypeInfo? instanceTypeInfo;
+
+            if (string.IsNullOrEmpty(rawInstanceType))
+            {
+                return ValidationResult.Valid();
+            }
+
+            try
+            {
+                instanceTypeInfo = await _awsResourceQueryer.DescribeInstanceType(rawInstanceType);
+            }
+            catch (ResourceQueryException ex)
+            {
+                // Check for the expected exception if the provided instance type is invalid
+                if (ex.InnerException is AmazonEC2Exception ec2Exception &&
+                    string.Equals(ec2Exception.ErrorCode, "InvalidInstanceType", StringComparison.InvariantCultureIgnoreCase))
+                {
+                    instanceTypeInfo = null;
+                }
+                else // Anything else is unexpected, so proceed with usual exception handling
+                {
+                   throw ex;
+                }
+            }
+            if (instanceTypeInfo != null)
+            {
+                return ValidationResult.Valid();
+            }
+            else
+            {
+                return ValidationResult.Failed($"The specified instance type {rawInstanceType} does not exist in the deployment region.");
+            }
+        }
+    }
+}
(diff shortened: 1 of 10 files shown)


normj (approved):

src/AWS.Deploy.Common/Recipes/Validation/OptionSettingItemValidators/InstanceTypeValidator.cs
```diff
@@ -0,0 +1,42 @@
...
+        public InstanceTypeValidator(IAWSResourceQueryer awsResourceQueryer)
+        {
+            _awsResourceQueryer = awsResourceQueryer;
+        }
+
+        public async Task<ValidationResult> Validate(object input, Recommendation recommendation)
+        {
+            var rawInstanceType = Convert.ToString(input);
+
+            // Instance type isn't required, so treat empty values as valid
```
philasmar, line 26: nit: I think the reason why this validator should return valid is not because of the instance type not being necessary in this case, but because we could add a Required validator if it was. I think it's better that we remove the comment.

src/AWS.Deploy.Orchestration/Data/AWSResourceQueryer.cs
```diff
@@ -167,6 +167,46 @@ await foreach (var response in listInstanceTypesPaginator.Responses)
...
+                    // Check for the expected exception if the provided instance type is invalid
+                    if (string.Equals(ex.ErrorCode, "InvalidInstanceType", StringComparison.InvariantCultureIgnoreCase))
+                    {
+                        return null;
+                    }
+                    else
+                    {   // Anything else is unexpected, so proceed with usual exception handling
+                        throw ex;
+                    }
+                }
```
philasmar, line 206: this logic should be added in the validator and not here. We shouldn't add business logic in the AWSResourceQueryer.  Let the service call error out if the instance type isn't found and catch that error in the validator.

src/AWS.Deploy.Orchestration/Data/AWSResourceQueryer.cs
```diff
@@ -167,6 +167,46 @@ await foreach (var response in listInstanceTypesPaginator.Responses)
...
+                    var response = await ec2Client.DescribeInstanceTypesAsync(request);
+
+                    if (response.InstanceTypes?.Count >= 1)
+                    {
+                        return response.InstanceTypes[0];
+                    }
+                    else
+                    {
+                        return null;
+                    }
```
philasmar, line 192: nit: could you do this using Linq? (return response.InstanceTypes.FirstOrDefault();)

philasmar (changes requested):

src/AWS.Deploy.Orchestration/Data/AWSResourceQueryer.cs
ashovlin: Why do we not want to add business logic in AWSResourceQueryer? There's no documentation on the interface. Could you point me to a design doc with intent?

----

We have similar logic around resource existence already in:
1. GetECRRepositories - https://github.com/aws/aws-dotnet-deploy/blob/e2ad9b5a026869aa8854182dd5df1f4cedc6fa46/src/AWS.Deploy.Orchestration/Data/AWSResourceQueryer.cs#L533-L536
2. GetCloudFormationStack - https://github.com/aws/aws-dotnet-deploy/blob/e2ad9b5a026869aa8854182dd5df1f4cedc6fa46/src/AWS.Deploy.Orchestration/Data/AWSResourceQueryer.cs#L573-L577
3. DescribeECRRepository - https://github.com/aws/aws-dotnet-deploy/blob/e2ad9b5a026869aa8854182dd5df1f4cedc6fa46/src/AWS.Deploy.Orchestration/Data/AWSResourceQueryer.cs#L711-L714
4. ParameterNotFoundException - https://github.com/aws/aws-dotnet-deploy/blob/e2ad9b5a026869aa8854182dd5df1f4cedc6fa46/src/AWS.Deploy.Orchestration/Data/AWSResourceQueryer.cs#L731-L734

Should I rework all of those too?

src/AWS.Deploy.Common/Recipes/Validation/OptionSettingItemValidators/InstanceTypeValidator.cs
ashovlin: Dropped the comment.

src/AWS.Deploy.Orchestration/Data/AWSResourceQueryer.cs
ashovlin: Done

src/AWS.Deploy.Orchestration/Data/AWSResourceQueryer.cs
philasmar: We don't have that documented anywhere, but since AWSResourceQueryer is basically doing some SDK calls to retrieve resources, I'd like to leave them as raw as possible to promote reusability of those methods. Any needed business logic can be added in the caller method. As for the other examples where we are not doing that, I usually update the methods whenever I am working on a change that touches those methods. So I wouldn't update the rest in this PR to avoid the scope creep.

96malhar (approved):

src/AWS.Deploy.Orchestration/Data/AWSResourceQueryer.cs
ashovlin: Moved the exception handling to the validator, and added a note on the interface for the future.

philasmar (approved):

PR #590, opened 2022-06-03
feat: Update supported target frameworks for recipes

normj:
*Description of changes:*
For all container based recipes add .NET 7 to the list of supported target frameworks and include the .NET 7 base images to the framework to image mapping.

Also removed .NET Core 2.1 from the list since the version is well out of support and no reason to GA with that support in the tool.

### DIFF
--- a/src/AWS.Deploy.DockerEngine/Properties/DockerFileConfig.json
+++ b/src/AWS.Deploy.DockerEngine/Properties/DockerFileConfig.json
@@ -3,6 +3,11 @@
         "SdkType": "Microsoft.NET.Sdk.Web",
         "ImageMapping": [
             {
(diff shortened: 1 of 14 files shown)


philasmar (approved):

normj:
> Changes look okay, but could you also ensure we've kicked off the work to remove any .NET Core 2.1 language from https://docs.aws.amazon.com/sdk-for-net/v3/developer-guide/deployment-tool.html
> 
> Might also be nice to use the BREAKING CHANGE syntax in the commit message to call this out in the changelog.

Added BREAKING CHANGE to commit message. The new docs say we support .NET Core 3.1 or later.

ashovlin (approved):

PR #595, opened 2022-06-06
feat: add validators to check if Subnets and Security Groups are part of the same VPC

philasmar:
*Description of changes:*
This change is to fix a behavior in the toolkit where a user selects subnets and security groups and then changes the selected VPC which invalidates the already selected subnets and security groups. This adds a validator that checks if the selected subnets and security groups are part of the selected vpc.

### DIFF
--- a/src/AWS.Deploy.Common/Recipes/Validation/OptionSettingItemValidators/SecurityGroupsInVpcValidator.cs
+++ b/src/AWS.Deploy.Common/Recipes/Validation/OptionSettingItemValidators/SecurityGroupsInVpcValidator.cs
@@ -1,0 +1,59 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+using System;
+using System.Linq;
+using System.Collections.Generic;
+using System.Threading.Tasks;
+using AWS.Deploy.Common.Data;
+using AWS.Deploy.Common.Extensions;
+
+namespace AWS.Deploy.Common.Recipes.Validation
+{
+    /// <summary>
+    /// Validates that the selected security groups are part of the selected VPC
+    /// </summary>
+    public class SecurityGroupsInVpcValidator : IOptionSettingItemValidator
+    {
+        private static readonly string defaultValidationFailedMessage = "The selected security groups are not part of the selected VPC.";
+        public string VpcId { get; set; } = "";
+        public string ValidationFailedMessage { get; set; } = defaultValidationFailedMessage;
+
+        private readonly IAWSResourceQueryer _awsResourceQueryer;
+        private readonly IOptionSettingHandler _optionSettingHandler;
+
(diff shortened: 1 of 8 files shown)


src/AWS.Deploy.Common/Recipes/Validation/OptionSettingItemValidators/SecurityGroupsInVpcValidator.cs
```diff
@@ -0,0 +1,59 @@
...
+            if (string.IsNullOrEmpty(vpcId))
+                return ValidationResult.Failed("The VpcId setting is not set or is empty. Make sure to set the VPC Id first.");
+
+            var securityGroupIds = (await _awsResourceQueryer.DescribeSecurityGroups(vpcId)).Select(x => x.GroupId);
+            if (input?.TryDeserialize<SortedSet<string>>(out var inputList) ?? false)
+            {
+                foreach (var securityGroup in inputList!)
+                {
+                    if (!securityGroupIds.Contains(securityGroup))
+                        return ValidationResult.Failed("The selected security group(s) are invalid since they do not belong to the currently selected VPC.");
```
96malhar, line 46: Should we be explicit about which SecurityGroups are invalid?

src/AWS.Deploy.Common/Recipes/Validation/OptionSettingItemValidators/SubnetsInVpcValidator.cs
```diff
@@ -0,0 +1,61 @@
...
+            if (string.IsNullOrEmpty(vpcId))
+                return ValidationResult.Failed("The VpcId setting is not set or is empty. Make sure to set the VPC Id first.");
+
+            var subnetIds = (await _awsResourceQueryer.DescribeSubnets(vpcId)).Select(x => x.SubnetId);
+            if (input?.TryDeserialize<SortedSet<string>>(out var inputList) ?? false)
+            {
+                foreach (var subnet in inputList!)
+                {
+                    if (!subnetIds.Contains(subnet))
+                        return ValidationResult.Failed("The selected subnet(s) are invalid since they do not belong to the currently selected VPC.");
```
96malhar, line 48: Should we be explicit about which Subnets are invalid?

src/AWS.Deploy.Common/Recipes/Validation/OptionSettingItemValidators/SecurityGroupsInVpcValidator.cs
philasmar: No need since users can't mix and match security groups from different VPCs. It's either they are all invalid or all valid.

src/AWS.Deploy.Common/Recipes/Validation/OptionSettingItemValidators/SubnetsInVpcValidator.cs
philasmar: No need since users can't mix and match subnets from different VPCs. It's either they are all invalid or all valid.

96malhar (approved):

normj (approved):

PR #597, opened 2022-06-06
fix: Fixed issue with deserialize types that do not have a parameterless constructor.

normj:
*Description of changes:*
In PR https://github.com/aws/aws-dotnet-deploy/pull/594 the CLI dependency to `AWS.Deploy.Recipes.CDK.Common` was removed to reduce package size. This had an unintended side effect of downgrading the version of `System.Text.Json` to version 4.7.2 from 5.0.0. Version 4.7.2 requires types that are deserialized to have a parameterless constructor. Many of the model types in ServerMode do not have parameterless constructor.

The fix is to force the CLI to use a newer version of System.Text.Json and add a test to make sure the explicit reference is not removed and allow fallback to a too old version.


philasmar (approved):

src/AWS.Deploy.CLI/AWS.Deploy.CLI.csproj
```diff
@@ -26,6 +26,7 @@
     <PackageReference Include="Swashbuckle.AspNetCore.Annotations" Version="6.1.2" />
     <PackageReference Include="Swashbuckle.AspNetCore.Swagger" Version="6.1.2" />
     <PackageReference Include="System.CommandLine" Version="2.0.0-beta1.20574.7" />
+    <PackageReference Include="System.Text.Json" Version="6.0.4" />
```
ashovlin, line 29: Could you update the version in THIRD_PARTY_LICENSES and the internal dependency tracking tool also?

ashovlin (changes requested):

src/AWS.Deploy.CLI/AWS.Deploy.CLI.csproj
normj: Done

ashovlin (approved):

normj:
Merging without waiting for CI build since the last change was just changing `THIRD_PARTY_LICENSES`

PR #598, opened 2022-06-07
feat: ApplyConfigSettings API now rejects updating readonly settings

philasmar:
*Issue #, if available:*
DOTNET-5993

*Description of changes:*
ApplyConfigSettings API now rejects updating readonly settings

### DIFF
--- a/test/AWS.Deploy.CLI.IntegrationTests/ServerModeTests.cs
+++ b/test/AWS.Deploy.CLI.IntegrationTests/ServerModeTests.cs
@@ -2,6 +2,7 @@
 // SPDX-License-Identifier: Apache-2.0
 
 using System;
+using System.Collections.Generic;
 using System.IO;
 using System.Linq;
 using System.Security.Cryptography;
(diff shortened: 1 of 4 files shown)


normj (approved):

test/AWS.Deploy.CLI.IntegrationTests/ServerModeTests.cs
```diff
@@ -244,6 +245,29 @@ public async Task WebFargateDeploymentNoConfigChanges()
                 Assert.Contains(existingDeployment.SettingsCategories, x => string.Equals(x.Id, AWS.Deploy.Common.Recipes.Category.DeploymentBundle.Id));
                 Assert.DoesNotContain(existingDeployment.SettingsCategories, x => string.IsNullOrEmpty(x.Id));
                 Assert.DoesNotContain(existingDeployment.SettingsCategories, x => string.IsNullOrEmpty(x.DisplayName));
+
+                await restClient.SetDeploymentTargetAsync(redeploymentSessionId, new SetDeploymentTargetInput
```
ashovlin, line 249: Tests should fail for one reason, it's not even clear to me from the naming or comments what this test was originally testing.

Could you either move this to something like `UpdatingReadOnlySettingsIsRejected` test, or at least add comments summarizing what each block of `Assserts` is looking for?

ashovlin (changes requested):

test/AWS.Deploy.CLI.IntegrationTests/ServerModeTests.cs
philasmar: Since updating readonly settings is only applicable for redeployment, I need a test that performs a successful deployment and then attempts to update readonly settings for redeployment. I didn't create a new test as we don't want to expand the number of integ tests and thus the time it takes for our tests to run especially when these tests are very similar. This test originally just did a fargate deployment using server mode. We started to add more assertions  to this test as we added more features. I will add comments explaining my reasoning.

ashovlin (approved):

PR #599, opened 2022-06-07
fix: Do not validate against terminated Beanstalk environments

96malhar:
*Description of changes:*
`_awsResourceQueryer.ListOfElasticBeanstalkEnvironments(...)` can return environments that were freshly terminated. 
If a user tries to deploy with an environment name that was used by a previously terminated environment, then the `ExistingResourceValidator` fails.

This PR prevents validating against terminated Beanstalk environments.

### DIFF
--- a/src/AWS.Deploy.Common/Recipes/Validation/OptionSettingItemValidators/ExistingResourceValidator.cs
+++ b/src/AWS.Deploy.Common/Recipes/Validation/OptionSettingItemValidators/ExistingResourceValidator.cs
@@ -5,6 +5,7 @@ using System.Linq;
 using System.Threading.Tasks;
 using Amazon.CloudControlApi.Model;
 using AWS.Deploy.Common.Data;
+using Amazon.ElasticBeanstalk;
 
 namespace AWS.Deploy.Common.Recipes.Validation
 {
@@ -40,7 +41,7 @@ namespace AWS.Deploy.Common.Recipes.Validation
 
                 case "AWS::ElasticBeanstalk::Environment":
                     var beanstalkEnvironments = await _awsResourceQueryer.ListOfElasticBeanstalkEnvironments(environmentName: resourceName);
-                    if (beanstalkEnvironments.Any(x => x.EnvironmentName.Equals(resourceName)))
+                    if (beanstalkEnvironments.Any(x => x.EnvironmentName.Equals(resourceName) && x.Status != EnvironmentStatus.Terminated))
                         return ValidationResult.Failed($"An Elastic Beanstalk environment already exists with the name '{resourceName}'. Check the AWS Console for more information on the existing resource.");
                     break;
 


philasmar (changes requested):
Why are we doing this? A terminated environment is still available in the console for a short while before the name can be reused for another environment. @normj any thoughts?

96malhar:
> Why are we doing this? A terminated environment is still available in the console for a short while before the name can be reused for another environment. @normj any thoughts?

I have consistently seen that whenever a CF stack is deleted, the Beanstalk environment immediately goes away from the console. But it does show up in the [DescribeEnvironments](https://awscli.amazonaws.com/v2/documentation/api/latest/reference/elasticbeanstalk/describe-environments.html) API call for about 20 mins after being terminated. 

The current state of the `ExistingResourceValidator` forces the customer to wait for sometime before they can perform deployments using the same environment name. IMO, this is undesirable as a terminated environment name is free to be used by a new environment.

philasmar:
> > Why are we doing this? A terminated environment is still available in the console for a short while before the name can be reused for another environment. @normj any thoughts?
> 
> I have consistently seen that whenever a CF stack is deleted, the Beanstalk environment immediately goes away from the console. But it does show up in the [DescribeEnvironments](https://awscli.amazonaws.com/v2/documentation/api/latest/reference/elasticbeanstalk/describe-environments.html) API call for about 20 mins after being terminated.
> 
> The current state of the `ExistingResourceValidator` forces the customer to wait for sometime before they can perform deployments using the same environment name. IMO, this is undesirable as a terminated environment name is free to be used by a new environment.

I just tested it and I didn't face an issue. Oddly, I remember having issues with deploying to terminated environments in the past. I agree with Alex on adding tests for that to mitigate the issue in case of a behavior change in the future.

96malhar:
> > > Why are we doing this? A terminated environment is still available in the console for a short while before the name can be reused for another environment. @normj any thoughts?
> > 
> > 
> > I have consistently seen that whenever a CF stack is deleted, the Beanstalk environment immediately goes away from the console. But it does show up in the [DescribeEnvironments](https://awscli.amazonaws.com/v2/documentation/api/latest/reference/elasticbeanstalk/describe-environments.html) API call for about 20 mins after being terminated.
> > The current state of the `ExistingResourceValidator` forces the customer to wait for sometime before they can perform deployments using the same environment name. IMO, this is undesirable as a terminated environment name is free to be used by a new environment.
> 
> I just tested it and I didn't face an issue. Oddly, I remember having issues with deploying to terminated environments in the past. I agree with Alex on adding tests for that to mitigate the issue in case of a behavior change in the future.

For now, I tested tested this manually multiple times and did not run into issues. 

Also, per [official AWS documentation](https://docs.aws.amazon.com/elasticbeanstalk/latest/dg/environment-management-rebuild.html), you can rebuild an environment within six weeks (42 days) of its termination. However, this rebuild is only possible if the corresponding application version is still available. In our case, when we teardown a CF stack, we delete the entire Beanstalk application (including all versions). Thus there is no possibility of rebuilding the environment

normj:
> > > > Why are we doing this? A terminated environment is still available in the console for a short while before the name can be reused for another environment. @normj any thoughts?
> > > 
> > > 
> > > I have consistently seen that whenever a CF stack is deleted, the Beanstalk environment immediately goes away from the console. But it does show up in the [DescribeEnvironments](https://awscli.amazonaws.com/v2/documentation/api/latest/reference/elasticbeanstalk/describe-environments.html) API call for about 20 mins after being terminated.
> > > The current state of the `ExistingResourceValidator` forces the customer to wait for sometime before they can perform deployments using the same environment name. IMO, this is undesirable as a terminated environment name is free to be used by a new environment.
> > 
> > 
> > I just tested it and I didn't face an issue. Oddly, I remember having issues with deploying to terminated environments in the past. I agree with Alex on adding tests for that to mitigate the issue in case of a behavior change in the future.
> 
> For now, I tested tested this manually multiple times and did not run into issues.
> 
> Also, per [official AWS documentation](https://docs.aws.amazon.com/elasticbeanstalk/latest/dg/environment-management-rebuild.html), you can rebuild an environment within six weeks (42 days) of its termination. However, this rebuild is only possible if the corresponding application version is still available. In our case, when we teardown a CF stack, we delete the entire Beanstalk application (including all versions). Thus there is no possibility of rebuilding the environment.
> 
> I can add an integ test for this but worried about the impact on our (already long) total test run time.

I agree with Malhar that if have validated this is the service behavior that an environment with the same name can be created as a just terminated environment we don't need to add a long and expensive integ test. It is highly unlikely that a service would make such a change in behavior that could break untold dev environment build setups.

ashovlin (approved):

philasmar (approved):

PR #602, opened 2022-06-08
feat: update VPCs in all recipes to match app runner recipe

philasmar:
*Description of changes:*
We recently updated the VPC Connector setting in App Runner which involved adding a default VPC for VPC ID, updating validators and adding dependencies. These updates have now been applied to the rest of the recipes where applicable.
For beanstalk recipe, this will cause a breaking change as the VPC setting has now been turned into an Object type which included the "UseVPC" setting which is set to false by default. So deployments already using a VPC will need to delete and recreate their stacks as the VPC setting is not updatable in beanstalk.
Additionally, child option settings "Updatable" value was being ignored if parent setting's "Updatable" is true. This change fixes the behavior.


ashovlin (approved):

normj (approved):
Make sure the release notes call out this is a breaking change.

PR #603, opened 2022-06-08
feat: Add support for returning tables of options via GetConfigSettingResources

ashovlin:
*Issue #, if available:* DOTNET-5907

*Description of changes:*
Adds support for returning tables of options via `GetConfigSettingResources`. This allows wrapping tools to display richer grids/tables when presenting the user a list of options to select from. 

Subnets and Security Groups for App Runner's VPC Connector are supported initially, but please leave a comment if there are any good candidates.


src/AWS.Deploy.CLI/Commands/TypeHints/ExistingSubnetsCommand.cs
```diff
@@ -37,25 +36,48 @@ private async Task<List<Subnet>> GetData(Recommendation recommendation, OptionSe
...
+                    new TypeHintResourceColumn("az", "Availability Zone")
+                }
+            };
+
+            foreach (var subnet in subnets.OrderBy(subnet => subnet.VpcId))
+            {
+                var row = new TypeHintResource(subnet.SubnetId, subnet.SubnetId);
+                row.ColumnValues["subnetId"] = subnet.SubnetId;
+                row.ColumnValues["vpcId"] = subnet.VpcId;
+                row.ColumnValues["az"] = subnet.AvailabilityZone;
```
philasmar, line 58: Based on my previous comment, if we get rid of the Id and just use the field name, can't we deduce the Column names based on the Keys in the dictionary? Making lines 45-50 obsolete?

src/AWS.Deploy.CLI/Commands/TypeHints/ExistingSecurityGroupsCommand.cs
```diff
@@ -37,30 +36,53 @@ private async Task<List<SecurityGroup>> GetData(Recommendation recommendation, O
...
+                resourceTable.Rows.Add(row);
+            }
+
+            return resourceTable;
         }
 
         public async Task<object> Execute(Recommendation recommendation, OptionSettingItem optionSetting)
         {
-            var availableSecurityGroups = (await GetData(recommendation, optionSetting)).OrderBy(x => x.VpcId).ToList();
+            var resourceTable = await GetResources(recommendation, optionSetting);
```
philasmar, line 68: I prefer if we use GetData and not GetResources, since we'd be doing extra work without a practical benefit. We'd need to loop over the security groups to create the resource table and then loop over the resource table to display the values.

src/AWS.Deploy.CLI/ServerMode/Models/TypeHintResourceColumn.cs
```diff
@@ -0,0 +1,28 @@
...
+    /// <summary>
+    /// Represents a column header for a list/grid of <see cref="TypeHintResourceSummary"/> rows
+    /// </summary>
+    public class TypeHintResourceColumn
+    {
+        /// <summary>
+        /// Internal Id for the column. This can be used as a key to access a specific
+        /// row's value via its <see cref="TypeHintResourceSummary.ColumnValues"/>
+        /// </summary>
+        public string Id { get; set; }
```
philasmar, line 15: Based on my earlier comment, I think we don't need this class.

src/AWS.Deploy.Common/TypeHintData/TypeHintResourceTable.cs
```diff
@@ -0,0 +1,26 @@
...
+    /// a list of existing resources when setting an OptionSettingItem
+    /// </summary>
+    public class TypeHintResourceTable
+    {
+        /// <summary>
+        /// Columns that should appear above the list of resources when presenting the
+        /// user a table or grid to select from
+        /// </summary>
+        /// <remarks>If this is null or empty, it implies that there is only a single column</remarks>
+        public List<TypeHintResourceColumn>? Columns { get; set; }
```
philasmar, line 19: I think we only need this to be a `List<string>?`

philasmar (changes requested):

src/AWS.Deploy.Common/TypeHintData/TypeHintResourceTable.cs
ashovlin: I wanted to model the columns as classes as opposed to strings. That way if we add additional properties in the future (width, sort order, etc.), it'd be an additive change to an existing class as opposed to a breaking change for consumers of server mode to transition from `List<string>` to `List<SomeNewClass>`.

We could use a simpler model for the CLI-specific classes instead of the `/ServerMode/` ones, but I opted to keep them as similar as possible.

src/AWS.Deploy.CLI/ServerMode/Models/TypeHintResourceColumn.cs
ashovlin: I wanted to model the columns as classes as opposed to strings. That way if we add additional properties in the future (width, sort order, etc.), it'd be an additive change to an existing class as opposed to a breaking change for consumers of server mode to transition from `List<string>` to `List<SomeNewClass>`.

We could use a simpler model for the CLI-specific classes instead of the `/ServerMode/` ones, but I opted to keep them as similar as possible.

src/AWS.Deploy.CLI/Commands/TypeHints/ExistingSecurityGroupsCommand.cs
ashovlin: I intentionally routed it through `GetResources`, in my opinion we shouldn't have two different code paths for _"prepare these resources for display"_ between CLI and server mode, that's partially what led to this task in the first place. I was tempted to rip out `GetData` entirely.

It's still one network call, I don't think we're at the point of optimizing the looping yet.

src/AWS.Deploy.Common/TypeHintData/TypeHintResourceColumn.cs
```diff
@@ -0,0 +1,28 @@
...
+    /// <summary>
+    /// Represents the column header for a list/grid of <see cref="TypeHintResource"/> rows
+    /// </summary>
+    public class TypeHintResourceColumn
+    {
+        /// <summary>
+        /// Internal Id for the column. This can be used as a key to access a specific
+        /// row's value via its <see cref="TypeHintResource.ColumnValues"/>
+        /// </summary>
+        public string Id { get; set; }
```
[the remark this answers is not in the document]
ashovlin: I liked the keys for:
* Readability, since `colums['foo']` describes which column you're accessing for a given row
* If we ever reorder columns, there are fewer code changes to make

But I was also slightly afraid of us mistyping a key and then getting a `KeyNotFound ` at runtime.

----

Dropped the Ids and dictionaries for now, and switched to lists with indices.

src/AWS.Deploy.CLI/Commands/TypeHints/ExistingSubnetsCommand.cs
ashovlin: My reply/fix to https://github.com/aws/aws-dotnet-deploy/pull/603#discussion_r898006860 kind of changed this one, but have dropped the Ids/dictionaries.

src/AWS.Deploy.CLI/Commands/TypeHints/DotnetBeanstalkPlatformArnCommand.cs
```diff
@@ -33,10 +33,16 @@ private async Task<List<PlatformSummary>> GetData()
...
 
-        public async Task<List<TypeHintResource>?> GetResources(Recommendation recommendation, OptionSettingItem optionSetting)
+        public async Task<TypeHintResourceTable> GetResources(Recommendation recommendation, OptionSettingItem optionSetting)
         {
             var platformArns = await GetData();
-            return platformArns.Select(x => new TypeHintResource(x.PlatformArn, $"{x.PlatformBranchName} v{x.PlatformVersion}")).ToList();
+
+            var resourceTable = new TypeHintResourceTable
+            {
+                Rows = platformArns.Select(x => new TypeHintResource(x.PlatformArn, $"{x.PlatformBranchName} v{x.PlatformVersion}")).ToList()
```
normj, line 42: Since we have columns support now you might as well put branch and version in separate columns.

normj (changes requested):
I like the change to return `TypeHintResourceTable` make sure you review the API change with the toolkit team to ensure it works for them.

philasmar (approved):

src/AWS.Deploy.CLI/Commands/TypeHints/DotnetBeanstalkPlatformArnCommand.cs
ashovlin: Added. I opted to keep the `$"{branch name} v{version}"` we have for the CLI, but we're now returning two columns for server mode.

normj (approved):

ashovlin:
> I like the change to return `TypeHintResourceTable` make sure you review the API change with the toolkit team to ensure it works for them.

I had reviewed an internal doc with them originally, but just ran past @awschristou again via Slack since the design had changed slightly.

PR #604, opened 2022-06-09
fix: recipe validators are not run in server mode

philasmar:
*Issue #, if available:*
DOTNET-5926

*Description of changes:*
This change runs the recipe validators in the StartDeployment API.


philasmar:
> Change looks good.
> 
> Might be nice to add a test (maybe in `GetApplyOptionSettings`?) to verify that this is left in place.

I added a test.

normj (approved):

PR #607, opened 2022-06-10
feat: added validator to check if health check interval is greater than timeout

philasmar:
*Issue #, if available:*
DOTNET-5893

*Description of changes:*
Added a Comparison validator to check if health check interval is greater than timeout.
Added a Health Check Timeout setting to the Fargate recipe to allow users to change it's value.


src/AWS.Deploy.Common/Recipes/Validation/OptionSettingItemValidators/ComparisonValidator.cs
```diff
@@ -0,0 +1,56 @@
...
+    {
+        GreaterThan
+    }
+
+    /// <summary>
+    /// The validator is typically used with OptionSettingItems which have a numeric type.
+    /// The validator requires two configuration options to be specific, the Operation and the SettingId.
+    /// The validator checks if the set value of the OptionSettingItem satisfies the comparison operation with SettingId.
+    /// </summary>
+    public class ComparisonValidator : IOptionSettingItemValidator
```
ashovlin, line 18: This seems very similar to [MinMaxConstraintValidator.cs](https://github.com/aws/aws-dotnet-deploy/blob/main/src/AWS.Deploy.Common/Recipes/Validation/RecipeValidators/MinMaxConstraintValidator.cs).

1. Could we reuse the existing validator here?
2. Or if this new one is preferred, could you also replace MinMax?

philasmar: I removed the MinMaxConstraint validator as it can be replaced by the new one.

src/AWS.Deploy.Recipes/RecipeDefinitions/aws-deploy-recipe-schema.json
```diff
@@ -352,7 +352,6 @@
                         "type": "string",
                         "enum": [
                             "FargateTaskSizeCpuMemoryLimits",
-                            "MinMaxConstraint",
```
ashovlin, line 355: I can't comment on it directly, but could you also remove the related block starting at line 383?

And since "Comparison" has properties, can you add the property typing block to the schema?

philasmar: Done

PR #613, opened 2022-06-15
fix: Fixed environment variables not being set correctly in AWS App Runner deployments

normj:
*Description of changes:*
The App Runner recipe allows users to configure environment variables but the environment variables were not being correctly being set on the App Runner CDK construct causing them to be ignored. The CDK project was passing the environment variables as a `List` but it needed to be an array.


ashovlin (approved):
Nice test design 👍

Issue #579, opened 2022-05-31
Publish failures when user profile path contains a space


awschristou:
**Describe the bug**
A bug was reported on the Publish to AWS experience in the AWS Toolkit for Visual Studio. Deployments will fail for (Windows) users with a space in their user profile path.

Issue: https://github.com/aws/aws-toolkit-visual-studio/issues/255

**To Reproduce**
1. Have a Windows user profile path containing a space (eg: `c:\users\foo bar\`)
2. Open a deployable project in Visual Studio
3. Right click on the project in solution explorer, select "Publish to AWS" 
4. Try to publish the project

**Logs**

```

Generating AWS Cloud Development Kit (AWS CDK) deployment project
Using bootstrapping template from C:\\Users\\Marco Ma\\.aws-dotnet-deploy\\CDKBootstrapTemplate.yaml
Unhandled exception. Amazon.JSII.Runtime.JsiiException: Child process exited unexpectedly!
   at Amazon.JSII.Runtime.Services.Runtime.ReadResponse()
   at Amazon.JSII.Runtime.Services.Client.ReceiveResponse[TResponse]()
   at Amazon.JSII.Runtime.Services.Client.Hello()
   at Amazon.JSII.Runtime.Services.ServiceContainer.BuildServiceProvider(ILoggerFactory loggerFactoryOverride)
   at Amazon.JSII.Runtime.Services.ServiceContainer.<>c.<.cctor>b__8_0()
   at System.Lazy`1.ViaFactory(LazyThreadSafetyMode mode)
   at System.Lazy`1.ExecutionAndPublication(LazyHelper executionAndPublication, Boolean useDefaultConstructor)
   at System.Lazy`1.CreateValue()
   at System.Lazy`1.get_Value()
   at Amazon.JSII.Runtime.Services.ServiceContainer.get_ServiceProvider()
   at Amazon.JSII.Runtime.Deputy.JsiiTypeAttributeBase.Load(Assembly assembly)
   at Amazon.JSII.Runtime.Deputy.JsiiTypeAttributeBase.Load(Assembly assembly)
   at Amazon.JSII.Runtime.Deputy.DeputyBase..ctor(DeputyProps props)
   at Constructs.Construct..ctor(DeputyProps props)
   at Amazon.CDK.Stage..ctor(DeputyProps props)
   at Amazon.CDK.App..ctor(IAppProps props)
[...]

PR #617, opened 2022-06-15
feat: Allow users to override the default deploy tool workspace directory

96malhar:
*Issue #, if available:*
DOTNET-6010

*Description of changes:*
The PR gives users the ability to set their custom deploy tool workspace directory. They can do so by setting a user scoped or machine scoped environment variable called `AWS_DOTNET_DEPLOYTOOL_WORKSPACE`. This variable must point to a directory that exists on the disk and does not contain any whitespace characters.

This PR also solves the issue of failed deployments when the user profile path contains a whitespace character - https://github.com/aws/aws-dotnet-deploy/issues/579
The underlying issue is with CDK and not with our deployment tool - https://github.com/aws/aws-cdk/issues/2532 and https://github.com/aws/jsii/issues/1296

I will create separate PR to update the documentation and the troubleshooting guide. I am unsure of how to cover this feature with an integration test as it would require `AWS_DOTNET_DEPLOYTOOL_WORKSPACE` to exist as a `User` or `Machine` scoped environment variable. But since our tests run in parallel, it will be caught up by other unintended integ tests too.

-----------

**Without the setting the `AWS_DOTNET_DEPLOYTOOL_WORKSPACE` environment variable**

Capture
---------------

**The same deployment is successful after the `AWS_DOTNET_DEPLOYTOOL_WORKSPACE` environment variable**

Capture
----------


src/AWS.Deploy.Orchestration/Utilities/EnvironmentVariableManager.cs
```diff
@@ -0,0 +1,36 @@
...
+
+        /// <summary>
+        /// Stores the environment variable with the specified value. The variable is scoped to the current process
+        /// </summary>
+        void SetEnvironmentVariable(string variable, string? value);
+    }
+
+    public class EnvironmentVariableManager : IEnvironmentVariableManager
+    {
+        public string? GetEnvironmentVariable(string variable, EnvironmentVariableTarget target)
```
normj, line 27: I don't think we need to worry about `EnvironmentVariableTarget`. You can just call the `GetEnvironmentVariable` that doesn't take that variable which will work for both User and Machine. Otherwise when ever we call this we will have to call it for both user and machine.

src/AWS.Deploy.Orchestration/Utilities/Helpers.cs
```diff
@@ -0,0 +1,90 @@
...
+        /// It will first look for AWS_DOTNET_DEPLOYTOOL_WORKSPACE environment variable set by the user. It will be used as the deploy tool workspace if it points to a valid directory whithout whitespace characters in its path.
+        /// If the environment variable is set, it will also create a temp directory inside the workspace and set process scoped TEMP and TMP environment variables that point to the temp directory.
+        /// This additional configuration is required due to a known issue in the CDK - https://github.com/aws/aws-cdk/issues/2532
+        /// If the override is not present, then it defaults to USERPROFILE/.aws-dotnet-deploy.
+        /// It will throw an exception if the USERPROFILE contains a whitespace character.
+        /// </summary>
+        public static string GetDeployToolWorkspaceDirectoryRoot(string userProfilePath, IDirectoryManager directoryManager, IEnvironmentVariableManager environmentVariableManager)
+        {
+            var overridenWorkspace = environmentVariableManager.GetEnvironmentVariable("AWS_DOTNET_DEPLOYTOOL_WORKSPACE", EnvironmentVariableTarget.User)
```
normj, line 53: Use at least a local constant for `AWS_DOTNET_DEPLOYTOOL_WORKSPACE`.

src/AWS.Deploy.Orchestration/Utilities/Helpers.cs
```diff
@@ -0,0 +1,90 @@
...
+               ?? environmentVariableManager.GetEnvironmentVariable("AWS_DOTNET_DEPLOYTOOL_WORKSPACE", EnvironmentVariableTarget.Machine);
+
+            var defaultWorkSpace = Path.Combine(userProfilePath, ".aws-dotnet-deploy");
+
+            if (overridenWorkspace == null)
+            {
+                return !defaultWorkSpace.Contains(" ")
+                    ? defaultWorkSpace
+                    : throw new InvalidDeployToolWorkspaceException(DeployToolErrorCode.InvalidDeployToolWorkspace, $"The USERPROFILE path ({userProfilePath}) contains a whitespace character and cannot be used as a workspace by the deployment tool. " +
+                    "Please refer to the troubleshooting guide for additional configuration required to handle this issue.");
```
normj, line 63: Can we be more prescriptive here and tell them refer to the troubleshooting guide for setting the `AWS_DOTNET_DEPLOYTOOL_WORKSPACE` environment variable to use an alternative workspace error.

src/AWS.Deploy.Orchestration/Utilities/Helpers.cs
```diff
@@ -0,0 +1,90 @@
...
+                    $"The AWS_DOTNET_DEPLOYTOOL_WORKSPACE environment variable has been set to \"{overridenWorkspace}\" but it does not point to a valid directory.");
+            }
+
+            if (overridenWorkspace.Contains(" "))
+            {
+                throw new InvalidDeployToolWorkspaceException(DeployToolErrorCode.InvalidDeployToolWorkspace,
+                    $"The AWS_DOTNET_DEPLOYTOOL_WORKSPACE environment variable ({overridenWorkspace}) contains a whitespace character and cannot be used as a workspace by the deployment tool.");
+            }
+
+            var tempDir = Path.Combine(overridenWorkspace, "temp");
```
normj, line 78: Add comment about why we are setting the temp environment variable to work around CDK issue.

normj (changes requested):
For testing since we have environment variables abstracted with `EnvironmentManager` I suggest adding a testing reset method on `DeployToolWorkspaceMetadata` that can inject a custom instance of `EnvironmentManager` with the workspace environment variable set.

src/AWS.Deploy.Orchestration/Utilities/EnvironmentVariableManager.cs
96malhar: Fixed

src/AWS.Deploy.Orchestration/Utilities/Helpers.cs
96malhar: Added a public constant.

src/AWS.Deploy.Orchestration/Utilities/Helpers.cs
96malhar: Done

src/AWS.Deploy.Orchestration/Utilities/Helpers.cs
96malhar: Done

96malhar:
> For testing since we have environment variables abstracted with `EnvironmentManager` I suggest adding a testing reset method on `DeployToolWorkspaceMetadata` that can inject a custom instance of `EnvironmentManager` with the workspace environment variable set.

Reworked the PR to use dependency injection and added an integ test.

normj (approved):
Just a minor comment about the doc update but otherwise I'm good with the change.

philasmar (approved):

PR #621, opened 2022-06-16
build: use PackageOutputAbsolutePath instead to match dotnet pack --output dir

ganeshnj:
## Motivation

The pipeline uses custom location for packing the NuGet package and therefore the local nuget cache doesn't have the expected package.

## Change
PackageOutputAbsolutePath MSBuild property is set with the same value provided in `dotnet pack --output` command, therefore, this change aligns them both.

## Result
On executing `dotnet pack -o "C:\Users\jangirg\temp"`, the package is created at the output path.

*Issue #, if available:*

*Description of changes:*


attilah (approved):

philasmar (approved):

ganeshnj:
Let's not merge this PR until we figure what is wrong with prod pipeline.

normj (approved):

PR #627, opened 2022-06-20
feat: add troubleshooting guide and cut a ticket links on exception

philasmar:
*Issue #, if available:*
DOTNET-6008

*Description of changes:*
add troubleshooting guide and cut a ticket links on exception
Screen Shot 2022-06-20 at 12 50 10 PM

### DIFF
--- a/src/AWS.Deploy.CLI/Commands/CommandFactory.cs
+++ b/src/AWS.Deploy.CLI/Commands/CommandFactory.cs
@@ -263,6 +263,10 @@ namespace AWS.Deploy.CLI.Commands
                         _toolInteractiveService.WriteErrorLine(e.Message);
                     }
 
+                    _toolInteractiveService.WriteErrorLine(string.Empty);
+                    _toolInteractiveService.WriteErrorLine("For more information, please visit our troubleshooting guide https://aws.github.io/aws-dotnet-deploy/troubleshooting-guide/.");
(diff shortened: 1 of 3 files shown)


src/AWS.Deploy.CLI/Commands/CommandFactory.cs
```diff
@@ -258,6 +258,10 @@ private Command BuildDeployCommand()
                         _toolInteractiveService.WriteErrorLine(e.Message);
                     }
 
+                    _toolInteractiveService.WriteErrorLine(string.Empty);
+                    _toolInteractiveService.WriteErrorLine("For more information, please visit our troubleshooting guide https://aws.github.io/aws-dotnet-deploy/troubleshooting-guide/missing-dependencies/.");
```
normj, line 262: Seems weird to send all error messages to `missing-dependencies`. I recognize that is the nature of our doc site given you can't just click on the `Trouble Shooting` node. I wonder if we should rename the `other-issues` page to `general-issues`, make that page the first in the `Trouble Shooting` pages and then link to that page.

normj (changes requested):

src/AWS.Deploy.CLI/Commands/CommandFactory.cs
philasmar: Done

96malhar (approved):

normj (approved):

PR #635, opened 2022-06-22
chore: update the CDK Bootstrap template to the latest version

philasmar:
*Description of changes:*
update the CDK Bootstrap template to the latest version


ashovlin (approved):

src/AWS.Deploy.Orchestration/CDK/CDKBootstrapTemplate.yaml
```diff
@@ -489,7 +490,7 @@ Resources:
       Type: String
       Name:
         Fn::Sub: /cdk-bootstrap/${Qualifier}/version
-      Value: "12"
+      Value: "13"
```
normj, line 493: You need to also update the [Constant]( https://github.com/aws/aws-dotnet-deploy/blob/5b78ecaec217908c19d92fabbdfcf1963821194a/src/AWS.Deploy.Constants/CDK.cs#L19)

If you are filling ambitious you get rid of the constant and parse the version from the bootstrap yaml file so we don't have to remember to update both places in the future.

normj (changes requested):

src/AWS.Deploy.Orchestration/CDK/CDKBootstrapTemplate.yaml
philasmar: Done

96malhar (changes requested):
Can we add a unit test for ` ReadCDKTemplateVersion` ?

96malhar:
> Can we add a unit test for ` ReadCDKTemplateVersion` ?

Synced up with @philasmar offline and verified that the new method is being covered by an existing test - https://github.com/aws/aws-dotnet-deploy/blob/main/test/AWS.Deploy.Orchestration.UnitTests/CDK/CDKProjectHandlerTests.cs#L126

96malhar (approved):

PR #636, opened 2022-06-22
fix: Throw an exception when attempting to generate a Dockerfile with project references located outside of the solution directory

ashovlin:
*Issue #, if available:* DOTNET-6011

*Description of changes:*
When generating a Dockerfile if one was not provided, we now throw an exception if we detect project references outside of the directory (or its children) of the solution:
image

This is to avoid a wide Docker execution context, which is potentially including folders/projects unrelated to the file.

This is similar to the error Visual Studio shows in the same situation:
image

I'll add a troubleshooting guide entry in a separate PR into this feature branch.

**Note to reviewers:**  A majority of the diff is a new sample project structured to match the failing case. Focus on the first two files, but then also the changes to the test `Dockerfile` and `ReferenceDockerfile`.


test/AWS.Deploy.CLI.UnitTests/DockerTests.cs
```diff
@@ -69,7 +96,7 @@ private string ResolvePath(string projectName)
             return Path.Combine(testsPath, "..", "testapps", "docker", projectName);
         }
 
-        private void AssertDockerFilesAreEqual(string path, string generatedFile = "Dockerfile", string referenceFile = "Dockerfile")
+        private void AssertDockerFilesAreEqual(string path, string generatedFile = "Dockerfile", string referenceFile = "ReferenceDockerfile")
```
ashovlin, line 99: I noticed this test was never doing what it seemed designed to do - compare a newly generated Dockerfile against a static reference file. We have been comparing the generated file against itself.

Updated the test, which also required updating a lot of the reference files to match the current template.

philasmar (approved):

PR #642, opened 2022-06-23
chore: Update all recipe versions to 1.0.0

normj:
*Description of changes:*
Since we are getting closer to GA this PR updates all of the recipe version numbers to 1.0.0.


ashovlin (approved):
Approving, but curious, will we now have to evolve the version number every time the recipe is modified? Following semver conventions?

normj:
> Approving, but curious, will we now have to evolve the version number every time the recipe is modified? Following semver conventions?

In the future we will have to build in features with major (breaking) version bumps in recipes and figure out how to handle that. Till we do that all changes to recipes will need to be backward compatible and the version must stay in the 1.X range.

philasmar (approved):

PR #645, opened 2022-06-24
feat: Add ability to save deployment settings to a config file

96malhar:
*Issue #, if available:*
DOTNET-6021

*Description of changes:*
This PR adds the ability to save deployment settings to a config file using the following command:
`dotnet aws deploy [--save-settings | --save-all-settings] <deployment_settings_path> --project-path <project_path>`

Documentation PR - https://github.com/aws/aws-dotnet-deploy/pull/668 (needs to be updated to include the `--save-settings-all` switch)

### DIFF
--- a/src/AWS.Deploy.CLI/Commands/CommandHandlerInput/DeployCommandHandlerInput.cs
+++ b/src/AWS.Deploy.CLI/Commands/CommandHandlerInput/DeployCommandHandlerInput.cs
@@ -10,13 +10,54 @@ namespace AWS.Deploy.CLI.Commands.CommandHandlerInput
 {
     public class DeployCommandHandlerInput
     {
+        /// <summary>
+        /// AWS credential profile used to make calls to AWS.
+        /// </summary>
         public string? Profile { get; set; }
+
+        /// <summary>
+        /// AWS region to deploy the application to. For example, us-west-2.
+        /// </summary>
         public string? Region { get; set; }
+
+        /// <summary>
+        /// Path to the project to deploy.
+        /// </summary>
         public string? ProjectPath { get; set; }
+
+        /// <summary>
+        /// Name of the cloud application.
+        /// </summary>
         public string? ApplicationName { get; set; }
+
+        /// <summary>
+        /// Path to the deployment settings file to be applied.
+        /// </summary>
         public string? Apply { get; set; }
+
+        /// <summary>
+        /// Flag to enable diagnostic output.
+        /// </summary>
         public bool Diagnostics { get; set; }
+
+        /// <summary>
+        /// Flag to disable interactivity to execute commands without any prompts.
+        /// </summary>
         public bool Silent { get; set; }
+
+        /// <summary>
+        /// The absolute or relative path of the CDK project that will be used for deployment.
+        /// </summary>
         public string? DeploymentProject { get; set; }
+
+        /// <summary>
+        /// The absolute or the relative JSON file path where the deployment settings will be saved. Only the settings modified by the user are persisted.
+        /// </summary>
+        public string? SaveSettings { get; set; }
+
+        /// <summary>
+        /// The absolute or the relative JSON file path where the deployment settings will be saved. All deployment settings are persisted.
+        /// </summary>
+        public string? SaveAllSettings { get; set; }
     }
 }
--- a/src/AWS.Deploy.CLI/Commands/DeployCommand.cs
+++ b/src/AWS.Deploy.CLI/Commands/DeployCommand.cs
@@ -113,7 +113,7 @@ namespace AWS.Deploy.CLI.Commands
             _deploymentSettingsHandler = deploymentSettingsHandler;
         }
 
-        public async Task ExecuteAsync(string applicationName, string deploymentProjectPath, DeploymentSettings? deploymentSettings = null)
+        public async Task ExecuteAsync(string applicationName, string deploymentProjectPath, SaveSettingsConfiguration saveSettingsConfig, DeploymentSettings? deploymentSettings = null)
         {
             var (orchestrator, selectedRecommendation, cloudApplication) = await InitializeDeployment(applicationName, deploymentSettings, deploymentProjectPath);
 
@@ -129,6 +129,12 @@ namespace AWS.Deploy.CLI.Commands
             }
 
             await CreateDeploymentBundle(orchestrator, selectedRecommendation, cloudApplication);
+
+            if (saveSettingsConfig.SettingsType != SaveSettingsType.None)
+            {
+                await _deploymentSettingsHandler.SaveSettings(saveSettingsConfig, selectedRecommendation, cloudApplication, _session);
+                _toolInteractiveService.WriteLine($"{Environment.NewLine}Successfully saved the deployment settings at {saveSettingsConfig.FilePath}");
+            }
 
             await orchestrator.DeployRecommendation(cloudApplication, selectedRecommendation);
 
@@ -193,11 +199,27 @@ namespace AWS.Deploy.CLI.Commands
                 cloudApplicationName = deploymentSettings?.ApplicationName ?? string.Empty;
 
             // Prompt the user with a choice to re-deploy to existing targets or deploy to a new cloud application.
-            if (string.IsNullOrEmpty(cloudApplicationName))
+            // This prompt is NOT needed if the user is just pushing the docker image to ECR.
+            if (string.IsNullOrEmpty(cloudApplicationName) && !string.Equals(deploymentSettings?.RecipeId, Constants.RecipeIdentifier.PUSH_TO_ECR_RECIPE_ID))
                 cloudApplicationName = AskForCloudApplicationNameFromDeployedApplications(compatibleApplications);
 
             // Find existing application with the same CloudApplication name.
-            var deployedApplication = allDeployedApplications.FirstOrDefault(x => string.Equals(x.Name, cloudApplicationName));
+            CloudApplication? deployedApplication = null;
+            if (!string.IsNullOrEmpty(deploymentSettings?.RecipeId))
+            {
+                // if the recommendation is specified via a config file then find the deployed application by matching the deployment type along with the cloudApplicationName
+                var recommendation = recommendations.FirstOrDefault(x => string.Equals(x.Recipe.Id, deploymentSettings.RecipeId));
+                if (recommendation == null)
+                {
+                    var errorMsg = "The recipe ID specified in the deployment settings file does not match any compatible deployment recipes.";
+                    throw new InvalidDeploymentSettingsException(DeployToolErrorCode.DeploymentConfigurationNeedsAdjusting, errorMsg);
+                }
+                deployedApplication = allDeployedApplications.FirstOrDefault(x => string.Equals(x.Name, cloudApplicationName) && x.DeploymentType == recommendation.Recipe.DeploymentType);
+            }
+            else
+            {
+                deployedApplication = allDeployedApplications.FirstOrDefault(x => string.Equals(x.Name, cloudApplicationName));
+            }
 
             Recommendation? selectedRecommendation = null;
             if (deployedApplication != null)
--- a/src/AWS.Deploy.Orchestration/DeploymentBundleHandler.cs
+++ b/src/AWS.Deploy.Orchestration/DeploymentBundleHandler.cs
@@ -171,6 +171,8 @@ namespace AWS.Deploy.Orchestration
                 }
             }
 
+            // The docker build command will fail if a relative path is provided
+            dockerExecutionDirectory = _directoryManager.GetAbsolutePath(projectDirectory, dockerExecutionDirectory);
             return dockerExecutionDirectory;
         }
 
--- a/src/AWS.Deploy.Orchestration/DeploymentSettingsHandler.cs
+++ b/src/AWS.Deploy.Orchestration/DeploymentSettingsHandler.cs
@@ -3,6 +3,7 @@
 
 using System;
 using System.Collections.Generic;
+using System.IO;
 using System.Linq;
 using System.Text;
 using System.Threading.Tasks;
@@ -25,17 +26,25 @@ namespace AWS.Deploy.Orchestration
         /// Iterates over the option setting values found at <see cref="DeploymentSettings"/> and applies them to the selected recommendation
         /// </summary>
         Task ApplySettings(DeploymentSettings deploymentSettings, Recommendation recommendation, IDeployToolValidationContext deployToolValidationContext);
+
+        /// <summary>
+        /// Save the deployment settings at the specified file path.
+        /// </summary>
+        /// <exception cref="FailedToSaveDeploymentSettingsException">Thrown if this operation fails.</exception>
+        Task SaveSettings(SaveSettingsConfiguration saveSettingsConfig, Recommendation recommendation, CloudApplication cloudApplication, OrchestratorSession orchestratorSession);
     }
 
     public class DeploymentSettingsHandler : IDeploymentSettingsHandler
     {
-        private readonly IFileManager _filemanager;
+        private readonly IFileManager _fileManager;
+        private readonly IDirectoryManager _directoryManager;
         private readonly IOptionSettingHandler _optionSettingHandler;
         private readonly IRecipeHandler _recipeHandler;
 
-        public DeploymentSettingsHandler(IFileManager fileManager, IOptionSettingHandler optionSettingHandler, IRecipeHandler recipeHandler)
+        public DeploymentSettingsHandler(IFileManager fileManager, IDirectoryManager directoryManager, IOptionSettingHandler optionSettingHandler, IRecipeHandler recipeHandler)
         {
-            _filemanager = fileManager;
+            _fileManager = fileManager;
+            _directoryManager = directoryManager;
             _optionSettingHandler = optionSettingHandler;
             _recipeHandler = recipeHandler;
         }
@@ -44,7 +53,7 @@ namespace AWS.Deploy.Orchestration
         {
             try
             {
-                var contents = await _filemanager.ReadAllTextAsync(filePath);
+                var contents = await _fileManager.ReadAllTextAsync(filePath);
                 var userDeploymentSettings = JsonConvert.DeserializeObject<DeploymentSettings>(contents);
                 return userDeploymentSettings;
             }
@@ -90,6 +99,82 @@ namespace AWS.Deploy.Orchestration
             }
 
             throw new InvalidDeploymentSettingsException(DeployToolErrorCode.DeploymentConfigurationNeedsAdjusting, errorMessage.Trim());
+        }
+
+        public async Task SaveSettings(SaveSettingsConfiguration saveSettingsConfig, Recommendation recommendation, CloudApplication cloudApplication, OrchestratorSession orchestratorSession)
+        {
+            if (saveSettingsConfig.SettingsType == SaveSettingsType.None)
+            {
+                // We are not throwing an expected exception here as this issue is not caused by the user.
+                throw new InvalidOperationException($"Cannot persist settings with {SaveSettingsType.None}");
+            }
+
+            if (!_fileManager.IsFileValidPath(saveSettingsConfig.FilePath))
+            {
+                var message = $"Failed to save deployment settings because {saveSettingsConfig.FilePath} is invalid or its parent directory does not exist on disk.";
+                throw new FailedToSaveDeploymentSettingsException(DeployToolErrorCode.FailedToSaveDeploymentSettings, message);
+            }
+
+            var projectDirectory = Path.GetDirectoryName(orchestratorSession.ProjectDefinition.ProjectPath);
+            if (string.IsNullOrEmpty(projectDirectory))
+            {
+                var message = "Failed to save deployment settings because the current deployment session does not have a valid project path";
+                throw new FailedToSaveDeploymentSettingsException(DeployToolErrorCode.FailedToSaveDeploymentSettings, message);
+            }
+
+            var deploymentSettings = new DeploymentSettings
+            {
+                AWSProfile = orchestratorSession.AWSProfileName,
+                AWSRegion = orchestratorSession.AWSRegion,
+                ApplicationName = recommendation.Recipe.DeploymentType == DeploymentTypes.ElasticContainerRegistryImage ? null : cloudApplication.Name,
+                RecipeId = cloudApplication.RecipeId,
+                Settings = new Dictionary<string, object>()
+            };
+
+            var optionSettings = recommendation.GetConfigurableOptionSettingItems();
+            foreach (var optionSetting in optionSettings)
+            {
+                if (saveSettingsConfig.SettingsType == SaveSettingsType.Modified && !_optionSettingHandler.IsOptionSettingModified(recommendation, optionSetting))
+                {
+                    continue;
+                }
+
+                var id = optionSetting.FullyQualifiedId;
+                var value = _optionSettingHandler.GetOptionSettingValue(recommendation, optionSetting);
+                if (optionSetting.TypeHint.HasValue && (optionSetting.TypeHint == OptionSettingTypeHint.FilePath || optionSetting.TypeHint == OptionSettingTypeHint.DockerExecutionDirectory))
+                {
+                    var path = value?.ToString();
+                    if (string.IsNullOrEmpty(path))
+                    {
+                        continue;
+                    }
+
+                    // All file paths or directory paths must be persisted relative the the customers .NET project.
+                    // This is a done to ensure that the resolved paths work correctly across all cloned repos.
+                    // The relative path is also canonicalized to work across Unix and Windows OS.
+                    var absolutePath = _directoryManager.GetAbsolutePath(projectDirectory, path);
+                    value = _directoryManager.GetRelativePath(projectDirectory, absolutePath)
+                                .Replace(Path.DirectorySeparatorChar, '/');
+                }
+                deploymentSettings.Settings[id] = value;
+            }
+
+            try
+            {
+                var content = JsonConvert.SerializeObject(deploymentSettings, new JsonSerializerSettings
+                {
+                    Formatting = Formatting.Indented,
+                    NullValueHandling = NullValueHandling.Ignore,
+                    ContractResolver = new SerializeModelContractResolver()
+                });
+
+                await _fileManager.WriteAllTextAsync(saveSettingsConfig.FilePath, content);
+            }
+            catch (Exception ex)
+            {
+                var message = $"Failed to save the deployment settings at {saveSettingsConfig.FilePath} due to the following error: {Environment.NewLine}{ex.Message}";
+                throw new FailedToSaveDeploymentSettingsException(DeployToolErrorCode.FailedToSaveDeploymentSettings, message, ex);
+            }
         }
     }
 }
--- a/src/AWS.Deploy.Orchestration/OptionSettingHandler.cs
+++ b/src/AWS.Deploy.Orchestration/OptionSettingHandler.cs
@@ -92,6 +92,22 @@ namespace AWS.Deploy.Orchestration
         }
 
         /// <summary>
+        /// Assigns a value to the OptionSettingItem based on the fullyQualifiedId
+        /// </summary>
+        /// <exception cref="ValidationFailedException">
+        /// Thrown if one or more <see cref="Validators"/> determine
+        /// <paramref name="value"/> is not valid.
+        /// </exception>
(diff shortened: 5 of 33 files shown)


src/AWS.Deploy.CLI/Commands/DeployCommand.cs
```diff
@@ -130,6 +130,19 @@ public async Task ExecuteAsync(string applicationName, string deploymentProjectP
 
             await CreateDeploymentBundle(orchestrator, selectedRecommendation, cloudApplication);
 
+            if (!string.IsNullOrEmpty(saveSettingsFilePath))
+            {
+                var errorMessage = await _deploymentSettingsHandler.SaveSettings(saveSettingsFilePath, selectedRecommendation, cloudApplication, _session);
```
philasmar, line 135: error messages should not be returned this way. We should use exceptions to present invalid state.

src/AWS.Deploy.Orchestration/DeploymentSettingsHandler.cs
```diff
@@ -91,5 +100,53 @@ public async Task ApplySettings(DeploymentSettings deploymentSettings, Recommend
...
+                    Formatting = Formatting.Indented,
+                    NullValueHandling = NullValueHandling.Ignore,
+                    ContractResolver = new SerializeModelContractResolver()
+                });
+
+                await _fileManager.WriteAllTextAsync(filePath, content);
+            }
+            catch (Exception ex)
+            {
+                return $"Failed to save the deployment settings at {filePath} due to the following error: {Environment.NewLine}{ex.Message}";
```
philasmar, line 146: An expected exception should be thrown here. Also make sure to pass the generic exception ex to the inner exception of the expected one.

philasmar (changes requested):

src/AWS.Deploy.Orchestration/DeploymentSettingsHandler.cs
```diff
@@ -91,5 +100,53 @@ public async Task ApplySettings(DeploymentSettings deploymentSettings, Recommend
...
+            if (string.IsNullOrEmpty(parentDirectory))
+            {
+                return $"Failed to save deployment settings because {filePath} is not a valid path";
+            }
+            if (!_directoryManager.Exists(parentDirectory))
+            {
+                return $"Failed to save deployment settings because part of {filePath} does not exist on disk";
+            }
+
+            var deploymentSettings = new DeploymentSettings
```
ashovlin, line 116: Doesn't the caller of `SaveSettings` (`DeployCommand.ExecuteAsync`) already have a `DeploymentSettings` object? Should we reuse that instead of recreating here?

I'm not sure if there's a good reason not to, just thinking about if we ever change how it's initialized there are now multiple places to update.

ashovlin (commented):
How will server mode use this?

src/AWS.Deploy.CLI/Commands/CommandHandlerInput/DeployCommandHandlerInput.cs
```diff
@@ -18,5 +18,6 @@ public class DeployCommandHandlerInput
         public bool Diagnostics { get; set; }
         public bool Silent { get; set; }
         public string? DeploymentProject { get; set; }
+        public string? SaveSettings { get; set; }
```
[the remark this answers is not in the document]
96malhar: Added comments on all properties in `DeployCommandHandlerInput`

src/AWS.Deploy.CLI/Commands/DeployCommand.cs
96malhar: Changed ` _deploymentSettingsHandler.SaveSettings(...)` to throw an expected exception

src/AWS.Deploy.CLI/Commands/DeployCommand.cs
```diff
@@ -130,6 +130,19 @@ public async Task ExecuteAsync(string applicationName, string deploymentProjectP
 
             await CreateDeploymentBundle(orchestrator, selectedRecommendation, cloudApplication);
 
+            if (!string.IsNullOrEmpty(saveSettingsFilePath))
+            {
+                var errorMessage = await _deploymentSettingsHandler.SaveSettings(saveSettingsFilePath, selectedRecommendation, cloudApplication, _session);
+                if (!string.IsNullOrEmpty(errorMessage))
+                {
+                    _toolInteractiveService.WriteErrorLine(errorMessage);
```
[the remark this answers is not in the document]
96malhar: I am not sure if we want to fail the entire deployment in we fail to save the settings file. For now I am wrapping `_deploymentSettingsHandler.SaveSettings(..)` in a `try-catch` and logging the error message in the catch block. Interested in @normj's and @ashovlin's opinion on it.

src/AWS.Deploy.Orchestration/DeploymentSettingsHandler.cs
```diff
@@ -91,5 +100,53 @@ public async Task ApplySettings(DeploymentSettings deploymentSettings, Recommend
...
+        public async Task<string> SaveSettings(string filePath, Recommendation recommendation, CloudApplication cloudApplication, OrchestratorSession orchestratorSession)
+        {
+            var parentDirectory = Path.GetDirectoryName(filePath);
+            if (string.IsNullOrEmpty(parentDirectory))
+            {
+                return $"Failed to save deployment settings because {filePath} is not a valid path";
+            }
+            if (!_directoryManager.Exists(parentDirectory))
+            {
+                return $"Failed to save deployment settings because part of {filePath} does not exist on disk";
```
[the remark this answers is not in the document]
96malhar: Fixed

src/AWS.Deploy.Orchestration/DeploymentSettingsHandler.cs
96malhar: The `DeploymentSettings` from `DeployCommand.ExecuteAsync` is only intialized if the `--apply` flag is provided.
https://github.com/aws/aws-dotnet-deploy/blob/dev/src/AWS.Deploy.CLI/Commands/CommandFactory.cs#L196-#L200

That object is only used to apply the deployment settings that were specified in a config file.

src/AWS.Deploy.Orchestration/DeploymentSettingsHandler.cs
```diff
@@ -91,5 +100,53 @@ public async Task ApplySettings(DeploymentSettings deploymentSettings, Recommend
...
+            var deploymentSettings = new DeploymentSettings
+            {
+                AWSProfile = orchestratorSession.AWSProfileName,
+                AWSRegion = orchestratorSession.AWSRegion,
+                ApplicationName = cloudApplication.Name,
+                RecipeId = cloudApplication.RecipeId,
+                Settings = new Dictionary<string, object>()
+            };
+
+            var optionSettings = recommendation.GetConfigurableOptionSettingItems();
```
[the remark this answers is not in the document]
96malhar: The deployment settings bundle can be specified by the user in the JSON config file under the `Settings` property

src/AWS.Deploy.Orchestration/DeploymentSettingsHandler.cs
96malhar: Done

96malhar:
> More generally, since there wasn't a good place to attach a comment - how will server mode use this feature?

Per my understanding, this is a CLI only feature as this feature is triggered by the `--save-settings` option inside the main `dotnet aws deploy` command. Currently server mode does not account for any options supplied with the top-level commands (unless I am missing something).

ashovlin:
> > More generally, since there wasn't a good place to attach a comment - how will server mode use this feature?
> 
> Per my understanding, this is a CLI only feature as this feature is triggered by the `--save-settings` option inside the main `dotnet aws deploy` command. Currently server mode does not account for any options supplied with the top-level commands (unless I am missing something).

Based on the discussion at the internal stand-up today, fine punting on this for now. `--save-settings` and `--apply` are primarily meant for automated/CI/etc. scenarios, which are orthogonal to the IDE experience.

src/AWS.Deploy.CLI/Commands/DeployCommand.cs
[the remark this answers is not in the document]
ashovlin: I was also wondering about the ordering of `SaveSettings` vs. the deployment itself. Do we want to be saving the settings before we know whether or not they'll be successful?

On Phil's original question, I lean towards failing the deployment to avoid anything unexpected:
1. User deploys settings A
2. User deploys settings B, but save fails so file is still at A
3. User deploys again, thinking they are applying B, but really applying A

src/AWS.Deploy.CLI/Commands/DeployCommand.cs
```diff
@@ -130,6 +130,23 @@ public async Task ExecuteAsync(string applicationName, string deploymentProjectP
...
+                    await _deploymentSettingsHandler.SaveSettings(saveSettingsFilePath, selectedRecommendation, cloudApplication, _session);
+                }
+                catch (FailedToSaveDeploymentSettingsException ex)
+                {
+                    success = false;
+                    _toolInteractiveService.WriteErrorLine(ex.Message);
+                }
+
+                if (success)
+                    _toolInteractiveService.WriteLine($"{Environment.NewLine} Successfully saved the deployment settings at {saveSettingsFilePath}");
```
normj, line 147: You can add this `WriteLine` just after the `SaveSettings` call on line 138. If there was an exception then `WriteLine` wouldn't be called. Then you can simplify the code and get rid of the local `success` variable.

src/AWS.Deploy.Orchestration/DeploymentSettingsHandler.cs
```diff
@@ -25,17 +26,25 @@ public interface IDeploymentSettingsHandler
         /// Iterates over the option setting values found at <see cref="DeploymentSettings"/> and applies them to the selected recommendation
         /// </summary>
         Task ApplySettings(DeploymentSettings deploymentSettings, Recommendation recommendation, IDeployToolValidationContext deployToolValidationContext);
+
+        /// <summary>
+        /// Save the deployment settings at the specified file path
+        /// </summary>
+        /// <returns>An error message if the operation is not successful. The error message will be empty if it is successful</returns>
```
normj, line 33: It doesn't return an error message, an exception is thrown if is not successful.

src/AWS.Deploy.Orchestration/DeploymentSettingsHandler.cs
```diff
@@ -91,5 +100,54 @@ public async Task ApplySettings(DeploymentSettings deploymentSettings, Recommend
...
+
+        public async Task SaveSettings(string filePath, Recommendation recommendation, CloudApplication cloudApplication, OrchestratorSession orchestratorSession)
+        {
+            var parentDirectory = Path.GetDirectoryName(filePath);
+            if (string.IsNullOrEmpty(parentDirectory))
+            {
+                var message = $"Failed to save deployment settings because {filePath} is not a valid path";
+                throw new FailedToSaveDeploymentSettingsException(DeployToolErrorCode.FailedToSaveDeploymentSettings, message);
+            }
+            if (!_directoryManager.Exists(parentDirectory))
```
normj, line 112: These 2 validations should have happened before the deployment even started. Not saying we can't also validate here is as well.

src/AWS.Deploy.Orchestration/DeploymentSettingsHandler.cs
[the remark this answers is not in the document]
normj: That means we need to make sure all settings with a File type hint like the dockerfile need to be written to the file as a relative path from the project path. The user might have specified in the interactive session an absolute path but we don't want users checking in files that are only valid for where they clone the repo.

src/AWS.Deploy.Orchestration/DeploymentSettingsHandler.cs
```diff
@@ -91,5 +100,54 @@ public async Task ApplySettings(DeploymentSettings deploymentSettings, Recommend
...
+                RecipeId = cloudApplication.RecipeId,
+                Settings = new Dictionary<string, object>()
+            };
+
+            var optionSettings = recommendation.GetConfigurableOptionSettingItems();
+            foreach (var optionSetting in optionSettings)
+            {
+                var id = optionSetting.FullyQualifiedId;
+                var value = _optionSettingHandler.GetOptionSettingValue(recommendation, optionSetting);
+                deploymentSettings.Settings[id] = value;
```
normj, line 132: We should talk with Irene if we should be writing all of the values or just the ones that are different then the default. Seems like we are creating a file that is overly complex looking but it does give you a full list of all of the possible values.

test/AWS.Deploy.CLI.IntegrationTests/WebAppNoDockerFileTests.cs
```diff
@@ -107,6 +108,9 @@ public async Task DefaultConfigurations()
             Assert.True(Directory.Exists(Path.Combine(_customWorkspace, "temp")));
             Assert.True(Directory.Exists(Path.Combine(_customWorkspace, "Projects")));
 
+            // Verify existence of the saved deployment settings file
+            Assert.True(File.Exists(saveSettingsFilePath));
```
normj, line 112: We aren't testing if the file can actually be used for deployment.

normj (changes requested):

test/AWS.Deploy.CLI.IntegrationTests/WebAppNoDockerFileTests.cs
96malhar: Modified an existing integ test to use the config file created via `_deploymentSettingsHandler.SaveSettings(...)` method

normj (approved):

src/AWS.Deploy.CLI/Commands/DeployCommand.cs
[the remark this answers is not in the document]
96malhar: Fixed this to fail the deployment if the operation to save the settings file fails.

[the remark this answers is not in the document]
96malhar: To Alex's comment, Currently, the settings file is created just before the deployment starts. At this point all option settings have been validated and the deployment bundle has been created. If the deployment fails, then chances are low that it happened due to a user error. I think it would be valuable to save the settings file for diagnostic purposes even if the deployment fails.

src/AWS.Deploy.Orchestration/DeploymentSettingsHandler.cs
96malhar: Fixed this to fail the deployment if the operation to save the settings file fails.

philasmar (changes requested):
Please have a look at Alex's comments as they are pretty severe

src/AWS.Deploy.Orchestration/DeploymentSettingsHandler.cs
96malhar: I think we should persist all values given that deploying via a config file is a slightly advance use case and users would prefer to have complete transparency over all the deployment settings. Also, we do not publicly document the default values for any of our option settings which makes it difficult for users to obtain this info without peeking into the recipe source code.

philasmar (approved):

src/AWS.Deploy.Orchestration/Orchestrator.cs
```diff
@@ -301,6 +301,10 @@ private async Task CreateContainerDeploymentBundle(CloudApplication cloudApplica
 
             await _deploymentBundleHandler.BuildDockerImage(cloudApplication, recommendation, imageTag);
 
+            // These option settings need to be persisted back as they are not always provided by the user and we have custom logic to determine their values
+            await _optionSettingHandler.SetOptionSettingValue(recommendation, Constants.Docker.DockerExecutionDirectoryOptionId, recommendation.DeploymentBundle.DockerExecutionDirectory);
```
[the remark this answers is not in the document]
96malhar: We are persisting these value after the successful invocation of the `BuildDockerImage` command. [This](https://github.com/aws/aws-dotnet-deploy/blob/48ef3bc4eef59896c2b17db5b6895fa4c3a575b1/src/AWS.Deploy.Orchestration/DeploymentBundleHandler.cs#L66-#L67) command retrieves the `DockerfilePath` and the `DockerExecutionDirectory` and sets them as part of the `recommendation.DeploymentBundle`. But it does not persist it as part of the actual option setting items. Also whatever values are persisted were valid and were successfully used by the `BuildDockerImage` command.

Ideally we'd prefer to set these values at a central location and the process that need these values just call the `_optionSettingsHandler.GetValue(..)` method. However that refactor requires some effort and I didn't want to tackle it as part of this story

src/AWS.Deploy.Orchestration/DeploymentBundleHandler.cs
```diff
@@ -171,6 +171,8 @@ private string GetDockerExecutionDirectory(Recommendation recommendation)
                 }
             }
 
+            // The docker build command will fail if a relative path is provided
+            dockerExecutionDirectory = _directoryManager.GetAbsolutePath(projectDirectory, dockerExecutionDirectory);
```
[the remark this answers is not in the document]
ashovlin: This was my bad, I mixed up file and directory. Agree with absolute paths for the command.

justinmk3:
Is this different than the legacy `aws-lambda-tools-defaults.json` file? If so, why? https://github.com/aws/aws-toolkit-vscode/issues/2513

96malhar:
> Is this different than the legacy `aws-lambda-tools-defaults.json` file? If so, why? [aws/aws-toolkit-vscode#2513](https://github.com/aws/aws-toolkit-vscode/issues/2513)
> 
> Will this be a documented format? Should sam cli and/or AWS Toolkits try to re-use this format?

@justinmk3 Currently this tool does not support deployments of AWS Lambda functions. This feature is used to persist deployment settings in a config file so that it can be re-used in the customer's CI/CD system. 

This is an example config file
```
{
  "AWSProfile": "Prod",
  "AWSRegion": "us-west-2",
  "ApplicationName": "WebAppNoDockerFile",
  "RecipeId": "AspNetAppElasticBeanstalkLinux",
  "Settings": {
    "BeanstalkApplication": {
      "CreateNew": true,
      "ApplicationName": "WebAppNoDockerFile"
    },
    "BeanstalkEnvironment": {
      "EnvironmentName": "WebAppNoDockerFile-dev"
    }, 
    "EnvironmentType": "SingleInstance",
    "LoadBalancerType": "Classic",
    "XRayTracingSupportEnabled": false,
    "ReverseProxy": "nginx",
    "EnhancedHealthReporting": "enhanced",
    "SelfContainedBuild": false
  }
}
```

In future we do plan to support deployments of Lambda functions via this tool. I will defer the question of whether this format will replace `aws-lambda-tools-defaults.json` to @ik130

src/AWS.Deploy.Orchestration/OptionSettingHandler.cs
```diff
@@ -333,5 +349,85 @@ public bool IsSummaryDisplayable(Recommendation recommendation, OptionSettingIte
...
+                return false;
+            }
+
+            if (optionSetting.Type.Equals(OptionSettingValueType.List))
+            {
+                var currentSet = GetOptionSettingValue<SortedSet<string>>(recommendation, optionSetting);
+                var defaultSet = GetOptionSettingDefaultValue<SortedSet<string>>(recommendation, optionSetting);
+
+                if (defaultSet == null)
+                    return currentSet?.Count > 0;
```
philasmar, line 372: i believe we could get an exception here. what happens if currentSet is null? currentSet?.Count would be null. does `null > 0` throw an error?

src/AWS.Deploy.Orchestration/OptionSettingHandler.cs
```diff
@@ -333,5 +349,85 @@ public bool IsSummaryDisplayable(Recommendation recommendation, OptionSettingIte
...
+                var defaultValue = GetOptionSettingDefaultValue<bool>(recommendation, optionSetting);
+                return defaultValue != currentValue;
+            }
+
+            if (optionSetting.Type.Equals(OptionSettingValueType.String))
+            {
+                var currentValue = GetOptionSettingValue<string>(recommendation, optionSetting);
+                var defaultValue = GetOptionSettingDefaultValue<string>(recommendation, optionSetting);
+
+                if (string.IsNullOrEmpty(currentValue) && string.IsNullOrEmpty(defaultValue))
```
philasmar, line 417: I don't think we should treat null or empty then same. An option setting could be set to empty but doesn't have a default value. That could be an updated option setting.

src/AWS.Deploy.Orchestration/OptionSettingHandler.cs
```diff
@@ -333,5 +349,85 @@ public bool IsSummaryDisplayable(Recommendation recommendation, OptionSettingIte
...
+                    return false;
+
+                return !string.Equals(currentValue, defaultValue);
+            }
+
+            // The option setting is of type Object and it has nested child settings.
+            // return true is any of the child settings are modified.
+            foreach (var childSetting in optionSetting.ChildOptionSettings)
+            {
+                if (childSetting != null && IsOptionSettingModified(recommendation, childSetting))
```
philasmar, line 427: how could childSetting be null?

philasmar (changes requested):

src/AWS.Deploy.Orchestration/OptionSettingHandler.cs
96malhar: Can you point to an example where the default value for a string option setting is null, but it is still relevant to the deployment when it holds a `string.empty` value?


For example, consider the [StartCommand](https://github.com/aws/aws-dotnet-deploy/blob/main/src/AWS.Deploy.Recipes/RecipeDefinitions/ASP.NETAppAppRunner.recipe#L133) and [EncryptionKmsKey](https://github.com/aws/aws-dotnet-deploy/blob/main/src/AWS.Deploy.Recipes/RecipeDefinitions/ASP.NETAppAppRunner.recipe#L282) in the AppRunner recipe. They do not have an explicit default value (i.e the default values are `null`). However, the CDK template will only use these settings if they are not null or empty (effectively treating null and empty as same) [proof#1](https://github.com/aws/aws-dotnet-deploy/blob/main/src/AWS.Deploy.Recipes/CdkTemplates/AspNetAppAppRunner/Generated/Recipe.cs#L184), [proof#2](https://github.com/aws/aws-dotnet-deploy/blob/main/src/AWS.Deploy.Recipes/CdkTemplates/AspNetAppAppRunner/Generated/Recipe.cs#L203)

src/AWS.Deploy.Orchestration/OptionSettingHandler.cs
```diff
@@ -333,5 +349,79 @@ public bool IsSummaryDisplayable(Recommendation recommendation, OptionSettingIte
...
+            // If the option setting is not displayable, that means its dependencies are not satisfied and it does not play any role in the deployment.
+            // We do not need to evaluate whether it has been modified or not.
+            if (!IsOptionSettingDisplayable(recommendation, optionSetting))
+            {
+                return false;
+            }
+
+            if (optionSetting.Type.Equals(OptionSettingValueType.List))
+            {
+                var currentSet = GetOptionSettingValue<SortedSet<string>>(recommendation, optionSetting) ?? new SortedSet<string>();
```
ashovlin, line 368: Will `OptionSettingValueType.List` -> `SortedSet` potentially drop entries for list options that accept duplicate values?

Even if there aren't any examples today, wouldn't want to make this fragile since it'd be tough to remember to come back and update this.

ashovlin (commented):
1. Thanks for back-documenting `DeployCommandHandlerInput`!
2. I'm not strictly opposed, but the comparison-with-default feels a little fragile.
    * It's a fair amount of code
    * We have to remember to update it as we add types
    * Another thing to think about if we ever evolve default values because a user can't "pin" the current default). 
   
 Did you consider any alternatives (such as tracking a state for each `OptionSettingValue`)?

src/AWS.Deploy.Orchestration/OptionSettingHandler.cs
96malhar: Your concerns are valid. However, for all option settings of type `OptionSettingValueType.List`, we are already processing them as a `SortedSet`

* [Exhibit A](https://github.com/aws/aws-dotnet-deploy/blob/66fb1e34bbea10bbd995596e1eec801b79ff9102/src/AWS.Deploy.CLI/Commands/DeployCommand.cs#L737)
* [Exhibit B](https://github.com/aws/aws-dotnet-deploy/blob/66fb1e34bbea10bbd995596e1eec801b79ff9102/src/AWS.Deploy.Common/Recipes/OptionSettingItem.ValueOverride.cs#L139)
* [Exhibit C](https://github.com/aws/aws-dotnet-deploy/blob/66fb1e34bbea10bbd995596e1eec801b79ff9102/src/AWS.Deploy.Orchestration/OptionSettingHandler.cs#L299)

Should those occurences be changed as well? If so, I'd prefer to to do that (along with the new instance) in a seperate PR.

96malhar:
> 1. Thanks for back-documenting `DeployCommandHandlerInput`!
> 2. I'm not strictly opposed, but the comparison-with-default feels a little fragile.
>    
>    * It's a fair amount of code
>    * We have to remember to update it as we add types
>    * Another thing to think about if we ever evolve default values because a user can't "pin" the current default).
> 
> Did you consider any alternatives (such as tracking a state for each `OptionSettingValue`)?

I did consider a stateful approach by adding a new `IsModified` property to the `OptionSettingItem` class and then updating it whenever `_optionSettingHandler.SetValue(..)` is called. However, I would still need to compare the new value with the default value to see if the option setting is modified.

Do you have an approach that circumvents comparison to default values?

ashovlin (approved):
> I did consider a stateful approach by adding a new IsModified property to the OptionSettingItem class and then updating it whenever _optionSettingHandler.SetValue(..) is called. However, I would still need to compare the new value with the default value to see if the option setting is modified.

> Do you have an approach that circumvents comparison to default values?

That's what I was imagining, I'm fine if you explored it and pivoted to this.

philasmar (approved):

src/AWS.Deploy.Orchestration/OptionSettingHandler.cs
```diff
@@ -333,5 +349,79 @@ public bool IsSummaryDisplayable(Recommendation recommendation, OptionSettingIte
...
+                return defaultSet.Count != currentSet.Count || !currentSet.All(x => defaultSet.Contains(x));
+            }
+
+            if (optionSetting.Type.Equals(OptionSettingValueType.KeyValue))
+            {
+                var currentDict = GetOptionSettingValue<Dictionary<string, string>>(recommendation, optionSetting) ?? new Dictionary<string, string>();
+                var defaultDict = GetOptionSettingDefaultValue<Dictionary<string, string>>(recommendation, optionSetting) ?? new Dictionary<string, string>();
+
+                // return true if both have different lengths or all keyValue pairs are not equal between currentDict and defaultDict
+                return defaultDict.Count != currentDict.Count ||
```
normj, line 381: Aren't we missing the case that the keys are all the same but a user changed the value of a key?

src/AWS.Deploy.Orchestration/OptionSettingHandler.cs
```diff
@@ -333,5 +349,79 @@ public bool IsSummaryDisplayable(Recommendation recommendation, OptionSettingIte
 
             return true;
         }
+
+        /// <summary>
+        /// Checks whether the option setting item has been modified by the user. If it has been modified, then it will hold a non-default value
+        /// </summary>
+        /// <returns>true if the option setting item has been modified or false otherwise</returns>
+        public bool IsOptionSettingModified(Recommendation recommendation, OptionSettingItem optionSetting)
```
normj, line 357: I'm worried this yet another place that we have to remember to update whenever we add a new type to `OptionSettingValueType`. Could you add a comment to the `OptionSettingValueType` enum that list out these are the places that need to be changed if adding a new data type.

normj (changes requested):

src/AWS.Deploy.Orchestration/OptionSettingHandler.cs
96malhar:
```
currentDict.All(keyPair => defaultDict.ContainsKey(keyPair.Key) && string.Equals(defaultDict[keyPair.Key], currentDict[keyPair.Key]))
```
The `...&& string.Equals(defaultDict[keyPair.Key], currentDict[keyPair.Key]))` checks that the values are also equal for the same key

src/AWS.Deploy.Orchestration/OptionSettingHandler.cs
96malhar: Done

normj (approved):

PR #646, opened 2022-06-24
docs: Fix broken link in the troubleshooting guide

96malhar:
*Description of changes:*
This PR fixes broken link in the troubleshooting guide.


ashovlin (approved):
We might want to start using https://mkdocs.readthedocs.io/en/0.15.2/user-guide/configuration/#strict everywhere to guard against this.

Though looks like it won't flag the link if it doesn't end in `*.md`, so that may not have helped here.

https://github.com/mkdocs/mkdocs/issues/738#issuecomment-237338417

normj (approved):

PR #654, opened 2022-06-28
fix: Fix integration tests failure from setting a custom workspace

96malhar:
*Description of changes:*
The `WebAppNoDockerFileTests` integ tests sets up a custom workspace to use the deploy tool. However, the custom workspace cannot reference the private nuget cache created at - `<USERPROFILE>/.aws-dotnet-deploy/Projects/nuget-cache`.
As a result, the integ tests cannot find the latest (in-development/unreleased) version of `AWS.Deploy.Recipes.CDK.Common.nupkg`

This PR creates a `nuget.config` file inside the custom workspace and adds a reference to the above mentioned nuget-cache

### DIFF
--- a/test/AWS.Deploy.CLI.IntegrationTests/Helpers/TestEnvironmentVariableManager.cs
+++ b/test/AWS.Deploy.CLI.IntegrationTests/Helpers/TestEnvironmentVariableManager.cs
@@ -1,0 +1,28 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+using System;
+using System.Collections.Generic;
+using System.Text;
+using AWS.Deploy.Orchestration.Utilities;
+
+namespace AWS.Deploy.CLI.IntegrationTests.Helpers
+{
+    public class TestEnvironmentVariableManager : IEnvironmentVariableManager
+    {
+        public readonly Dictionary<string, string> store = new Dictionary<string, string>();
+
+        public string GetEnvironmentVariable(string variable)
+        {
+            return store.ContainsKey(variable) ? store[variable] : null;
+        }
+
+        public void SetEnvironmentVariable(string variable, string value)
+        {
+            if (string.Equals(variable, "AWS_DOTNET_DEPLOYTOOL_WORKSPACE"))
+                store[variable] = value;
+            else
+                Environment.SetEnvironmentVariable(variable, value);
+        }
+    }
+}
--- a/test/AWS.Deploy.CLI.IntegrationTests/Helpers/Utilities.cs
+++ b/test/AWS.Deploy.CLI.IntegrationTests/Helpers/Utilities.cs
@@ -1,0 +1,41 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+using System;
+using System.Collections.Generic;
+using System.IO;
+using System.Text;
+using AWS.Deploy.Orchestration.Utilities;
+using Microsoft.Extensions.DependencyInjection;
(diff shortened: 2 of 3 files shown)


test/AWS.Deploy.CLI.IntegrationTests/Helpers/TestEnvironmentVariableManager.cs
```diff
@@ -0,0 +1,28 @@
...
+        {
+            return store.ContainsKey(variable) ? store[variable] : null;
+        }
+
+        public void SetEnvironmentVariable(string variable, string value)
+        {
+            if (string.Equals(variable, "AWS_DOTNET_DEPLOYTOOL_WORKSPACE"))
+                store[variable] = value;
+            else
+                Environment.SetEnvironmentVariable(variable, value);
```
philasmar, line 25: Why is `AWS_DOTNET_DEPLOYTOOL_WORKSPACE` a special case? If this is a test implementation of IEnvironmentVariableManager, let's not impact the actual environment. So store all variables in the dictionary.

philasmar (changes requested):

test/AWS.Deploy.CLI.IntegrationTests/Helpers/Utilities.cs
```diff
@@ -0,0 +1,41 @@
...
+        /// It also adds a nuget.config file that references a private nuget-cache. This cache holds the latest (in-development/unreleased) version of AWS.Deploy.Recipes.CDK.Common.nupkg file
+        /// </summary>
+        public static void OverrideDefaultWorkspace(ServiceProvider serviceProvider, string customWorkspace)
+        {
+            var environmentVariableManager = serviceProvider.GetRequiredService<IEnvironmentVariableManager>();
+            environmentVariableManager.SetEnvironmentVariable("AWS_DOTNET_DEPLOYTOOL_WORKSPACE", customWorkspace);
+            Directory.CreateDirectory(customWorkspace);
+
+            var nugetCachePath = Path.Combine(Environment.GetFolderPath(Environment.SpecialFolder.UserProfile), ".aws-dotnet-deploy", "Projects", "nuget-cache");
+            nugetCachePath = nugetCachePath.Replace(Path.DirectorySeparatorChar, '/');
```
ashovlin, line 27: Why are we replacing an OS-agnostic separator with an OS-specific one? Does this test work on both Windows and Linux?

test/AWS.Deploy.CLI.IntegrationTests/Helpers/TestEnvironmentVariableManager.cs
96malhar: There are certain environment variables (like `TEMP` and `TMP`) which we override if a custom workspace is set. These overrides need to recognized by different command line processes like `cdk deploy..`. That us why we set an explicit environment variable

philasmar: After a discussion, the EnvironmentVariableManager is not passed to the CDKProjectHandler. Please add a backlog item for us to do that work. For now, I'm fine with this change.

test/AWS.Deploy.CLI.IntegrationTests/Helpers/Utilities.cs
philasmar: The '/' should work on windows, but if we use '\' in windows, then we will need to escape that character in json. GIven that we ran the test on Windows and it worked, if the the PR check passes on linux, we should be good to go.

ashovlin: Okay, fine if both Windows and Linux passed.

ashovlin (approved):

PR #660, opened 2022-07-01
ci: Add default encryption for CodeBuild bucket

dscpinheiro:
*Description of changes:* Enables server-side encryption with Amazon S3-managed keys (SSE-S3) for `CodeBuildArtifactsBucket`.


philasmar (changes requested):
Could you please rebase your branch on top of `dev`? Also, after merging this in, we need to go update the existing stack with this new template.

PR #676, opened 2022-07-21
chore: Add integ test confirming server mode connect behavior

normj:
*Description of changes:*
Server mode has the app sec requirement to only be connected via `127.0.0.1` or `localhost`. This restriction has already been implemented in Server mode [here](https://github.com/aws/aws-dotnet-deploy/blob/main/src/AWS.Deploy.CLI/ServerMode/Startup.cs#L31). This PR just adds the integ test to make sure we keep the required behavior.

### DIFF
--- a/test/AWS.Deploy.CLI.IntegrationTests/ServerModeTests.cs
+++ b/test/AWS.Deploy.CLI.IntegrationTests/ServerModeTests.cs
@@ -5,6 +5,8 @@ using System;
 using System.Collections.Generic;
 using System.IO;
 using System.Linq;
+using System.Net;
+using System.Net.Http;
 using System.Security.Cryptography;
 using System.Text;
 using System.Threading;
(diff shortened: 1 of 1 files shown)


test/AWS.Deploy.CLI.IntegrationTests/ServerModeTests.cs
```diff
@@ -62,6 +64,39 @@ public Task<AWSCredentials> ResolveCredentials()
...
+        /// </summary>
+        /// <returns></returns>
+        [Fact]
+        public async Task ConfirmLocalhostOnly()
+        {
+            var portNumber = 4900;
+            var serverCommand = new ServerModeCommand(_serviceProvider.GetRequiredService<IToolInteractiveService>(), portNumber, null, true);
+            var cancelSource = new CancellationTokenSource();
+
+            var serverTask = serverCommand.ExecuteAsync(cancelSource.Token);
```
floralph, line 79: I don't see serverTask used anywhere below, was this intentional or did it's usage get missed?

ashovlin (approved):

test/AWS.Deploy.CLI.IntegrationTests/ServerModeTests.cs
normj: You are correct I was never using the `serverTask` variable nor do I need to. The lifecycle of the `serverCommand` is controlled by the cancellation token. I removed the variable declaration for `serverTask`.

philasmar (approved):

PR #678, opened 2022-07-28
fix: deployments created in preview version don't work in GA version

philasmar:
*Issue #, if available:*
DOTNET-6124

*Description of changes:*
The issue was caused by a breaking change we introduced when we changed the Type of the Security group settings in the fargate recipe. The type changed from `string` to `List` which broke the workflow when the tool tried to apply previous settings on redeployments. This change skips broken settings when applying previous settings and it gives users a message that we were unable to set that setting. I also added a validation error when this happens so that the VS Toolkit tells the user that there is an issue with one of the settings, giving the user the option to either fix the issue before deployment.

### DIFF
--- a/src/AWS.Deploy.Orchestration/Orchestrator.cs
+++ b/src/AWS.Deploy.Orchestration/Orchestrator.cs
@@ -151,6 +151,8 @@ namespace AWS.Deploy.Orchestration
         {
             if (_optionSettingHandler == null)
                 throw new InvalidOperationException($"{nameof(_optionSettingHandler)} is null as part of the orchestartor object");
+            if (_interactiveService == null)
+                throw new InvalidOperationException($"{nameof(_interactiveService)} is null as part of the orchestrator object");
 
             var recommendationCopy = recommendation.DeepCopy();
             recommendationCopy.IsExistingCloudApplication = true;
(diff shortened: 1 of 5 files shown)


src/AWS.Deploy.Orchestration/Orchestrator.cs
```diff
@@ -159,7 +161,15 @@ public async Task<Recommendation> ApplyRecommendationPreviousSettings(Recommenda
...
                 if (previousSettings.TryGetValue(optionSetting.Id, out var value))
                 {
-                    await  _optionSettingHandler.SetOptionSettingValue(recommendationCopy, optionSetting, value, skipValidation: true);
+                    try
+                    {
+                        await _optionSettingHandler.SetOptionSettingValue(recommendationCopy, optionSetting, value, skipValidation: true);
+                    }
+                    catch (UnsupportedOptionSettingType ex)
+                    {
+                        _interactiveService.LogInfoMessage($"Unable to retrieve value of '{optionSetting.Name}' from previous deployment. Make sure to set it again prior to redeployment.");
```
normj, line 170: Change `LogInfoMessage` to `LogErrorMessage`.

normj (approved):
Approving assuming you make the one minor change suggested. Preemptively approving since I'm going on PTO soon and I don't want you blocked by me.

ashovlin (approved):

src/AWS.Deploy.Orchestration/Orchestrator.cs
philasmar: Fixed

Issue #675, opened 2022-07-18
Deploy fails to get credentials with SSO profile


Aracron:
### Describe the bug

When attempting to deploy via the command line, the tool fails to get the credentials for an SSO profile, and is unable to proceed.

### Expected Behavior

The tool should be able to deploy with SSO profiles.

### Current Behavior

When calling "dotnet aws deploy", and choosing an SSO profile, the tool fails with "Unable to create AWS credentials for profile".  If I pass the profile name instead, I get "Failed to get credentials for profile".

If I choose a non-SSO profile instead, the tool proceeds to the next step.

### Reproduction Steps

Attempt to deploy project to AWS via "dotnet aws deploy" command line using an SSO profile.

### Additional Information/Context

We have standardized on using SSO for custom development/deployment to AWS.  

### Version used

1.0.3

### Operating System and version

Windows 10

ashovlin:
Thank you for the report.

We don't include the SSO dependencies from the .NET SDK (which this tool uses under the hood), and we're swallowing the `Assembly AWSSDK.SSOOIDC could not be found or loaded exception` here:
https://github.com/aws/aws-dotnet-deploy/blob/48ef3bc4eef59896c2b17db5b6895fa4c3a575b1/src/AWS.Deploy.CLI/AWSUtilities.cs#L120-L124

I'm not aware of a way to work around it, and have created a ticket tracking this issue in our internal backlog.

ashovlin:
This should be fixed in [version 1.3.7](https://github.com/aws/aws-dotnet-deploy/releases/tag/1.3.7) of the CLI released yesterday.

This hasn't been fixed in the Visual Studio Toolkit yet, but should in a future release when they update the internal version of the deploy tool.

ashovlin:
Per the [Visual Studio Toolkit's changelog](https://github.com/aws/aws-toolkit-visual-studio/blob/main/CHANGELOG.md#13600-2022-10-12), this should now be addressed in their version 1.36.0.0 and beyond.

PR #696, opened 2022-09-07
Support deploying using AWS IAM Identity Center single sign-on (SSO) credentials

ashovlin:
*Issue #, if available:* #675 and DOTNET-6153

*Description of changes:* This introduces the SDK's SSO and SSOOIDC packages to support profiles configured to use [SSO](https://docs.aws.amazon.com/sdkref/latest/guide/access-sso.html) credentials.

We were previously swallowing the `Assembly AWSSDK.SSO could not be found or loaded` exception here:
https://github.com/aws/aws-dotnet-deploy/blob/48ef3bc4eef59896c2b17db5b6895fa4c3a575b1/src/AWS.Deploy.CLI/AWSUtilities.cs#L120-L124

I've also updated our internal attribution tool and regenerated the THIRD_PARTY_LICENSES file.

I did not introduce new automated testing, we don't have long-lived SSO setup in our test account and it'd require a fair amount of manipulation to set everything up automatically. **Let me know if you think it'd be worth it though.**

I did manually test via both the CLI and by replacing the deploy tool within VS Toolkit 1.35.2.0.
* https://docs.aws.amazon.com/cli/latest/userguide/cli-configure-sso.html

### DIFF
--- a/THIRD_PARTY_LICENSES
+++ b/THIRD_PARTY_LICENSES
@@ -21,6 +21,8 @@
 ** Amazon.JSII.Runtime; version 1.54.0 -- https://www.nuget.org/packages/Amazon.JSII.Runtime
 ** AWSSDK.CloudControlApi; version 3.7.2 -- https://www.nuget.org/packages/AWSSDK.CloudControlApi/
 ** AWSSDK.SimpleSystemsManagement; version 3.7.16 -- https://www.nuget.org/packages/AWSSDK.SimpleSystemsManagement/
+** AWSSDK.SSO; version 3.7.0.201 -- https://www.nuget.org/packages/AWSSDK.SSO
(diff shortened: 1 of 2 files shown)


THIRD_PARTY_LICENSES
```diff
@@ -344,7 +348,9 @@ Copyright (c) .NET Foundation and Contributors
 Copyright (c) .NET Foundation and Contributors
 ** Microsoft.OpenApi; version 1.2.3 -- https://www.nuget.org/packages/Microsoft.OpenApi/
 Copyright (c) Microsoft Corporation.
-
+** System.Text.Json; version 6.0.4 -- https://www.nuget.org/packages/System.Text.Json
```
ashovlin, line 351: This moved when I regenerated the whole file from the internal attribution tool.

normj (approved):

philasmar (approved):

Issue #153, opened 2021-03-16
Redeploy: DeploymentBundle Settings are lost 


ppittle:
**Describe the bug**
DeploymnetBundle Settings like `DotnetPublishBuildConfiguration` and `DotnetPublishAdditionalBuildArguments` in Elastic Beanstalk recipe and `DockerBuildArgs` in Fargate deployments are not persisted.  When doing a redeploy, you have to re-endter these values.

**To Reproduce**
1. In a web project run `dotnet aws deploy`. 
2. Select Beanstalk
3. Customize the `DotnetPublishAdditionalBuildArguments`: `"/p:DefineConstants=TEST"`
4. Deploy
5. Run `dotnet aws deploy`. 
6. Select previously deployed stack
7. Notice that `DotnetPublishAdditionalBuildArguments` is blank.

**Expected behavior**
`DotnetPublishAdditionalBuildArguments`  should have value  `"/p:DefineConstants=TEST"`

**Screenshots**
![image](https://user-images.githubusercontent.com/1190907/111376271-7c922d80-865c-11eb-8063-f451e2b63434.png)

![image](https://user-images.githubusercontent.com/1190907/111376329-8caa0d00-865c-11eb-9f6e-bb837d64bea4.png)

**Environment (please complete the following information):**
 - CLI Version: [eg. 1.0.0] `0.1.1-beta.20210315205437.06db70a`
 - Include the output of `dotnet --info` `5.0.201`
 - OS Info: `Windows 10`

**Additional context**

---

This is a :bug: bug report

ashishdhingra:
Appears to be fixed in version `1.10.4`:
```
Current settings (select number to change its value)
----------------------------------------------------
1 . EC2 Instance Type: 
2 . Beanstalk Platform: arn:aws:elasticbeanstalk:us-east-1::platform/.NET Core running on 64bit Amazon Linux 2/2.5.0
3 . Managed Platform Updates: 
	Enable Managed Platform Updates: True
	Preferred Start Time: Sun:00:00
	Update Level: minor
4 . Enable AWS X-Ray Tracing Support: False
5 . Reverse Proxy: nginx
6 . Environment Variables: 
7 . Virtual Private Cloud (VPC): 
	Use a VPC : False
8 . Dotnet Build Configuration: Release
9 . Dotnet Publish Args: /p:DefineConstants=TEST
10. Self Contained Build: False
```

PR #701, opened 2022-09-14
fix: Persist deployment bundle settings across deployments

96malhar:
*Issue #, if available:*
* DOTNET-6032
* https://github.com/aws/aws-dotnet-deploy/issues/153

*Description of changes:*
This PR persists deployment bundle option settings across re-deployments.

It achieves this by writing the deployment bundle option settings in the CloudFormation template metadata under the `aws-dotnet-deploy-deployment-bundle-settings` property.
image

**Note** - All file and directory paths are written relative to the user's .NET project directory path.

### DIFF
--- a/src/AWS.Deploy.Orchestration/CdkAppSettingsSerializer.cs
+++ b/src/AWS.Deploy.Orchestration/CdkAppSettingsSerializer.cs
@@ -4,19 +4,30 @@
 using System.Collections.Generic;
 using System.IO;
 using AWS.Deploy.Common;
+using AWS.Deploy.Common.IO;
 using AWS.Deploy.Common.Recipes;
 using AWS.Deploy.Recipes.CDK.Common;
 using Newtonsoft.Json;
 
 namespace AWS.Deploy.Orchestration
 {
-    public class CdkAppSettingsSerializer
+    public interface ICdkAppSettingsSerializer
+    {
+        /// <summary>
+        /// Creates the contents for the appsettings.json file inside the CDK project. This file is deserialized into <see cref="IRecipeProps{T}"/> to be used the by the CDK templates.
+        /// </summary>
+        string Build(CloudApplication cloudApplication, Recommendation recommendation, OrchestratorSession session);
+    }
+
+    public class CdkAppSettingsSerializer : ICdkAppSettingsSerializer
     {
         private readonly IOptionSettingHandler _optionSettingHandler;
+        private readonly IDirectoryManager _directoryManager;
 
-        public CdkAppSettingsSerializer(IOptionSettingHandler optionSettingHandler)
+        public CdkAppSettingsSerializer(IOptionSettingHandler optionSettingHandler, IDirectoryManager directoryManager)
         {
             _optionSettingHandler = optionSettingHandler;
+            _directoryManager = directoryManager;
         }
 
         public string Build(CloudApplication cloudApplication, Recommendation recommendation, OrchestratorSession session)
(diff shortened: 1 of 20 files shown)


src/AWS.Deploy.Recipes.CDK.Common/RecipeProps.cs
```diff
@@ -70,8 +71,9 @@ public interface IRecipeProps<T>
     /// <summary>
     /// A representation of the settings transferred from the AWS .NET deployment tool to the CDK project.
     /// </summary>
-    /// <typeparam name="T"></typeparam>
-    public class RecipeProps<T> : IRecipeProps<T>
```
normj, line 74: These are breaking changes here removing `ECRRepositoryName` and adding a new `DeploymentBundleSetting` expecting the CDK project to understand. You got around this by modifying all of our recipes but if a user has a deployment project then this change would break them when they take an updated version of the tooling. We need to find a way to implement this feature with a breaking change.

normj (changes requested):

src/AWS.Deploy.Recipes.CDK.Common/RecipeProps.cs
96malhar: I added the `DeploymentBundleSettings` in the `RecipeProps` as a `string` and reverted all modifications to the CDKTemplates. These way all persisted CDK projects will still work with the updated version of the tooling.

philasmar (changes requested):

src/AWS.Deploy.Orchestration/CdkAppSettingsSerializer.cs
```diff
@@ -33,24 +39,19 @@ public string Build(CloudApplication cloudApplication, Recommendation recommenda
                 recommendation.Recipe.Version,
                 session.AWSAccountId,
                 session.AWSRegion,
-                new ()
+                settings: Helpers.GetOptionSettingsContainer(_optionSettingHandler, recommendation, session.ProjectDefinition, _directoryManager, OptionSettingsType.Recipe),
+                deploymentBundleSettings: JsonConvert.SerializeObject(deploymentBundleContainer, Formatting.Indented)
```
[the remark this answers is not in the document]
96malhar: The `DeloymentBundleSettings` is defined as a `string` property in the `IRecipeProps`. This property only exists so that it can be written to the CloudFormation template metadata and can be retrieved to preload settings during a re-deployment. This property is not used by the CDKTemplates and therefore it makes sense to not store it as any other concrete type apart from `string`

src/AWS.Deploy.Recipes.CDK.Common/RecipeProps.cs
```diff
@@ -139,15 +149,16 @@ public RecipeProps()
 #nullable restore warnings
 
         public RecipeProps(string stackName, string projectPath, string recipeId, string recipeVersion,
-             string? awsAccountId, string? awsRegion,  T settings)
+             string? awsAccountId, string? awsRegion,  T settings, string deploymentBundleSettings)
```
normj, line 152: Technically this is a breaking change for the recipes by changing the constructor signature. Since we are mostly using this for serialization it might not matter but I could see it causing problems if a user hasn't updated their CLI but did update the reference to `AWS.Deploy.Recipes.CDK.Common` to include the breaking change. The CLI would not be sending over deployment bundles which I think would try and set this non nullable string to null. To be safe we should make this setting nullable and not part of the constructor.

normj (changes requested):

normj (approved):

src/AWS.Deploy.Recipes.CDK.Common/RecipeProps.cs
96malhar: Done

philasmar (approved):

PR #704, opened 2022-09-21
ci: increase codebuild ci role assume duration

philasmar:
*Description of changes:*
The CodeBuild PR check is failing when it hits the 1 hour mark. I have increased the maximum duration of the IAM roles we are assuming to allow for longer builds. This PR increases the duration in seconds for the assume role to allow builds to run longer than the default 60 seconds.

### DIFF
--- a/.github/workflows/codebuild-ci.yml
+++ b/.github/workflows/codebuild-ci.yml
@@ -17,6 +17,7 @@ jobs:
         uses: aws-actions/configure-aws-credentials@v1
         with:
           role-to-assume: ${{ secrets.CI_AWS_ROLE_ARN }}
+          role-duration-seconds: 7200
           aws-region: us-west-2
       - name: Run CodeBuild
         id: codebuild


.github/workflows/codebuild-ci.yml
```diff
@@ -17,6 +17,7 @@ jobs:
         uses: aws-actions/configure-aws-credentials@v1
         with:
           role-to-assume: ${{ secrets.CI_AWS_ROLE_ARN }}
+          role-duration-seconds: 120
```
ashovlin, line 20: Two minutes is adequate? This doesn't have to match the hour that it takes the job to run?

96malhar, line 20: Shouldn't this value be >= 3600 to map it to an hour?

96malhar: Per the docs the default value is 1 hour - https://github.com/aws-actions/configure-aws-credentials#assuming-a-role

ashovlin (approved):

.github/workflows/codebuild-ci.yml
philasmar: Fixed

96malhar (approved):

.github/workflows/codebuild-ci.yml
philasmar: Fixed

Issue #679, opened 2022-08-01
Internal Load Balancer Option from CLI (ECS Fargate)


cbossie:
### Describe the feature

This is a request to include the option for internal or internet-facing load balancer in the "dotnet aws deploy" command advanced settings for Elastic Load Balancer (option 9). The current option is internet facing by default and it is not modifiable via the CLI.

This option would be available only for a new elastic load balancer.

### Use Case

This would enable the deployment of an internal application (e.g. non internet-facing) into a customer's VPC. The default is internet facing, which may not be an acceptable option for some content. Although the user can use and modify a deployment project to achieve this, they currently are not able to specify this setting using the direct CLI.

### Proposed Solution

The solution would be as follows. When creating a new load balancer using the dotnet deployment tool CLI functionality, the default would remain "internet facing", maintaining expectations for users that are already using the tool.

When the user runs "dotnet aws deploy", and selects "3: ASP.NET Core App to Amazon ECS using AWS Fargate" and then types in "more" for Advanced settings, they will select "9" for the elastic load balancer.

The current options look lie the following:
9 . Elastic Load Balancer:
   Create New Load Balancer: True
   Deregistration delay (seconds): 60
   Health Check Path: /
   Health Check Timeout: 5
   Health Check Interval: 30
   Healthy Threshold Count: 5
   Unhealthy Threshold Count: 2

I propose adding an additional option, 'Load Balancer Schema', which would look like one of the following:
"Load Balancer Schema: internet-facing"
or
"Load Balancer Schema: internal"

When the user selects "9" to walk through the settings, if they select "y" for "Create New Load Balancer", then they will have the option to select the load balancer schema. Something like:

Load Balancer Schema:
Do you want the load balancer to be internet facing?
y/n (default y)

Then the rest of parameter selection will proceed as normal.

### Other Information

Ultimately, I believe that this will be a minor-ish change as an update to the ECS Fargate recipe. By adding "internet facing" to the settings object that is consumed in the "ConfigureLoadBalancer" method in the ECS Recipe.cs (https://github.com/aws/aws-dotnet-deploy/blob/main/src/AWS.Deploy.Recipes/CdkTemplates/AspNetAppEcsFargate/Generated/Recipe.cs  -  around line 209), it should be able to override the hard-coded "true" value.

I did a quick test by manually changing that value in the recipe and it appeared to work correctly.

### Acknowledgements

- [ ] I may be able to implement this feature request
- [ ] This feature might incur a breaking change

[...]

PR #706, opened 2022-09-22
 Enable toggling "Internet-Facing" for the ECS Fargate Load Balancer

ashovlin:
*Issue #, if available:* DOTNET-6294 and #679 

*Description of changes:* Adds a new option setting to the ASP.NET ECS Fargate recipe to expose the [internetFacing?](https://docs.aws.amazon.com/cdk/api/v2/docs/aws-cdk-lib.aws_elasticloadbalancing.LoadBalancerProps.html#internetfacing) property in the underlying CDK recipe.
* The GitHub issue suggested exposing `Load Balancer Schema: internet-facing`, but this appears to be mapped to a boolean in CDK.
* I was not able to mark this updatable because the redeployment fails due to https://github.com/aws-cloudformation/cloudformation-coverage-roadmap/issues/780
* Added automated tests that examine the generated CloudFormation template. Did manual testing with both settings, verified that my resulting application was accessible from an instance running in the same VPC but not my laptop.

### DIFF
--- a/src/AWS.Deploy.Recipes/RecipeDefinitions/ASP.NETAppECSFargate.recipe
+++ b/src/AWS.Deploy.Recipes/RecipeDefinitions/ASP.NETAppECSFargate.recipe
@@ -1,7 +1,7 @@
 {
     "$schema": "./aws-deploy-recipe-schema.json",
     "Id": "AspNetAppEcsFargate",
-    "Version": "1.0.1",
+    "Version": "1.1.0",
     "Name": "ASP.NET Core App to Amazon ECS using AWS Fargate",
     "DeploymentType": "CdkProject",
     "DeploymentBundle": "Container",
@@ -702,6 +702,15 @@
                             "Value": "Path"
                         }
                     ]
+                },
+                {
+                    "Id": "InternetFacing",
+                    "Name": "Internet-Facing",
+                    "Description": "Should the load balancer have an internet-routable address? Internet-facing load balancers can route requests from clients over the internet. Internal load balancers can route requests only from clients with access to the VPC.",
+                    "Type": "Bool",
+                    "DefaultValue": true,
+                    "AdvancedSetting": false,
+                    "Updatable": false
                 }
             ]
         },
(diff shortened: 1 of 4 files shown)


src/AWS.Deploy.Recipes/RecipeDefinitions/ASP.NETAppECSFargate.recipe
```diff
@@ -702,6 +702,15 @@
...
                     ]
+                },
+                {
+                    "Id": "InternetFacing",
+                    "Name": "Internet-Facing",
+                    "Description": "Should the load balancer have an internet-routable address? Internet-facing load balancers can route requests from clients over the internet. Internal load balancers can route requests only from clients with access to the VPC.",
+                    "Type": "Bool",
+                    "DefaultValue": true,
+                    "AdvancedSetting": false,
+                    "Updatable": false
```
ashovlin, line 713: Left this at `false` because of https://github.com/aws-cloudformation/cloudformation-coverage-roadmap/issues/780 during redeployment

normj (approved):
I did a manual test of doing a deployment with the current released version and then a redeployment with this branch to ensure the new property didn't reset existing value for `InternetFacing`.

src/AWS.Deploy.Recipes/RecipeDefinitions/ASP.NETAppECSFargate.recipe
```diff
@@ -702,6 +702,15 @@
                             "Value": "Path"
                         }
                     ]
+                },
+                {
+                    "Id": "InternetFacing",
+                    "Name": "Internet-Facing",
+                    "Description": "Should the load balancer have an internet-routable address? Internet-facing load balancers can route requests from clients over the internet. Internal load balancers can route requests only from clients with access to the VPC.",
```
96malhar, line 709: nit: maybe change _"Should the load balancer have an internet-routable address?"_ -> _"Determines if the load balancer has an internet-routable address."_

96malhar (approved):

src/AWS.Deploy.Recipes/RecipeDefinitions/ASP.NETAppECSFargate.recipe
ashovlin: We do use a question for most other `"Bool"` options. For example, all the others in the Fargate recipe:
* "Do you want to create a new ECS cluster?"
* "Do you want to create a new role?"
* "Do you want to use the default VPC for the deployment?"
* "Do you want to create a new VPC?"
* "Do you want to create a new Load Balancer?"
* "Do you want to enable AutoScaling?"

I'm still inclined to match those for consistency, unless you feel strongly.

96malhar: I am fine with the wording if there is precedent for it👍

PR #712, opened 2022-09-29
chore: Truncate long error messages

96malhar:
*Issue #, if available:*
DOTNET-6257

*Description of changes:*
This PR truncates long error messages so that external telemetery systems that are hooked to the deploy tool via server mode are not overloaded.

Currently, the truncation is only performed in the `GetDeploymentStatus` API call. However, I am unsure if this is the only place from where the data is fed into external telemetry systems. The same concern has also been brought up via internal communication.

### DIFF
--- a/src/AWS.Deploy.CLI/ServerMode/Controllers/DeploymentController.cs
+++ b/src/AWS.Deploy.CLI/ServerMode/Controllers/DeploymentController.cs
@@ -615,13 +615,14 @@ namespace AWS.Deploy.CLI.ServerMode.Controllers
                 if (state.DeploymentTask.Exception != null)
                 {
                     var innerException = state.DeploymentTask.Exception.InnerException;
+                    var message = innerException.GetTruncatedErrorMessage();
                     if (innerException is DeployToolException deployToolException)
                     {
-                        output.Exception = new DeployToolExceptionSummary(deployToolException.ErrorCode.ToString(), deployToolException.Message, deployToolException.ProcessExitCode);
+                        output.Exception = new DeployToolExceptionSummary(deployToolException.ErrorCode.ToString(), message, deployToolException.ProcessExitCode);
                     }
                     else
                     {
-                        output.Exception = new DeployToolExceptionSummary(DeployToolErrorCode.UnexpectedError.ToString(), innerException?.Message ?? string.Empty);
+                        output.Exception = new DeployToolExceptionSummary(DeployToolErrorCode.UnexpectedError.ToString(), message);
                     }
                 }
             }
--- a/src/AWS.Deploy.Common/Exceptions.cs
+++ b/src/AWS.Deploy.Common/Exceptions.cs
@@ -318,5 +318,27 @@ namespace AWS.Deploy.Common
 
             return $"{Environment.NewLine}{e.Message}{Environment.NewLine}{e.StackTrace}{PrettyPrint(e.InnerException)}";
         }
+
+        /// <summary>
+        /// Returns the truncated error message by only preserving the leading and trailing k characters.
+        /// The message will only be truncated if its length is greater than 2*k.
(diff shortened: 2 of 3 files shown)


src/AWS.Deploy.CLI/ServerMode/Controllers/DeploymentController.cs
```diff
@@ -615,13 +615,14 @@ public IActionResult GetDeploymentStatus(string sessionId)
                 if (state.DeploymentTask.Exception != null)
                 {
                     var innerException = state.DeploymentTask.Exception.InnerException;
+                    var message = innerException.GetTruncatedErrorMessage();
                     if (innerException is DeployToolException deployToolException)
                     {
-                        output.Exception = new DeployToolExceptionSummary(deployToolException.ErrorCode.ToString(), deployToolException.Message, deployToolException.ProcessExitCode);
+                        output.Exception = new DeployToolExceptionSummary(deployToolException.ErrorCode.ToString(), message, deployToolException.ProcessExitCode);
```
awschristou, line 621: At this point, it looks like the only variable thing is the error code. You could assign `output.Exception` after the if/else clause

src/AWS.Deploy.Common/Exceptions.cs
```diff
@@ -318,5 +318,26 @@ public static string PrettyPrint(this Exception? e)
...
+            if (string.IsNullOrEmpty(message))
+                return string.Empty;
+
+            if (message.Length <= 2*k)
+                return message;
+
+            var firstKChars = message.Substring(0, k);
+            var lastkChars = message.Substring(message.Length - k);
+            var seperator = $"{Environment.NewLine}...{Environment.NewLine}...{Environment.NewLine}...{Environment.NewLine}";
+            return $"This error message is truncated to only include leading and trailing {k} characters:{Environment.NewLine}{firstKChars}{seperator}{lastkChars}";
```
awschristou, line 340: * You're truncating to 1000 characters by default. but you append 83 characters (close to 10% of the length) onto it. Can this be shortened a little?
* Would it make sense to put this in the truncation instead of at the start of the error? Something like
```
error text
error text
...
Error truncated to the first and last 500 characters
...
error text
```

src/AWS.Deploy.CLI/ServerMode/Controllers/DeploymentController.cs
96malhar: We also add the process exit code based on the if/else clause. I prefer to eave this as is for now.

src/AWS.Deploy.Common/Exceptions.cs
96malhar: Changed the truncated message to match the suggestion.

awschristou (approved):

normj (approved):

src/AWS.Deploy.Common/Exceptions.cs
```diff
@@ -318,5 +318,26 @@ public static string PrettyPrint(this Exception? e)
 
             return $"{Environment.NewLine}{e.Message}{Environment.NewLine}{e.StackTrace}{PrettyPrint(e.InnerException)}";
         }
+
+        /// <summary>
+        /// Returns the truncated error message by only preserving the leading and trailing k characters.
+        /// The message will only be truncated if its length is greater than 2*k.
+        /// This method returns <see cref="string.Empty"/> if the error message is null or empty.
+        /// </summary>
+        public static string GetTruncatedErrorMessage(this Exception? e, int k = 500)
```
ashovlin, line 327: Please document parameters using the XML tags, and there's usually a better parameter name than a single letter.

96malhar: Done

ashovlin (approved):

PR #713, opened 2022-10-04
fix: json comments in beanstalk windows manifest caused exception

philasmar:
*Issue #, if available:*
DOTNET-6340

*Description of changes:*
* Added JSON Options to the System.Text.Json deserializer to skip reading comments instead of failing serialization.
* Added exception handling over the deseerialization block to provide user friendly error messages if serialization exceptions happen.

### DIFF
--- a/src/AWS.Deploy.Orchestration/ServiceHandlers/AWSElasticBeanstalkHandler.cs
+++ b/src/AWS.Deploy.Orchestration/ServiceHandlers/AWSElasticBeanstalkHandler.cs
@@ -93,7 +93,7 @@ namespace AWS.Deploy.Orchestration.ServiceHandlers
         {
             try
             {
-                return JsonSerializer.Deserialize<T>(json?.ToString() ?? string.Empty);
+                return JsonSerializer.Deserialize<T>(json?.ToString() ?? string.Empty, new JsonSerializerOptions { ReadCommentHandling = JsonCommentHandling.Skip });
             }
             catch
             {
--- a/src/AWS.Deploy.Orchestration/Utilities/DeployedApplicationQueryer.cs
+++ b/src/AWS.Deploy.Orchestration/Utilities/DeployedApplicationQueryer.cs
@@ -271,19 +271,41 @@ namespace AWS.Deploy.Orchestration.Utilities
 
             if (recipeId.Equals(Constants.RecipeIdentifier.EXISTING_BEANSTALK_WINDOWS_ENVIRONMENT_RECIPE_ID))
             {
-                var manifestPath = Path.Combine(Path.GetDirectoryName(projectPath) ?? string.Empty, Constants.ElasticBeanstalk.WindowsManifestName);
-                if (_fileManager.Exists(manifestPath))
+                var windowsManifest = await GetBeanstalkWindowsManifest(projectPath);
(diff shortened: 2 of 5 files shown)


src/AWS.Deploy.Orchestration/Utilities/DeployedApplicationQueryer.cs
```diff
@@ -274,11 +274,24 @@ private async Task<List<CloudApplication>> GetExistingBeanstalkEnvironments()
...
+                        });
+                        if (manifest.Deployments.AspNetCoreWeb.Count != 0)
+                        {
+                            optionSettings[Constants.ElasticBeanstalk.IISWebSiteOptionId] = manifest.Deployments.AspNetCoreWeb[0].Parameters.IISWebSite;
+                            optionSettings[Constants.ElasticBeanstalk.IISAppPathOptionId] = manifest.Deployments.AspNetCoreWeb[0].Parameters.IISPath;
+                        }
+                    }
+                    catch (Exception ex)
+                    {
+                        throw new InvalidWindowsManifestFileException(
```
awschristou, line 291: This exception seems like it is only applicable to the Deserialize call - I would imagine that any exception arising after Deserialize would be a different problem. Should the deserialize call and the try-catch that raises InvalidWindowsManifestFileException be moved to a dedicated function?

philasmar: I separated the IO into a different function.

awschristou (approved):

test/AWS.Deploy.Orchestration.UnitTests/DeployedApplicationQueryerTests.cs
```diff
@@ -471,8 +471,38 @@ public async Task GetPreviousSettings_BeanstalkEnvironment()
...
+                      ""iisPath"": ""/path"",
+                      ""iisWebSite"": ""Default Web Site Custom""
+                    }
+                  }
+                ]
+              }
+            }")]
+        [InlineData(@"{
+              ""manifestVersion"": 1,
+              // comments
```
ashovlin, line 492: Can you assert that we're not deleting the user's comments when we write it back out?

src/AWS.Deploy.Orchestration/ServiceHandlers/AWSElasticBeanstalkHandler.cs
```diff
@@ -93,7 +93,7 @@ private T GetOrCreateNode<T>(object? json) where T : new()
         {
             try
             {
-                return JsonSerializer.Deserialize<T>(json?.ToString() ?? string.Empty);
+                return JsonSerializer.Deserialize<T>(json?.ToString() ?? string.Empty, new JsonSerializerOptions { ReadCommentHandling = JsonCommentHandling.Skip });
```
ashovlin, line 96: Are we overwriting and effectively deleting the user's comments when we write the file back out?

If so, should we not do that?

philasmar: Yes we are, but it shouldn't matter since we are not replacing the manifest file in the user's project. The updated file is only added to the .net zip deployment bundle.

ashovlin (approved):

96malhar (approved):

PR #716, opened 2022-10-05
Enable specifying one or more subnets when configuring the VPC settings for ECS Fargate recipes [HOLD FOR REWORK]

ashovlin:
*Issue #, if available:* DOTNET-6065

*Description of changes:* This allows users to choose a subset of a VPC's subnets when configuring the three ECS Fargate recipes.
* Subnets are required in the other recipes, but I *did not* do so here because it would be a backwards incompatible change.
    * If the user does not specify subnets when selecting a VPC, we will continue to fallback on the CDK construct's defaulting logic. I added this to the option description.
* The three ECS Fargate recipes did already have a security group setting, but it's not a child setting of VPC like Beanstalk or App Runner. Left this alone, also for backwards compatibility. 

~~I added unit tests for expanded type hint~~, but did not add automated tests that exercise a full CDK deployment.
* It doesn't seem like we have a lot of these, so I don't believe we've been adding them as we add new option settings. Let me know if you would like to see some though.

----
*10/28/22 Update:* Added a commit reworking this to add subnets as their own option setting, not as a child option under VPC. This should allow Visual Studio Toolkit to pick up this change automatically. However it does introduce some strange behavior for an edge case that I'm out of ideas to solve without a breaking change, see https://github.com/aws/aws-dotnet-deploy/pull/716#discussion_r1008222641.


src/AWS.Deploy.Orchestration/OptionSettingHandler.cs
```diff
@@ -210,8 +210,8 @@ public OptionSettingItem GetOptionSetting(RecipeDefinition recipe, string? jsonP
                 optionSetting = optionSettings.FirstOrDefault(os => os.Id.Equals(ids[i]));
                 if (optionSetting == null)
                 {
-                    throw new OptionSettingItemDoesNotExistException(DeployToolErrorCode.OptionSettingItemDoesNotExistInRecipe, $"An option setting item with the specified fully qualified Id '{jsonPath}' cannot be found in the" +
-                    $" '{recipe.Name}' recipe.");
+                    throw new OptionSettingItemDoesNotExistException(DeployToolErrorCode.OptionSettingItemDoesNotExistInRecipe, $"An option setting item with the Id '{ids[i]}' " +
```
ashovlin, line 213: I had a typo in my recipe which tripped my up for a bit, this would have helped. It's otherwise unrelated to the specific change.

96malhar (approved):

ashovlin:
Note to reviewers: Based on internal discussions, planning to rework this to make the new subnet options a sibling of VPC, instead of a child. That'll preserve the intent of the current `Vpc`/`VpcCommand` typehint, which will not require changes in the Visual Studio Toolkit.

src/AWS.Deploy.Common/Recipes/Validation/OptionSettingItemValidators/SubnetsInVpcValidator.cs
```diff
@@ -39,15 +51,49 @@ public async Task<ValidationResult> Validate(object input, Recommendation recomm
...
+            if (!string.IsNullOrEmpty(DefaultVpcOptionPath))
+            {
+                var useDefaultVpcSetting = _optionSettingHandler.GetOptionSetting(recommendation, DefaultVpcOptionPath);
+                var useDefaultVpcValue = _optionSettingHandler.GetOptionSettingValue<bool>(recommendation, useDefaultVpcSetting);
+
+                // For at least the ECS Fargate recipes, it's possible for the user to:
+                //   1. Select a non-default, existing VPC
+                //   2. Then set "Default VPC?" to true
+                // Because we have the default VPC selection take precedence in CDK projects over an explicit VPC Id,
+                // we want to validate the subnets against the default VPC in this case.
```
ashovlin, line 63: The user can get in a weird state that looks like this:
image

This block is my proposed fix, but it still requires the use to redo their selections.

I did explore setting the `VpcId` option to the default VPC automatically when the radio box is checked:
* I tried to do this in a new `Validator`, because I'm not aware of a precedent for options changing each others' settings based on some values.
* It would require a new network request from that validator to lookup the default VPC ID again.
* But it wouldn't update the Toolkit UI  - they'd still see the previously-selected VPC in the disabled box.

normj: I'm surprised the subnets even show up if `Default VPC` is selected. Should we update the recipe `Depends On` conditions to have the IsDefaultVPC = `false`.

ashovlin: I don't really have a strong opinion. Both the Beanstalk and App Runner recipes allow you to select subnets even if the default VPC is selected.

PR #723, opened 2022-11-07
chore: remove .NET Core 3.1 support due to end of life approaching

philasmar:
*Issue #, if available:*
DOTNET-6486

*Description of changes:*
* Remove support for .NET Core 3.1 due to it's end of life next month
* Add support for Apple Silicon Macs with Arm version of NET 6 installed (since .NET Core 3.1 has no ARM support)
* Update SDK references to the latest available


src/AWS.Deploy.ServerMode.Client/AWS.Deploy.ServerMode.Client.csproj
```diff
@@ -1,7 +1,7 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
-    <TargetFramework>netstandard2.0</TargetFramework>
+    <TargetFramework>netstandard2.1</TargetFramework>
```
normj, line 4: This is used by the toolkit which can only use .NET Standard 2.0. We need to not move this to `netstandard2.1`. You might want to add a comment here why we have to keep this at .NET Standard 2.0 to prevent somebody from attempting to change this in the future.

normj (changes requested):

README.md
```diff
@@ -5,7 +5,7 @@ This repository contains the AWS Deploy Tool for .NET CLI - the opinionated tool
...
 
-The goal of the deployment tool is to deploy cloud-native .NET applications that are built with .NET Core 3.1 and above. A cloud-native .NET application is written in .NET with the intent to deploy to Linux. It is not tied to any Windows specific technology such as Windows registry, IIS or MSMQ, and can be deployed on virtualized compute. The tool **cannot** be used to deploy .NET Framework, Desktop, Xamarin, or other applications that do not fit the "cloud-native" criteria.
+The goal of the deployment tool is to deploy cloud-native .NET applications that are built with .NET 6 and above. A cloud-native .NET application is written in .NET with the intent to deploy to Linux. It is not tied to any Windows specific technology such as Windows registry, IIS or MSMQ, and can be deployed on virtualized compute. The tool **cannot** be used to deploy .NET Framework, Desktop, Xamarin, or other applications that do not fit the "cloud-native" criteria.
```
ashovlin, line 8: We're not actually dropping support for _deploying_ .NET Core 3.1 applications yet, right? I don't see the corresponding change in the `MSProperty` -> `TargetFramework` rules.

I think the docs should remain with .NET Core 3.1 as the base as long as the recipes still support it.

src/AWS.Deploy.ServerMode.Client/AWS.Deploy.ServerMode.Client.csproj
philasmar: Done

README.md
philasmar: I agree. I left mention of netcoreapp3.1 in the docs.

src/AWS.Deploy.Recipes/CdkTemplates/AspNetAppAppRunner/AspNetAppAppRunner.csproj
```diff
@@ -26,7 +26,7 @@
     <!-- CDK Construct Library dependencies -->
     <PackageReference Include="Amazon.CDK.Lib" Version="2.43.1" />
 
-    <PackageReference Include="Microsoft.Extensions.Configuration.Json" Version="3.1.7" />
+    <PackageReference Include="Microsoft.Extensions.Configuration.Json" Version="6.0.0" />
```
ashovlin, line 29: Throughout, for any `PackageReference` changes please update both the internal license tracking tool and then regenerate THIRD_PARTY_LICENSES.

test/AWS.Deploy.CLI.IntegrationTests/BlazorWasmTests.cs
```diff
@@ -53,8 +53,7 @@ public BlazorWasmTests()
         }
 
         [Theory]
-        [InlineData("testapps", "BlazorWasm31", "BlazorWasm31.csproj")]
-        [InlineData("testapps", "BlazorWasm50", "BlazorWasm50.csproj")]
+        [InlineData("testapps", "BlazorWasm60", "BlazorWasm60.csproj")]
```
ashovlin, line 56: Should we leave 3.1 and 5 under test since we're not removing support for deploying them as far as I understand?

src/AWS.Deploy.Common/Recipes/Validation/OptionSettingItemValidators/RequiredValidator.cs
```diff
@@ -19,11 +19,11 @@ public class RequiredValidator : IOptionSettingItemValidator
         public Task<ValidationResult> Validate(object input, Recommendation recommendation, OptionSettingItem optionSettingItem)
         {
             var message = ValidationFailedMessage.Replace("{{OptionSetting}}", optionSettingItem.Name);
-            if (input?.TryDeserialize<SortedSet<string>>(out var inputList) ?? false)
+            if (input?.TryDeserialize<SortedSet<string>>(out var inputList) ?? false && inputList != null)
             {
                 return Task.FromResult<ValidationResult>(new()
                 {
-                    IsValid = inputList.Any(),
+                    IsValid = inputList!.Any(),
```
ashovlin, line 26: Why do we need to suppress here with the null check added above?

ashovlin (changes requested):

test/AWS.Deploy.CLI.IntegrationTests/BlazorWasmTests.cs
normj: I think the value of testing each framework as an integ test is minimal. We don't do anything different for whatever framework being targeted. I'm fine not having the 3.1 and 5.0 tests.

normj (approved):

src/AWS.Deploy.Common/Recipes/Validation/OptionSettingItemValidators/RequiredValidator.cs
philasmar: even though there is a null check on line 22, visual studio is still indicating that this could be null, which is why I suppressed it.

src/AWS.Deploy.Recipes/CdkTemplates/AspNetAppAppRunner/AspNetAppAppRunner.csproj
philasmar: Done

ashovlin (approved):

PR #729, opened 2022-11-29
docs: Update credentials documentation to discuss assuming CDK deployment roles, and document the --output parameter on the deployment-project command.

ashovlin:
*Issue #, if available:* DOTNET-6524, DOTNET-6525

*Description of changes:* Updates the documentation site in two places:
1. We were missing the `--output` parameter on the `dotnet aws deployment-project generate` page
     * We also had some parameters that don't exist for that command listed, removed them.
3. When deploying with CDK V2, the deployment process assumes roles that were created by the CDK bootstrap. Our permissions page lists managed policies that help the user get through service calls we make during defaulting, type hints, validation, but didn't mention the need to assume roles during deployment.

     STS doesn't appear to provide managed policies, so I added an example of an inline or custom policy that will accomplish this.

### DIFF
--- a/site/content/docs/commands/project.md
+++ b/site/content/docs/commands/project.md
@@ -1,20 +1,26 @@
 # deployment-project generate command
 
 ### Usage
-    dotnet aws deployment-project - Generates and saves the deployment CDK project in the user provided location.
+
+`dotnet aws deployment-project` - Generates and saves the deployment CDK project in the user provided location.
 
 ### Synopsis
-    dotnet aws deployment-project generate [-d|--diagnostics] [-s|--silent] [--profile <PROFILE>] [--region <REGION>] [--project-path <PROJECT-PATH>] [--project-display-name <DISPLAY-NAME>]
+
+```
+dotnet aws deployment-project generate [-o|--output <OUTPUT-PATH>] [-d|--diagnostics] [--project-path <PROJECT-PATH>] [--project-display-name <DISPLAY-NAME>]
+```
 
 ### Description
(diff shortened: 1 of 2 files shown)


site/content/docs/commands/project.md
```diff
@@ -7,18 +7,19 @@
...
 ```
 
 ### Description
-Generates and saves the [deployment CDK project](../deployment-projects/index.md) in a user provided directory path without proceeding with a deployment. Allows user to customize the CDK project before deploying the application.
+Generates and saves the [deployment CDK project](../deployment-projects/index.md) in a user-provided directory without proceeding with a deployment. This allows the user to customize the CDK project before deploying the application.
 
-* The `--output` switch sets the directory where the deployment project will be saved.
-* The `--project-display-name` switch sets the name that will be shown when the .NET project is being deployed.
+* The `--output` switch sets the directory where the CDK deployment project will be saved.
+* The `--project-path` switch sets the path to the project to deploy.
```
normj, line 17: replace `the project to deploy` to `the project to generate a deployment project for` since we aren't deploying the project.

normj (approved):
One minor nit but otherwise looks good and approving.

site/content/docs/commands/project.md
ashovlin: Agreed, updated with your suggestion.

I had originally just copy+pasted what we had here to match the CLI's `--help`, but looks like we're using a shared string there across all commands. It could probably be improved, but would rather keep this PR focused on the docs site.
https://github.com/aws/aws-dotnet-deploy/blob/3bc852c68eaad581ab92a12cae8c6d93625e72ee/src/AWS.Deploy.CLI/Commands/CommandFactory.cs#L42

normj (approved):

philasmar (approved):

PR #735, opened 2022-12-16
Validate security groups belong to the selected VPC for the ECS Fargate recipes

ashovlin:
*Issue #, if available:* DOTNET-6329

*Description of changes:* This applies the existing `SecurityGroupsInVpcValidator` validator to the security groups option in the ASP.NET and console service ECS Fargate recipes. 

Previously it was possible to select security groups that don't match the selected VPC, then the CloudFormation deployment would fail.
```
Stack Deployments Failed: Error: The stack named FargateExtraSecurityGroups failed creation, it may need to be manually deleted from the AWS console: ROLLBACK_COMPLETE: Resource handler returned message: "Invalid request provided: CreateService error: Security group sg-<redacted> does not appear to belong to the same VPC as the input subnets.
```

Caveats:
* We don't have a security groups option in the ECS Fargate scheduled task recipe currently
* Because the console service recipe is still collecting security groups as a comma-separated string instead of a `List`, I expanded the validator to handle that input as well.

### DIFF
--- a/src/AWS.Deploy.Common/Recipes/Validation/OptionSettingItemValidators/SecurityGroupsInVpcValidator.cs
+++ b/src/AWS.Deploy.Common/Recipes/Validation/OptionSettingItemValidators/SecurityGroupsInVpcValidator.cs
@@ -16,7 +16,17 @@ namespace AWS.Deploy.Common.Recipes.Validation
     public class SecurityGroupsInVpcValidator : IOptionSettingItemValidator
     {
         private static readonly string defaultValidationFailedMessage = "The selected security groups are not part of the selected VPC.";
+
+        /// <summary>
+        /// Path to the OptionSetting that stores a selected Vpc Id
+        /// </summary>
         public string VpcId { get; set; } = "";
+
+        /// <summary>
+        /// Path to the OptionSetting that determines if the default VPC should be used
+        /// </summary>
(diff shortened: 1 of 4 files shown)


philasmar (commented):
The change looks good to me. Just wondering about the Console Task recipe, and this is outside the scope of this PR, but it would be nice, if it doesn't take much time, to align the service recipe and task recipe by adding the `ECSServiceSecurityGroups` option setting to the task recipe (if applicable).

philasmar (approved):

96malhar (approved):

ashovlin:
> The change looks good to me. Just wondering about the Console Task recipe, and this is outside the scope of this PR, but it would be nice, if it doesn't take much time, to align the service recipe and task recipe by adding the `ECSServiceSecurityGroups` option setting to the task recipe (if applicable).

We discussed internally (I think it was on 12/22?) - I think we want to avoid any new `ECSServiceSecurityGroups`/`String` (comma seperated IDs) now that we have `ECS Service Security Groups`/`ExistingSecurityGroups` (the proper list) in the Beanstalk and App Runner recipes. But I'm not sure if we want to introduce that inconsistency amongst the Fargate recipes, or else try to move all three at once. Not going to address in this PR, but will create an internal backlog item.

PR #737, opened 2022-12-20
Address semgrep issues around Path.Combine and build warnings

ashovlin:
*Issue #, if available:* DOTNET-6361

*Description of changes:* Addresses warnings from both semgrep and the .NET build.
* For the `Path.Combine` - they were either cases where we want to support relative paths, or else after reviewing code appear to be a false-positive.
* Commit message is intentionally not following conventional commits to avoid a changelog entry as there are no user-facing changes.

### DIFF
--- a/test/AWS.Deploy.CLI.Common.UnitTests/AWS.Deploy.CLI.Common.UnitTests.csproj
+++ b/test/AWS.Deploy.CLI.Common.UnitTests/AWS.Deploy.CLI.Common.UnitTests.csproj
@@ -1,4 +1,4 @@
-<Project Sdk="Microsoft.NET.Sdk">
+﻿<Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
     <TargetFramework>net6.0</TargetFramework>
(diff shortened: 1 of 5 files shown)


test/AWS.Deploy.CLI.Common.UnitTests/AWS.Deploy.CLI.Common.UnitTests.csproj
```diff
@@ -33,8 +33,6 @@
   <ItemGroup>
     <PackageReference Include="Microsoft.NET.Test.Sdk" Version="16.5.0" />
     <PackageReference Include="Should-DotNetStandard" Version="1.0.0" />
-    <PackageReference Include="xunit" Version="2.4.0" />
-    <PackageReference Include="xunit.runner.visualstudio" Version="2.4.0" />
```
ashovlin, line 37: These are duplicated up on lines 21+22.

test/AWS.Deploy.CLI.UnitTests/Utilities/MockPaginatedEnumerable.cs
```diff
@@ -20,10 +20,10 @@ public MockPaginatedEnumerable(T[] data)
 
         public IAsyncEnumerator<T> GetAsyncEnumerator(CancellationToken cancellationToken = default)
         {
-            return new MockAsyncEnumerator<T>(_data);
+            return new MockAsyncEnumerator(_data);
         }
 
-        class MockAsyncEnumerator<T> : IAsyncEnumerator<T>
```
ashovlin, line 26: This addresses:
```
Type parameter 'T' has the same name as the type parameter from outer type 'MockPaginatedEnumerable<T>'
```

philasmar (approved):

96malhar (approved):

PR #745, opened 2023-01-13
ci: doc generator step was failing to push if there are no changes

philasmar:
*Description of changes:*
ci: doc generator step was failing to push if there are no changes
The Doc generator GitHub workflow failed if there are no doc changes in the release. This changes makes it so that the Workflow checks for changes first before proceeding.

### DIFF
--- a/.github/workflows/doc-builder.yml
+++ b/.github/workflows/doc-builder.yml
@@ -40,7 +40,16 @@ jobs:
         run: |
           cd ./src/AWS.Deploy.DocGenerator
           dotnet run --project ./AWS.Deploy.DocGenerator.csproj
+      - name: Check For Git Untracked Changes
+        id: gitcheck
+        shell: pwsh
+        run: |
+          $newFiles=$(git ls-files --others --exclude-standard | wc -l)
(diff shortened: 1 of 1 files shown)


.github/workflows/doc-builder.yml
```diff
@@ -40,7 +40,16 @@ jobs:
         run: |
           cd ./src/AWS.Deploy.DocGenerator
           dotnet run --project ./AWS.Deploy.DocGenerator.csproj
+      - name: Check For Git Untracked Changes
+        id: gitcheck
+        shell: pwsh
+        run: |
+          $newFiles=$(git ls-files --others --exclude-standard | wc -l)
```
96malhar, line 47: Should we restrict the checks to only `.md` files?

philasmar: The following step only adds the doc site. No other changes will be committed.

96malhar (approved):

ashovlin (approved):

ashovlin:
Merged before CI finished since this is to unblock a GitHub action.

PR #756, opened 2023-03-03
fix: Set container memory limit to the same as the task to ensure the…

normj:
*Description of changes:*
As part of investigating [this memory](https://github.com/dotnet/runtime/issues/82815) issue with AWS Fargate it was discovered that .NET has incorrect understanding of the amount of heap space when only setting the task memory size. .NET looks at the containers hard memory limit which is not the same as the task memory.

The change is set our configured memory for both the task and container. We can do that in this tooling because we are only deploying a single container as part of the task definition.

### DIFF
--- a/test/AWS.Deploy.CLI.IntegrationTests/AWS.Deploy.CLI.IntegrationTests.csproj
+++ b/test/AWS.Deploy.CLI.IntegrationTests/AWS.Deploy.CLI.IntegrationTests.csproj
@@ -1,4 +1,4 @@
-﻿<Project Sdk="Microsoft.NET.Sdk">
+<Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
     <TargetFramework>net6.0</TargetFramework>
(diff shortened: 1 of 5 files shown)


edwinloubser:
@normj would these changes resolve the issue for Lambda as well?

normj:
@edwinloubser This change won't affect Lambda because it doesn't handle Lambda deployment. Using the `DOTNET_GCHeapHardLimit` environment variable in Lambda will help .NET in Lambda have a better understanding of the amount of available memory for the GC.

ashovlin (approved):

test/AWS.Deploy.CLI.IntegrationTests/AWS.Deploy.CLI.IntegrationTests.csproj
```diff
@@ -8,6 +8,7 @@
 
   <ItemGroup>
     <PackageReference Include="AWSSDK.CloudWatchLogs" Version="3.7.101.7" />
+    <PackageReference Include="AWSSDK.ECS" Version="3.7.105.4" />
```
philasmar, line 11: Could you please update the attribution document to this new version of ECS? or use the version we are already referencing? `3.7.100.14`

philasmar (changes requested):

test/AWS.Deploy.CLI.IntegrationTests/AWS.Deploy.CLI.IntegrationTests.csproj
normj: Reverted to use the same version we are using the deploy tool itself `3.7.100.14`

philasmar (approved):

edwinloubser:
@normj understood and thank you for confirming.

PR #770, opened 2023-05-24
chore: update CDK bootstrap template

ashishdhingra:
*Issue #, if available:*

*Description of changes:*
chore: update CDK bootstrap template to version 18.
___

### DIFF
--- a/src/AWS.Deploy.Orchestration/CDK/CDKBootstrapTemplate.yaml
+++ b/src/AWS.Deploy.Orchestration/CDK/CDKBootstrapTemplate.yaml
@@ -204,8 +204,8 @@ Resources:
             Status: Enabled
             NoncurrentVersionExpiration:
               NoncurrentDays: 365
-    UpdateReplacePolicy: Delete
-    DeletionPolicy: Delete
+    UpdateReplacePolicy: Retain
+    DeletionPolicy: Retain
   StagingBucketPolicy:
     Type: AWS::S3::BucketPolicy
     Properties:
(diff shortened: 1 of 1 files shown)


96malhar (approved):

96malhar:
Some of the text is spilled across 2 lines but that's how it is in the source template as well - https://github.com/aws/aws-cdk/blob/main/packages/aws-cdk/lib/api/bootstrap/bootstrap-template.yaml

ashovlin:
> Some of the text is spilled across 2 lines but that's how it is in the source template as well - https://github.com/aws/aws-cdk/blob/main/packages/aws-cdk/lib/api/bootstrap/bootstrap-template.yaml

I'm guessing Phil typically updates these via a similar method to what the GitHub action does to check for a diff:
```
npm update -g aws-cdk
cdk bootstrap --show-template > newTemplate.yml
```

Could you use those contents instead? I think it's the same functionally, but will be an easier diff to review?

ashovlin (commented):
> Some of the text is spilled across 2 lines but that's how it is in the source template as well - https://github.com/aws/aws-cdk/blob/main/packages/aws-cdk/lib/api/bootstrap/bootstrap-template.yaml

I'm guessing Phil typically updates these via a similar method to what the GitHub action does to check for a diff:
```
npm update -g aws-cdk
cdk bootstrap --show-template > newTemplate.yml
```

Could you use those contents instead? I think it's the same functionally, but will be an easier diff to review since it matches existing whitespace/quotes/etc.

ashovlin (approved):

96malhar (approved):

PR #776, opened 2023-06-16
fix: incorrect message when deployment settings file doesn't exist

cresvi:
*Description of changes:*
While running some test scenarios in CodeCatalyst using different deployment settings files I kept the error message `An error occured while trying to deserialize the deployment settings file located at ...`. I was sure the file schema was correct, but was not sure what was the error. I cloned the repo and debugged the CLI. I identified that the same error message is shown to the user whether there was an error deserializing the deployment settings file, of if the file cannot be found.

This fix adds a specific error message telling users to check the path they provided exist to avoid this confusion. For scenarios when the file exists but there is a problem deserializing the JSON, the error message now also shows the specifics of what is wrong with the file. 

---

### DIFF
--- a/src/AWS.Deploy.Orchestration/DeploymentSettingsHandler.cs
+++ b/src/AWS.Deploy.Orchestration/DeploymentSettingsHandler.cs
@@ -60,11 +60,11 @@ namespace AWS.Deploy.Orchestration
             }
             catch (IOException ex)
             {
-                throw new InvalidDeploymentSettingsException(DeployToolErrorCode.DeploymentConfigurationNotFound, $"An error occured while trying to read the deployment settings file located at {filePath}. Make sure the file exists and is readable.", ex);
+                throw new InvalidDeploymentSettingsException(DeployToolErrorCode.DeploymentConfigurationNotFound, $"An error occurred while trying to read the deployment settings file located at {filePath}. Make sure the file exists and is readable.", ex);
             }
             catch (Exception ex)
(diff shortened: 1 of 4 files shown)


src/AWS.Deploy.Orchestration/DeploymentSettingsHandler.cs
```diff
@@ -58,9 +58,13 @@ public DeploymentSettingsHandler(IFileManager fileManager, IDirectoryManager dir
                 var userDeploymentSettings = JsonConvert.DeserializeObject<DeploymentSettings>(contents);
                 return userDeploymentSettings;
             }
+            catch (IOException ex)
+            {
+                throw new InvalidDeploymentSettingsException(DeployToolErrorCode.DeploymentConfigurationNotFound, $"An error occurred while trying to read the deployment settings file located at {filePath}. Make sure the path you provided exists on your file system, and is readable.", ex);
```
normj, line 63: Before the `try` block can you call `_fileManager.Exists` and if that returns back false you can give a specific error message that the file does not exist.

normj (changes requested):

src/AWS.Deploy.Orchestration/DeploymentSettingsHandler.cs
cresvi: Done, please review the new changes.

normj (approved):

PR #787, opened 2023-09-20
feat: add support for NodeJS 18 in generated dockerfiles

philasmar:
*Description of changes:*
* Upgrade NodeJS to v18 since v16 reached it's end of life and deployments were showing a deprecation notice.
* Upgraded project dependencies
* Adapted project to allow building it with .NET8 SDK
* Updated the way we were installing NodeJS on our dockerfiles since the script we were using has also been deprecated


src/AWS.Deploy.Orchestration/Data/AWSResourceQueryer.cs
```diff
@@ -512,7 +512,7 @@ public async Task<Vpc> GetDefaultVpc()
                             }
                         }
                     })
-                .Vpcs.FirstOrDefaultAsync(),
+                .Vpcs.FirstAsync(),
```
normj, line 515: What is the reason for this change? Won't this cause an error message to the user where we just add "Sequence contains no elements" to the end of there are is no default VPC?

normj (changes requested):

src/AWS.Deploy.Orchestration/Data/AWSResourceQueryer.cs
philasmar: Building the solution with .NET8 threw a "possible null reference" exception for that block of code. I switched back to `FirstOrDefaultAsync` and instead I am not returning a nullable VPC.

normj (approved):

site/content/docs/getting-started/pre-requisites.md
```diff
@@ -22,7 +22,7 @@ dotnet --version
 
 #### Node.js
 
-* The deployment tool requires the [AWS Cloud Development Kit (CDK)](https://docs.aws.amazon.com/cdk/latest/guide/), and the AWS CDK requires [Node.js](https://nodejs.org/en/download/). AWS CDK requires Node.js, versions 14.x, 16.x, 18.x (or later) - we recommend installing the latest LTS version.
+* The deployment tool requires the [AWS Cloud Development Kit (CDK)](https://docs.aws.amazon.com/cdk/latest/guide/), and the AWS CDK requires [Node.js](https://nodejs.org/en/download/). AWS CDK requires Node.js 18.x (or later) - we recommend installing the latest LTS version.
```
ashovlin, line 25: nit: Does CDK require 18 now? They still say 14 here: https://github.com/aws/aws-cdk#getting-started

Fine if _we_ require it, just don't want to mislead anybody.

src/AWS.Deploy.CLI/AWS.Deploy.CLI.csproj
```diff
@@ -47,6 +47,10 @@
     <None Include="..\..\README.md" Pack="true" PackagePath="" />
   </ItemGroup>
 
+  <ItemGroup>
+    <PackageReference Update="Nerdbank.GitVersioning" Version="3.6.133" />
```
ashovlin, line 51: Curious, why this addition here now?

src/AWS.Deploy.CLI/AWS.Deploy.CLI.csproj
```diff
@@ -21,15 +21,15 @@
...
 
   <ItemGroup>
-    <PackageReference Include="AWSSDK.IdentityManagement" Version="3.7.100.15" />
-    <PackageReference Include="AWSSDK.CloudFormation" Version="3.7.101.9" />
-    <PackageReference Include="AWSSDK.SecurityToken" Version="3.7.100.14" />
-    <PackageReference Include="AWSSDK.SSO" Version="3.7.100.14" />
-    <PackageReference Include="AWSSDK.SSOOIDC" Version="3.7.100.14" />
-    <PackageReference Include="Swashbuckle.AspNetCore.Annotations" Version="6.1.2" />
-    <PackageReference Include="Swashbuckle.AspNetCore.Swagger" Version="6.1.2" />
+    <PackageReference Include="AWSSDK.IdentityManagement" Version="3.7.200.42" />
```
ashovlin, line 24: Could you update https://github.com/aws/aws-dotnet-deploy/blob/main/THIRD_PARTY_LICENSES and the internal tracker as well for all the version bumps?

site/content/docs/getting-started/pre-requisites.md
philasmar: I updated the wording to not explicitly indicate that CDK is requiring a specific version, rather a specific version is required to use the tool.

src/AWS.Deploy.CLI/AWS.Deploy.CLI.csproj
philasmar: This was a miss. While using the Nuget Package Manager, it added it to the csproj instead of the `Directory.Build.props` file. Good catch.

src/AWS.Deploy.CLI/AWS.Deploy.CLI.csproj
philasmar: I went over all the dependencies and updated them. Also removed any dependency not referenced explicitly in the project.

ashovlin (approved):

ashovlin (approved):
Looks good after https://github.com/aws/aws-dotnet-deploy/compare/b667a50dc3b63566ef412e6aecc81637ad5d520f..8e839e30eb2a203b08f212be2a11baa49fc48426

normj (approved):

PR #797, opened 2023-10-17
update image uploader console app to work with latest changes

philasmar:
*Description of changes:*
* Update the console app to work with the changes introduced in https://github.com/aws/aws-dotnet-deploy/pull/793

### DIFF
--- a/test/AWS.Deploy.DockerImageUploader/App.cs
+++ b/test/AWS.Deploy.DockerImageUploader/App.cs
@@ -9,6 +9,10 @@ using System.Collections.Generic;
 using System;
 using System.Threading.Tasks;
 using System.IO;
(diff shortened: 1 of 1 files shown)


96malhar (approved):

ashovlin (approved):
Code looks fine. For my understanding, does the recommendation matter (and we're relying on whatever happens to be first today)? Or is any fine?

philasmar:
> Code looks fine. For my understanding, does the recommendation matter (and we're relying on whatever happens to be first today)? Or is any fine?

Recommendation doesn't matter. It is now required to be passed to the `GenerateDockerfile` function just so we can check the target framework. So any would be fine.

Issue #806, opened 2023-12-05
Add support for deploying .NET 6 applications to Amazon Linux 2023 Beanstalk platforms


shruti0085:
### Describe the feature

Creating an issue here to track https://github.com/aws/aws-toolkit-visual-studio/issues/371#issuecomment-1819465803 on the Visual Studio Toolkit.
The Visual Studio Toolkit wraps the .NET Deployment Tool, so any changes to behavior will have to originate in this repository.

Deploy tool does not support deploying to Beanstalk platforms of Amazon Linux 2023 today. The latest platform version as of(12/4/2023) for .NET 6 project is `.NET Core Running on 64 bit Amazon Linux 2 V2.6.2`

### Use Case

Snippet from the feature-request
I encountered the same behavior and compared a working case to a non-working case. I noted that the non working case was using Linux 2023 as opposed to Linux 2... Linux 2023 environment is new default behavior, via the AWS create environment wizard. I tried creating a new environment using ".NET Core running on 64bit Amazon Linux 2/2.6.2" and that worked. I also tried creating an environment via the AWS toolkit and noted that the created environment was Linux 2, supporting my suspicion that this version of the AWS toolkit is not supporting Linux 2023.

### Proposed Solution

Add support for deploying to Beanstalk platforms for Amazon Linux 2023 with the deploy tool

### Acknowledgements

- [ ] I may be able to implement this feature request
- [ ] This feature might incur a breaking change

### Version used

1.16.7

### Operating System and version

Windows 10

ashishdhingra:
Needs review with the team for prioritization.

ashovlin:
This has been addressed in https://github.com/aws/aws-dotnet-deploy/releases/tag/1.18.6.

PR #807, opened 2023-12-05
fix: Enable selecting the '.NET 6 on AL2023' platform when deploying to Elastic Beanstalk on Linux

ashovlin:
*Issue #, if available:* #806 (I had tested this quickly when talking to @shruti0085 internally about this issue)

*Description of changes:*

This widens the string we use to filter Beanstalk Platform options when deploying to Linux, to include the new ".NET 6 on AL2023" platform that was [released in October](https://docs.aws.amazon.com/elasticbeanstalk/latest/relnotes/release-2023-10-19-al2023.html).

image

**Considerations**:
* The sort order we're getting back from the service call is what I think we want, so didn't introduce custom sorting like we have for Windows here: https://github.com/aws/aws-dotnet-deploy/blob/03abb117585784ad0fec360e7c5093e37ddf7993/src/AWS.Deploy.Orchestration/Data/AWSResourceQueryer.cs#L627
* Per the [docs](https://docs.aws.amazon.com/elasticbeanstalk/latest/platforms/platform-history-dotnetlinux.html), the new platform only has .NET 6 installed. However our Beanstalk recipes still allow `netcoreapp3.1` or greater:
https://github.com/aws/aws-dotnet-deploy/blob/03abb117585784ad0fec360e7c5093e37ddf7993/src/AWS.Deploy.Recipes/RecipeDefinitions/ASP.NETAppElasticBeanstalkLinux.recipe#L36
Deploying an application with an older target (in this case .NET 5) to the new AL2023 platform results in:
```
Dec  5 22:10:04 ip-172-31-83-112 web[2147]: You must install or update .NET to run this application.
Dec  5 22:10:04 ip-172-31-83-112 web[2147]: App: /var/app/current/ASPNet5.dll
Dec  5 22:10:04 ip-172-31-83-112 web[2147]: Architecture: x64
Dec  5 22:10:04 ip-172-31-83-112 web[2147]: Framework: 'Microsoft.AspNetCore.App', version '5.0.0' (x64)
```

However **this is true even for the recent AL2 platform versions as well**. Unsure if we want to unblock AL2023 initially with this, then better handle the compatibility between the application being deploying and the specific selected platform later? Or try to tackle both issues together

### DIFF
--- a/src/AWS.Deploy.Constants/ElasticBeanstalk.cs
+++ b/src/AWS.Deploy.Constants/ElasticBeanstalk.cs
@@ -22,7 +22,7 @@ namespace AWS.Deploy.Constants
         public const string HealthCheckURLOptionNameSpace = "aws:elasticbeanstalk:application";
         public const string HealthCheckURLOptionName = "Application Healthcheck URL";
 
-        public const string LinuxPlatformType = ".NET Core";
+        public const string LinuxPlatformType = ".NET";
         public const string WindowsPlatformType = "Windows Server";
 
         public const string IISAppPathOptionId = "IISAppPath";
--- a/src/AWS.Deploy.Constants/ElasticBeanstalk.cs
+++ b/src/AWS.Deploy.Constants/ElasticBeanstalk.cs
@@ -22,7 +22,7 @@ namespace AWS.Deploy.Constants
         public const string HealthCheckURLOptionNameSpace = "aws:elasticbeanstalk:application";
         public const string HealthCheckURLOptionName = "Application Healthcheck URL";
 
-        public const string LinuxPlatformType = ".NET";
+        public const string LinuxPlatformType = ".NET ";
         public const string WindowsPlatformType = "Windows Server";
 
         public const string IISAppPathOptionId = "IISAppPath";


ashovlin:
> Deploying an application with an older target (in this case .NET 5) to the new AL2023 platform results in:
> ```
> Dec  5 22:10:04 ip-172-31-83-112 web[2147]: You must install or update .NET to run this application.
> Dec  5 22:10:04 ip-172-31-83-112 web[2147]: App: /var/app/current/ASPNet5.dll
> Dec  5 22:10:04 ip-172-31-83-112 web[2147]: Architecture: x64
> Dec  5 22:10:04 ip-172-31-83-112 web[2147]: Framework: 'Microsoft.AspNetCore.App', version '5.0.0' (x64)
> ```
> However this is true even for the recent AL2 platform versions as well. Unsure if we want to unblock AL2023 initially with this, then better handle the compatibility between the application being deploying and the specific selected platform later? Or try to tackle both issues together

We discussed this in today's GitHub triage meeting with @normj , we're comfortable NOT closing this gap at this point. We expect users to be moving off of .NET Core 3.1 and .NET 5 anyway.

src/AWS.Deploy.Constants/ElasticBeanstalk.cs
```diff
@@ -22,7 +22,7 @@ internal static class ElasticBeanstalk
         public const string HealthCheckURLOptionNameSpace = "aws:elasticbeanstalk:application";
         public const string HealthCheckURLOptionName = "Application Healthcheck URL";
 
-        public const string LinuxPlatformType = ".NET Core";
+        public const string LinuxPlatformType = ".NET";
```
normj, line 25: Can we add a trailing space, `".NET "`,  to help avoid getting false positives like `.NETWORK`. Unlikely to happen but figure it doesn't hurt to add.

normj (changes requested):

src/AWS.Deploy.Constants/ElasticBeanstalk.cs
ashovlin: Done via https://github.com/aws/aws-dotnet-deploy/pull/807/commits/7ad351b8373436f48d05971784a763e67ac7a504

philasmar (approved):

PR #808, opened 2023-12-20
Adjust role used by GitHub action that builds docs.

ashovlin:
*Issue #, if available:* N/A

*Description of changes:*
1. Fixes a CI failure on #807. We recently adjusted some internal accounts where we run integration tests, so `CI_AWS_ROLE_ARN` is no longer valid. This reuses the `CI_MAIN_TESTING_ACCOUNT_ROLE_ARN` instead.
    * The doc generator uses server mode, which requires AWS credentials to pass its health check. I didn't actually see it make any requests when running locally.
3. Also upgraded the credential action while here, there was a warning about old Node versions.

### DIFF
--- a/.github/workflows/DetectDocGeneratorChanges.yml
+++ b/.github/workflows/DetectDocGeneratorChanges.yml
@@ -12,9 +12,9 @@ jobs:
 
     steps:
     - name: Configure AWS Credentials
-      uses: aws-actions/configure-aws-credentials@v1
+      uses: aws-actions/configure-aws-credentials@8c3f20df09ac63af7b3ae3d7c91f105f857d8497 #v4
       with:
-        role-to-assume: ${{ secrets.CI_AWS_ROLE_ARN }}
+        role-to-assume: ${{ secrets.CI_MAIN_TESTING_ACCOUNT_ROLE_ARN }}
         role-duration-seconds: 7200
         aws-region: us-west-2
     - uses: actions/checkout@v3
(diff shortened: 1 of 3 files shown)


.github/workflows/doc-builder.yml
```diff
@@ -16,9 +16,9 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - name: Configure AWS Credentials
-        uses: aws-actions/configure-aws-credentials@v1
+        uses: aws-actions/configure-aws-credentials@8c3f20df09ac63af7b3ae3d7c91f105f857d8497 #v4
         with:
-          role-to-assume: ${{ secrets.CI_AWS_ROLE_ARN }}
```
ashovlin, line 21: I think we tore this down while performance maintenance on the account it maps to. Let me know if you don't think its appropriate to reuse CI_MAIN_TESTING_ACCOUNT_ROLE_ARN for this use case.

96malhar (approved):

.github/workflows/DetectDocGeneratorChanges.yml
```diff
@@ -12,9 +12,9 @@ jobs:
 
     steps:
     - name: Configure AWS Credentials
-      uses: aws-actions/configure-aws-credentials@v1
+      uses: aws-actions/configure-aws-credentials@8c3f20df09ac63af7b3ae3d7c91f105f857d8497 #v4
       with:
-        role-to-assume: ${{ secrets.CI_AWS_ROLE_ARN }}
+        role-to-assume: ${{ secrets.CI_MAIN_TESTING_ACCOUNT_ROLE_ARN }}
```
philasmar, line 17: i think ideally we should use this account to determine which test runner to use, but i'm fine with this for now.

philasmar (approved):

PR #815, opened 2024-03-05
Fix escaping for Lambda invocation that selects which account to run integration tests in

ashovlin:
*Issue #, if available:* N/A

*Description of changes:* Same as https://github.com/awslabs/aws-dotnet-messaging/pull/94, I'm still not sure what changed that led to this but fixing it up.

### DIFF
--- a/.github/workflows/codebuild-ci.yml
+++ b/.github/workflows/codebuild-ci.yml
@@ -35,7 +35,7 @@ jobs:
         id: lambda
         shell: pwsh
         run: |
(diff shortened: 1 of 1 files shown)


philasmar (approved):

ashovlin:
Now the tests are running, but the CI is failing for a different reason. Looks like the version of CDK the Blazor recipe is using is attempting to deploy a Node 14 Lambda which is no longer supported. We'll look at upgrading the CDK dependency in another PR.

PR #818, opened 2024-03-08
chore: Update the version of Amazon.CDK.Lib that is referenced by the recipe CDK projects

ashovlin:
*Issue #, if available:* N/A

*Description of changes:* 

Our integration tests began failing for the Blazor recipe:
* Example CI failure: https://github.com/aws/aws-dotnet-deploy/actions/runs/8193118953/job/22406129833
  > 2024-03-07T18:42:09.0042659Z BlazorWasm60c01024a63d82 |  6/14 | 6:37:45 PM | CREATE_FAILED        | AWS::Lambda::Function                           | Custom::S3AutoDeleteObjectsCustomResourceProvider/Handler (CustomS3AutoDeleteObjectsCustomResourceProviderHandler9D90184F) Resource handler returned message: "The runtime parameter of nodejs14.x is no longer supported for creating or updating AWS Lambda functions. We recommend you use the new runtime (nodejs20.x) while creating or updating functions. (Service: Lambda, Status Code: 400, Request ID: 3674fb8b-b71f-4624-a88d-3df206a02179)" (RequestToken: f152ff23-1334-fc62-7d71-aba03b7982cb, HandlerErrorCode: InvalidRequest)

By setting `AutoDeleteObjects` on the S3 bucket in the recipe to true, a CDK-defined Node 14 Lambda is deployed (which is [no longer supported as of 1/19/24](https://docs.aws.amazon.com/lambda/latest/dg/lambda-runtimes.html#runtimes-deprecated)):
https://github.com/aws/aws-dotnet-deploy/blob/5505e4042a4759c762deef9683824b88bec3ea28/src/AWS.Deploy.Recipes/CdkTemplates/BlazorWasm/Generated/Recipe.cs#L57

This updates the `Amazon.CDK.Lib` for all of the recipes to the latest. I believe they went to Node 18 for custom resources in 2.87 via https://github.com/aws/aws-cdk/pull/26212.
* [WIP] - I haven't done any manual testing yet. Not sure if we might have any issues where a project deployed via 2.43.1 might fail to redeploy on 2.131.0, which we don't have covered by the current automated tests.

### DIFF
--- a/src/AWS.Deploy.Recipes/CdkTemplates/BlazorWasm/Generated/Recipe.cs
+++ b/src/AWS.Deploy.Recipes/CdkTemplates/BlazorWasm/Generated/Recipe.cs
@@ -132,6 +132,7 @@ namespace BlazorWasm
                     var loggingBucket = new Bucket(this, nameof(AccessLoggingBucket), InvokeCustomizeCDKPropsEvent(nameof(AccessLoggingBucket), this, new BucketProps
                     {
                         RemovalPolicy = RemovalPolicy.RETAIN,
+                        AccessControl = BucketAccessControl.LOG_DELIVERY_WRITE
                     }));
 
                     distributionProps.LogBucket = loggingBucket;
--- a/THIRD_PARTY_LICENSES
+++ b/THIRD_PARTY_LICENSES
@@ -21,7 +21,7 @@
 ** AWSSDK.ElasticLoadBalancingV2; version 3.7.201.27 -- https://www.nuget.org/packages/AWSSDK.ElasticLoadBalancingV2/
 ** AWSSDK.Core; version 3.7.202.11 -- https://www.nuget.org/packages/AWSSDK.Core
 ** AWSSDK.CloudWatchLogs; version 3.7.200.42 -- https://www.nuget.org/packages/AWSSDK.CloudWatchLogs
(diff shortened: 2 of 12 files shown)


philasmar (approved):

src/AWS.Deploy.Recipes/CdkTemplates/BlazorWasm/Generated/Recipe.cs
```diff
@@ -132,6 +132,7 @@ private void ConfigureCloudFrontDistribution(Configuration settings)
                     var loggingBucket = new Bucket(this, nameof(AccessLoggingBucket), InvokeCustomizeCDKPropsEvent(nameof(AccessLoggingBucket), this, new BucketProps
                     {
                         RemovalPolicy = RemovalPolicy.RETAIN,
+                        AccessControl = BucketAccessControl.LOG_DELIVERY_WRITE
```
ashovlin, line 135: Unrelated to the CDK version, noticed that this is failing as well due to a change in S3 behavior last year. See https://github.com/aws/aws-cdk/issues/27571 for more info.

ashovlin:
I've manually tested that deploying an application using 1.18.6 (currently live), and then redeploying with this branch works for the Fargate and App Runner recipes. I'm struggling to redeploy a Beanstalk application with 1.18.6 both before and after, but don't think it'd be related to this change.

normj (approved):

PR #819, opened 2024-03-08
Adjust Security Group type hints to show the group name and id

awschristou:
Security Group IDs don't provide enough context on their own for users to choose the right ones from their account. We received Toolkit feedback that users ended up logging into the web console in order to verify they are selecting the correct security group(s) when deploying their project.

This change updates the display value for security groups, so that it shows the group name and id.

Here is an example using this change:
image


ashovlin (approved):
CI failure looks the same as what is being address via #818

PR #823, opened 2024-04-19
test: Add project for executing canary tests using the published tool

ashovlin:
*Issue #, if available:* DOTNET-7485

*Description of changes:* This is the test project that will be used to power an internal canary that will attempt deployments using the _published version of the tool_ with the test apps we use for integration tests.
* It uses the `/testapps` and their settings files where possible, but I did redefine some settings files to handle the docker execution directory correctly.
* It initially covers the initial deployment for each recipe. We may expand it in the future to cover redeployments and/or more advanced configuration.

### DIFF
--- a/buildtools/smoketests.proj
+++ b/buildtools/smoketests.proj
@@ -1,0 +1,90 @@
+<Project ToolsVersion="Current" DefaultTargets="smoke-tests" xmlns="http://schemas.microsoft.com/developer/msbuild/2003"> 
+    <PropertyGroup>
+        <RootPath Condition="'$(RootPath)'==''">$(MSBuildProjectDirectory)</RootPath>
+
+        <!-- We need to skip recipes that use Docker when running on Windows CodeBuild -->
+        <ShouldRunDockerTests>true</ShouldRunDockerTests>
+        <ShouldRunDockerTests Condition="'$(OS)' == 'Windows_NT' AND '$(CODEBUILD_BUILD_ID)' != ''">false</ShouldRunDockerTests>
+
+        <!-- Generate a suffix for CloudFormation stack names, in case multiple smoke tests are running at once -->
+        <SafeOperatingSystem>$(OS)</SafeOperatingSystem>
+        <SafeOperatingSystem Condition="'$(OS)' == 'Windows_NT'">Windows</SafeOperatingSystem> <!-- need to remove _ for CloudFormation-->
+        <Suffix>$(SafeOperatingSystem)-$([System.DateTimeOffset]::UtcNow.ToUnixTimeSeconds())</Suffix>
+
+        <BlazorAppName>CanaryBlazorApp-$(Suffix)</BlazorAppName>
+        <ECSFargateAppName>CanaryECSFargateApp-$(Suffix)</ECSFargateAppName>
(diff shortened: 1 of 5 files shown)


buildtools/smoketests.proj
```diff
@@ -0,0 +1,90 @@
+<Project ToolsVersion="Current" DefaultTargets="smoke-tests" xmlns="http://schemas.microsoft.com/developer/msbuild/2003"> 
+    <PropertyGroup>
+        <RootPath Condition="'$(RootPath)'==''">$(MSBuildProjectDirectory)</RootPath>
+
+        <!-- We need to skip recipes that use Docker when running on Windows CodeBuild -->
+        <ShouldRunDockerTests>true</ShouldRunDockerTests>
```
philasmar, line 6: nit: could you specify an inverse condition here? Its easy to mess this up if in the future someone mistakenly reorders them.

buildtools/smoketests.proj
```diff
@@ -0,0 +1,90 @@
...
+        <ShouldRunDockerTests Condition="'$(OS)' == 'Windows_NT' AND '$(CODEBUILD_BUILD_ID)' != ''">false</ShouldRunDockerTests>
+
+        <!-- Generate a suffix for CloudFormation stack names, in case multiple smoke tests are running at once -->
+        <SafeOperatingSystem>$(OS)</SafeOperatingSystem>
```
philasmar, line 10: nit: could you specify an inverse condition here? Its easy to mess this up if in the future someone mistakenly reorders them.

philasmar: I'm worried about multiple runs of the canary, failed deployments and failing to delete some stacks. It is possible to get to a state where we have stacks that linger to the next canary run (although unlikely). Some of our tests reuse the same names and IDs. I would feel better if we can use a more unique suffix like a guid instead of the OS name. Though I don't want us to over complicate things. I'm fine moving forward and seeing what the canaries will be like in action.

philasmar (approved):
We should add to our runbook an entry to remind us to add more canary tests when adding new recipes.

buildtools/smoketests.proj
ashovlin: The full suffix includes `$([System.DateTimeOffset]::UtcNow.ToUnixTimeSeconds()` below down on L12.

So each run of the canary will use unique stack names based on the timestamp. If the Windows and Linux runners start at the same second, then the `SafeOperatingSystem` will disambiguate them.

buildtools/smoketests.proj
ashovlin: Flipped in https://github.com/aws/aws-dotnet-deploy/pull/823/commits/c10ee39db77ccf1479dbff8a9583bf4536931195. 

(had to refresh myself on https://en.wikipedia.org/wiki/De_Morgan%27s_laws :-D )

buildtools/smoketests.proj
philasmar: I remember deleting that comment. Dunno why it posted. I deleted it when I read a few lines later where you were doing `$([System.DateTimeOffset]::UtcNow.ToUnixTimeSeconds()`

normj (approved):

PR #830, opened 2024-05-23
Cache AWS API read service calls in ServerMode for a session 

normj:
*Description of changes:*
To improve performance editing the deployment configuration mode this PR adds a cache for the AWS API read service calls made. Currently all the AWS API calls are being called every time any setting is changed including editing the name of a deployment. Depending on user's connection this can cause a significant user experience delay.

The cache is per deployment session. If a user does create resources after the deployment session has started they will not show up till the user restarts the deployment. This is a trade off to avoid constantly checking for new resources providing a poor user experience.

The cache is implemented in the new `SessionAWSResourceQuery` which implements the `IAWSResourceQuery` interface. It wraps around the standard `AWSResourceQuery` implementation but caches the results in its internal cache for subsequent calls. In the `DeploymentController` which sets the the DI container for each incoming request it was changed to inject `SessionAWSResourceQuery` instead of `AWSResourceQuery`. The instance of `SessionAWSResourceQuery` is stored on the `SessionState` so subsequent calls for a session inject the instance from session state into the DI container.


src/AWS.Deploy.CLI/Commands/DeployCommand.cs
```diff
@@ -586,7 +586,7 @@ private async Task CreateDeploymentBundle(Orchestrator orchestrator, Recommendat
             {
                 if (_toolInteractiveService.DisableInteractive)
                 {
-                    throw ex;
+                    throw;
```
normj, line 589: Not sure what changed in my dev environment by these `throw ex` started giving me compiler errors due to warnings be treated as errors.

src/AWS.Deploy.CLI/ServerMode/Services/SessionAWSResourceQuery.cs
```diff
@@ -0,0 +1,378 @@
...
+                    {
+                        values.Add("IEnumerable");
+                        foreach (var subItem in itemEnumerable)
+                        {
+                            if (subItem is null)
+                                values.Add("null");
+                            else
+                                values.Add(subItem.ToString() ?? "null");
+                        }
+                    }
```
ashovlin, line 99: Checking my understanding - for something like `GetS3BucketLocation("my-bucket")`, should this be `"GetS3BucketLocation,mybucket"`? I'm seeing `"GetS3BucketLocation,IEnumerable,m,y,-,b,u,c,k,e,t"`.

Should there be an `item.ToString()` for the case where `item` is not `null` && not an `IEnumerable`?

ashovlin: Also, wonder if we should be throwing an exception for if we fail to add something to `values` for a given `item`. 

I think it'd be rare for us to add new type of signature to the interface, but if we do I think it'd be tough to remember to make sure it works here.

src/AWS.Deploy.CLI/ServerMode/Controllers/DeploymentController.cs
```diff
@@ -74,17 +74,18 @@ public async Task<IActionResult> StartDeploymentSession(StartDeploymentSessionIn
...
                 input.ProjectPath,
                 input.AWSRegion,
-                (await awsResourceQueryer.GetCallerIdentity(input.AWSRegion)).Account,
                 await _projectParserUtility.Parse(input.ProjectPath)
                 );
 
+            var serviceProvider = CreateSessionServiceProvider(state);
+            var awsResourceQueryer = serviceProvider.GetRequiredService<IAWSResourceQueryer>();
+
+            state.AWSAccountId = (await awsResourceQueryer.GetCallerIdentity(input.AWSRegion)).Account;
```
ashovlin, line 87: Just curious, why move out of the constructor?

normj: Before the `serviceProvider` could be created before the session state but now the session state has to be created first in order to create the `serviceProvider` which would allow injecting the `SessionAWSResourceQueryer` into the service collection. 

Since we don't have an `IAWSRequesourceQueryer` used to get the account id at the time of creating the session state I had to move the `AWSAccountId` to being set after the creation of both the session state and the `serviceProvider`.

ashovlin (approved):
Approving, but could you squash or include at least one conventional commit message when merging to update the release notes correctly?

normj:
> Approving, but could you squash or include at least one conventional commit message when merging to update the release notes correctly?

Forgot we were using semantic commit messages for this repo but yes I will squash the commits when merges.

96malhar (approved):

normj:
Merged as part of PR https://github.com/aws/aws-dotnet-deploy/pull/833

PR #832, opened 2024-06-03
Fix: Display correct versioning info

96malhar:
*Description of changes:*
Currently, the output from the `--version` switch is suffixed with the git commit hash which is not desirable.
image.

This PR fixes the output of the `--version` switch so that it reflects the actual assembly version without the commit hash
image


normj (approved):

philasmar (changes requested):
Changes look good. However, you need to add a unit test to improve the code coverage of the new code path.

96malhar:
> Changes look good. However, you need to add a unit test to improve the code coverage of the new code path.

Added new integration test to exercise the new logic.

philasmar (approved):

PR #837, opened 2024-06-20
Rework System Capabilities Check (Caching, Timeout, and Conditional on Recipe)

ashovlin:
*Issue #, if available:* DOTNET-7552

*Description of changes:* The deploy tool checks whether the user has Node.JS and Docker installed, since those are required for most deployments.

1. Applied a timeout of one minute to these checks. We've seen cases where Docker doesn't wake up from resource saver mode, then the check hangs indefinitely.
2. We now check selectively for each based on the selected, rather than always checking for both. For example, Elastic Beanstalk doesn't require Docker, and pushing an image to ECR doesn't require Node/CDK.
3. We now cache successful results for an hour. We don't cache unsuccessful results, since a user may have taken an action (such as starting Docker or switching it to Linux mode) after a failed check.

I also consolidated two different helpers for unit testing `CommandLineWrapper` into one.


src/AWS.Deploy.Orchestration/SystemCapabilityEvaluator.cs
```diff
@@ -27,105 +28,161 @@ public class SystemCapabilityEvaluator : ISystemCapabilityEvaluator
...
+        private const int CAPABILITY_EVALUATION_TIMEOUT_MS = 60000; // one minute
 
-        public async Task<SystemCapabilities> Evaluate()
-        {
-            var dockerTask = HasDockerInstalledAndRunning();
-            var nodeTask = GetNodeJsVersion();
+        /// <summary>
+        /// How long to cache the results of a VALID Node/Docker/etc. check 
+        /// </summary>
+        private static readonly TimeSpan DEPENDENCY_CACHE_INTERVAL = TimeSpan.FromHours(1);
```
ashovlin, line 39: Open to feedback on the lengths of time.

src/AWS.Deploy.CLI/Utilities/CommandLineWrapper.cs
```diff
@@ -45,8 +45,8 @@ public class CommandLineWrapper : ICommandLineWrapper
             bool redirectIO = true,
             string? stdin = null,
             IDictionary<string, string>? environmentVariables = null,
-            CancellationToken cancelToken = default,
-            bool needAwsCredentials = false)
+            bool needAwsCredentials = false,
+            CancellationToken cancellationToken = default)
```
ashovlin, line 49: Flipped these everywhere for https://learn.microsoft.com/en-us/dotnet/fundamentals/code-analysis/quality-rules/ca1068

src/AWS.Deploy.Orchestration/SystemCapabilityEvaluator.cs
```diff
@@ -27,105 +28,161 @@ public class SystemCapabilityEvaluator : ISystemCapabilityEvaluator
...
-            var systemCapabilities = await Evaluate();
+            var missingCapabilitiesForRecipe = new List<SystemCapability>();
             string? message;
+
+            // We only need to check that Node is installed if the user is deploying a recipe that uses CDK
             if (selectedRecommendation.Recipe.DeploymentType == DeploymentTypes.CdkProject)
             {
-                if (systemCapabilities.NodeJsVersion == null)
+                // If we haven't cached that NodeJS installation is valid, or the cache is expired
+                if (_nodeDependencyValidUntilUtc == null || DateTime.UtcNow >= _nodeDependencyValidUntilUtc)
```
96malhar, line 138: nit: Consider setting the default value of `_nodeDependencyValidUntilUtc` to [DateTime.MinValue](https://learn.microsoft.com/en-us/dotnet/api/system.datetime.minvalue?view=net-8.0) instead on null.

src/AWS.Deploy.Orchestration/SystemCapabilityEvaluator.cs
```diff
@@ -27,105 +28,161 @@ public class SystemCapabilityEvaluator : ISystemCapabilityEvaluator
...
+            // We only need to check that Docker is installed if the user is deploying a recipe that uses Docker
             if (selectedRecommendation.Recipe.DeploymentBundle == DeploymentBundleTypes.Container)
             {
-                if (!systemCapabilities.DockerInfo.DockerInstalled)
-                {
-                    message = "Install and start Docker version appropriate for your OS. This deployment option requires Docker, which was not detected.";
-                    capabilities.Add(new SystemCapability(DOCKER_DEPENDENCY_NAME, message, DOCKER_INSTALLATION_URL));
-                }
-                else if (!systemCapabilities.DockerInfo.DockerContainerType.Equals("linux", StringComparison.OrdinalIgnoreCase))
+                if (_dockerDependencyValidUntilUtc == null || DateTime.UtcNow >= _dockerDependencyValidUntilUtc)
```
96malhar, line 164: nit: Consider setting the default value of `_dockerDependencyValidUntilUtc` to [DateTime.MinValue](https://learn.microsoft.com/en-us/dotnet/api/system.datetime.minvalue?view=net-8.0) instead on null.

96malhar (approved):

src/AWS.Deploy.Orchestration/SystemCapabilityEvaluator.cs
philasmar: I'm not a fan of us having a fixed time. I prefer if we clear the cache once a user starts a deployment, or after a fixed time if they dont. If a customer messes anything up with their node or docker installations after a successful check if performed, they'd have to wait an hour for us to recheck or they'd have to kill the deploy tool instance which i think would cause confusion.

src/AWS.Deploy.Orchestration/SystemCapabilityEvaluator.cs
```diff
@@ -27,105 +28,161 @@ public class SystemCapabilityEvaluator : ISystemCapabilityEvaluator
...
+            catch (TaskCanceledException)
+            {
+                // If the check timed out, treat Docker as not installed
+                return new DockerInfo(false, "");
+            }
         }
 
         /// <summary>
         /// From https://docs.aws.amazon.com/cdk/latest/guide/work-with.html#work-with-prerequisites,
         /// min version is 10.3
```
philasmar, line 96: could you please update this comment to either mention the latest version we support (since we've updated it) or remove this version from the comment.

philasmar (commented):
Please add tests that go over the non-covered areas to address the CodeCov warnings

src/AWS.Deploy.Orchestration/SystemCapabilityEvaluator.cs
ashovlin: From discussion, look at if we should tie the SystemCapabilityEvaluator lifetime to the session in server mode.

src/AWS.Deploy.Orchestration/SystemCapabilityEvaluator.cs
ashovlin: Removed it in https://github.com/aws/aws-dotnet-deploy/pull/837/commits/b22ed89708bf1b4e05ec4b89d774f841be431331. This method is just returning the version anyway, the minimum check is done elsewhere.

src/AWS.Deploy.Orchestration/SystemCapabilityEvaluator.cs
ashovlin: Done in https://github.com/aws/aws-dotnet-deploy/pull/837/commits/b22ed89708bf1b4e05ec4b89d774f841be431331

src/AWS.Deploy.Orchestration/SystemCapabilityEvaluator.cs
ashovlin: Done in https://github.com/aws/aws-dotnet-deploy/pull/837/commits/b22ed89708bf1b4e05ec4b89d774f841be431331

src/AWS.Deploy.Orchestration/SystemCapabilityEvaluator.cs
ashovlin: In https://github.com/aws/aws-dotnet-deploy/pull/837/commits/b22ed89708bf1b4e05ec4b89d774f841be431331 I now clear the cache when the deployment starts from either CLI or server mode.

ashovlin:
> Please add tests that go over the non-covered areas to address the CodeCov warnings

Fixed now. For the one outstanding warning, that's where' we're reporting that Docker is in "windows" mode. I added a test for this, but it returns early if not running on Windows because we assume "linux" otherwise.
* https://github.com/aws/aws-dotnet-deploy/pull/837/checks?check_run_id=26821102918

philasmar (commented):
I'm good with the changes, but Codecov still shows a warning. Could you please tackle that?

philasmar (approved):

ashovlin:
> I'm good with the changes, but Codecov still shows a warning. Could you please tackle that?

Wrote a summary up in https://github.com/aws/aws-dotnet-deploy/pull/837#issuecomment-2197500039, essentially those lines will only ever be invoked on Windows (it's when Docker is running in Windows mode, which isn't possible on Linux/OSX), so the test that I did add returns early. I don't think it's worth mocking out the OS detection just for this coverage.

----

Noticed another issue while doing final testing in Visual Studio, I didn't realize that `SystemCapabilityEvaluator` was being recreated per server mode API call, which defeats the point of the caching. Addressed in https://github.com/aws/aws-dotnet-deploy/pull/837/commits/3723451b970e68bda2ec9763356fca57938ceda0

philasmar (approved):

96malhar (approved):

PR #839, opened 2024-07-02
feat: Remove disabled fields from showing on redeployment in server mode

philasmar:
*Issue #, if available:*
DOTNET-6316

*Description of changes:*
* Added a 'VisibileOnRedeployment' boolean on the option settings in the Recipe which determines the visibility of a setting during a redeployment when 'Updatable' is false
* The setting is false by default, meaning any setting with 'Updatable' = false will not be visible during redeployment.
* I have set the fields that have important names as visible during redeployment
* I have tested this behavior with the toolkit and CLI

### DIFF
--- a/src/AWS.Deploy.CLI/ServerMode/Controllers/DeploymentController.cs
+++ b/src/AWS.Deploy.CLI/ServerMode/Controllers/DeploymentController.cs
@@ -208,7 +208,9 @@ namespace AWS.Deploy.CLI.ServerMode.Controllers
                     Value = optionSettingHandler.GetOptionSettingValue(recommendation, setting),
                     Advanced = setting.AdvancedSetting,
                     ReadOnly = recommendation.IsExistingCloudApplication && !setting.Updatable,
-                    Visible = optionSettingHandler.IsOptionSettingDisplayable(recommendation, setting),
+                    Visible =
+                        optionSettingHandler.IsOptionSettingDisplayable(recommendation, setting) &&
+                        !(recommendation.IsExistingCloudApplication && !setting.Updatable && !setting.VisibleOnRedeployment),
                     SummaryDisplayable = optionSettingHandler.IsSummaryDisplayable(recommendation, setting),
                     AllowedValues = setting.AllowedValues,
                     ValueMapping = setting.ValueMapping,
--- a/src/AWS.Deploy.CLI/ServerMode/Controllers/DeploymentController.cs
+++ b/src/AWS.Deploy.CLI/ServerMode/Controllers/DeploymentController.cs
@@ -200,7 +200,12 @@ namespace AWS.Deploy.CLI.ServerMode.Controllers
 
             foreach (var setting in configurableOptionSettings)
             {
-                var settingSummary = new OptionSettingItemSummary(setting.Id, setting.FullyQualifiedId, setting.Name, setting.Description, setting.Type.ToString())
+                var settingSummary = new OptionSettingItemSummary(
+                    setting.Id,
+                    setting.FullyQualifiedId,
+                    setting.Name,
+                    setting.Description,
+                    setting.Type.ToString())
                 {
                     Category = setting.Category,
                     TypeHint = setting.TypeHint?.ToString(),
@@ -210,6 +215,8 @@ namespace AWS.Deploy.CLI.ServerMode.Controllers
                     ReadOnly = recommendation.IsExistingCloudApplication && !setting.Updatable,
                     Visible =
                         optionSettingHandler.IsOptionSettingDisplayable(recommendation, setting) &&
+                        // Updating visibility of settings in server-mode to be determined by 'VisibleOnRedeployment'
+                        // when performing a redeployment and 'Updatable' is set to false.
                         !(recommendation.IsExistingCloudApplication && !setting.Updatable && !setting.VisibleOnRedeployment),
                     SummaryDisplayable = optionSettingHandler.IsSummaryDisplayable(recommendation, setting),
                     AllowedValues = setting.AllowedValues,
--- a/src/AWS.Deploy.Recipes/RecipeDefinitions/ASP.NETAppAppRunner.recipe
+++ b/src/AWS.Deploy.Recipes/RecipeDefinitions/ASP.NETAppAppRunner.recipe
@@ -289,6 +289,7 @@
             "Type": "String",
             "AdvancedSetting": true,
             "Updatable": false,
+            "VisibleOnRedeployment": true,
             "Validators": [
                 {
                     "ValidatorType": "Regex",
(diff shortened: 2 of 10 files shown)


src/AWS.Deploy.CLI/ServerMode/Controllers/DeploymentController.cs
```diff
@@ -208,7 +208,9 @@ private List<OptionSettingItemSummary> ListOptionSettingSummary(IOptionSettingHa
                     Value = optionSettingHandler.GetOptionSettingValue(recommendation, setting),
                     Advanced = setting.AdvancedSetting,
                     ReadOnly = recommendation.IsExistingCloudApplication && !setting.Updatable,
-                    Visible = optionSettingHandler.IsOptionSettingDisplayable(recommendation, setting),
+                    Visible =
+                        optionSettingHandler.IsOptionSettingDisplayable(recommendation, setting) &&
+                        !(recommendation.IsExistingCloudApplication && !setting.Updatable && !setting.VisibleOnRedeployment),
```
awschristou, line 213: - can test coverage be applied to this new logic 
- "visible on redeployment" sounds definitive, but this isn't the case. It appears that "Updatable" takes priority (which makes sense). How can future maintainers avoid assumptions when they set VisibleOn to true and expect their field to always appear?

philasmar: I added a server mode test to mimic the behavior in the toolkit. This change only applies to server mode.
The JSON schema for the recipe explains this. I did not want to make the setting name very long.

src/AWS.Deploy.Recipes/RecipeDefinitions/ASP.NETAppAppRunner.recipe
```diff
@@ -100,6 +100,7 @@
             "TypeHint": "AppRunnerService",
             "AdvancedSetting": false,
             "Updatable": false,
+            "VisibleOnRedeployment": true,
```
ashovlin, line 103: Throughout, I'd probably be a little more conservative and leave anywhere we're displaying an ID/ARN/etc. visible if possible. In case a user is using this as a reference and then doing something else with the associated resources in the console or on another screen.
* AppRunner - KMS key
* Fargate - VPC Id, Load Balancer ARN
* Beanstalk - IAM role ARN, EC2 key pair

src/AWS.Deploy.CLI/ServerMode/Controllers/DeploymentController.cs
philasmar: Initially my changes were in `IsOptionSettingDisplayable` but that impacts both the CLI, server mode, summary view and detailed view of option settings which is not what I want. So I moved it to server-mode specifically.

src/AWS.Deploy.CLI/ServerMode/Controllers/DeploymentController.cs
```diff
@@ -208,7 +208,9 @@ private List<OptionSettingItemSummary> ListOptionSettingSummary(IOptionSettingHa
...
                     SummaryDisplayable = optionSettingHandler.IsSummaryDisplayable(recommendation, setting),
```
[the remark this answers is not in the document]
philasmar: I opted to leave the summary view as is to give the customer an idea of all the settings they have already selected. However, not all of them are editable, so when they click `Edit settings` they will only see the ones that are editable as well as the disabled ones that we want visible.

src/AWS.Deploy.Recipes/RecipeDefinitions/ASP.NETAppAppRunner.recipe
philasmar: Since the summary screen shows all the currently selected options, I wanted the edit screen to be mostly the updatable settings which is why I wasn't very generous with visibility of disabled settings. Curious what @normj and @96malhar think? and as you said, changing the visibility of the disabled settings is not a breaking change and we can update based on feedback.

src/AWS.Deploy.CLI/ServerMode/Controllers/DeploymentController.cs
ashovlin: Ah, that'd be good context to add as a comment, but not a deal breaker if you're not already revisiting this file again.

src/AWS.Deploy.CLI/ServerMode/Controllers/DeploymentController.cs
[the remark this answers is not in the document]
ashovlin: The highlighted `Create New` is specifically what was called out first in https://github.com/aws/aws-toolkit-visual-studio/issues/429, so I still think it'd be great to get it out of the summary. The "Create New Roles" too.

src/AWS.Deploy.Recipes/RecipeDefinitions/ASP.NETAppAppRunner.recipe
96malhar: As a user, I'd prefer to always see the AWS resources associated with my application across all configuration views (including the edit screen) even if I cannot modify them. 

To be explicit about the immutability of certain settings on a re-deployment, we should append a line to the OptionSetting description saying - _"The configured value cannot be changed during a redeployment to a existing Cloud Application"_. This line should be appended wherever `Updatable` is set to `false` and `VisibleOnRedeployment` is set to `true`

96malhar (changes requested):

src/AWS.Deploy.CLI/ServerMode/Controllers/DeploymentController.cs
philasmar: Added a comment

src/AWS.Deploy.CLI/ServerMode/Controllers/DeploymentController.cs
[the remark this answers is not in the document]
philasmar: As discussed internally, the summary view will show all currently applied settings while the edit view will show the updatable settings plus the ones we decided should be visible. In order to remove the `Create New` and similar options from the summary view, I need to add custom logic which I prefer to treat as a separate effort.

src/AWS.Deploy.Recipes/RecipeDefinitions/ASP.NETAppAppRunner.recipe
philasmar: I have made the resources that Alex mentioned visible. Also, I have experimented with adding a message to the description but it looked weird and repetetive in the toolkit. This message should be added by the Toolkit and should look different than the description so it could stand out.

96malhar (approved):

ashovlin (approved):
Good with this as an incremental improvement, but I still think we should try to get "Create" out of the summary pane before fully resolving https://github.com/aws/aws-toolkit-visual-studio/issues/429

philasmar:
> Good with this as an incremental improvement, but I still think we should try to get "Create" out of the summary pane before fully resolving [aws/aws-toolkit-visual-studio#429](https://github.com/aws/aws-toolkit-visual-studio/issues/429)

I will create a backlog item to remove the "Create" out of the summary pane.

PR #847, opened 2024-07-19
fix: remove auto self-contained build for .NET 8 beanstalk deployments

philasmar:
*Issue #, if available:*
DOTNET-7212

*Description of changes:*
In order to support deploying .NET8 apps on Beanstalk prior to Beanstalk adding .NET8 support, we had to force .NET8 apps to be self-contained in order to allow for Beanstalk support. Beanstalk has since added .NET8 support, so this PR removes this restriction to allow .NET8 apps to be deployed to Beanstalk not as self-contained apps.

This PR also improves the sorting of Beanstalk platforms to show the latest .NET versions first.

### DIFF
--- a/src/AWS.Deploy.Orchestration/Data/AWSResourceQueryer.cs
+++ b/src/AWS.Deploy.Orchestration/Data/AWSResourceQueryer.cs
@@ -516,7 +516,7 @@ namespace AWS.Deploy.Orchestration.Data
                 "Error attempting to retrieve the default VPC");
         }
 
-        public async Task<List<PlatformSummary>> GetElasticBeanstalkPlatformArns(params BeanstalkPlatformType[]? platformTypes)
+        public async Task<List<PlatformSummary>> GetElasticBeanstalkPlatformArns(string? targetFramework, params BeanstalkPlatformType[]? platformTypes)
         {
             if(platformTypes == null || platformTypes.Length == 0)
             {
@@ -557,7 +557,9 @@ namespace AWS.Deploy.Orchestration.Data
             var allPlatformSummaries = new List<PlatformSummary>();
             if (platformTypes.Contains(BeanstalkPlatformType.Linux))
             {
-                allPlatformSummaries.AddRange(await fetchPlatforms(Constants.ElasticBeanstalk.LinuxPlatformType));
+                var linuxPlatforms = await fetchPlatforms(Constants.ElasticBeanstalk.LinuxPlatformType);
+                linuxPlatforms = SortElasticBeanstalkLinuxPlatforms(targetFramework, linuxPlatforms);
+                allPlatformSummaries.AddRange(linuxPlatforms);
             }
             if (platformTypes.Contains(BeanstalkPlatformType.Windows))
             {
@@ -581,9 +583,9 @@ namespace AWS.Deploy.Orchestration.Data
             return platformVersions;
         }
 
-        public async Task<PlatformSummary> GetLatestElasticBeanstalkPlatformArn(BeanstalkPlatformType platformType)
+        public async Task<PlatformSummary> GetLatestElasticBeanstalkPlatformArn(string? targetFramework, BeanstalkPlatformType platformType)
         {
-            var platforms = await GetElasticBeanstalkPlatformArns(platformType);
+            var platforms = await GetElasticBeanstalkPlatformArns(targetFramework, platformType);
 
             if (!platforms.Any())
             {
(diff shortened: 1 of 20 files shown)


philasmar:
Add a platform version check for the version that added .NET8. If a user is redeploying we shouldn't break them.

src/AWS.Deploy.Orchestration/Data/AWSResourceQueryer.cs
```diff
@@ -593,15 +595,73 @@ public async Task<PlatformSummary> GetLatestElasticBeanstalkPlatformArn(Beanstal
...
+                if (!int.TryParse(frameworkSplit[1], out var dotnetVersion))
+                {
+                    dotnetVersionMap[platform.PlatformArn] = 0;
+                    continue;
+                }
+
+                dotnetVersionMap[platform.PlatformArn] = dotnetVersion;
+            }
 
+            return platforms.OrderByDescending(x => new Version(x.PlatformVersion)).ThenByDescending(x => dotnetVersionMap[x.PlatformArn]).ToList();
```
96malhar, line 658: I think this condition should be inverted. We first want to sort the platforms in descending order of the .NET version and then in descending order of platform versions. 
```suggestion
            return platforms.OrderByDescending(x => dotnetVersionMap[x.PlatformArn]).ThenByDescending(x => new Version(x.PlatformVersion)).ToList();
```

96malhar (changes requested):

test/AWS.Deploy.Orchestration.UnitTests/AWSResourceQueryerTests.cs
```diff
@@ -175,6 +175,85 @@ public void SortElasticBeanstalkWindowsPlatforms()
...
+                    PlatformArn = "arn:aws:elasticbeanstalk:us-west-2::platform/.NET Core running on 64bit Amazon Linux 2/2.6.0",
+                    PlatformBranchName = ".NET Core running on 64bit Amazon Linux 2",
+                    PlatformVersion = "2.6.0",
+                    PlatformOwner = "7"
+                },
+                new PlatformSummary
+                {
+                    PlatformArn = "arn:aws:elasticbeanstalk:us-west-2::platform/.NET Core running on 64bit Amazon Linux 2/2.5.7",
+                    PlatformBranchName = ".NET Core running on 64bit Amazon Linux 2",
+                    PlatformVersion = "2.5.7",
```
96malhar, line 244: If you change this `PlatformVersion` to 1000.1.0 -> This platform summary shows up on the top of sorted list. This is not intended since we want the all .NET Core platforms to show up after .NET X.

Refer to https://github.com/aws/aws-dotnet-deploy/pull/847/files#r1688563597

96malhar (approved):

src/AWS.Deploy.Orchestration/DeploymentBundleHandler.cs
```diff
@@ -108,21 +112,70 @@ public async Task PushDockerImageToECR(Recommendation recommendation, string rep
...
+        /// <summary>
+        /// The supported .NET versions on Elastic Beanstalk are dependent on the available platform versions.
+        /// These versions do not always have the required .NET runtimes installed so we need to perform extra checks
+        /// and perform a self-contained publish when creating the deployment bundle if needed.
+        /// </summary>
+        private void SwitchToSelfContainedBuildIfNeeded(Recommendation recommendation)
+        {
+            if (recommendation.Recipe.TargetService == RecipeIdentifier.TARGET_SERVICE_ELASTIC_BEANSTALK)
+            {
+                var targetFramework = recommendation.ProjectDefinition.TargetFramework ?? string.Empty;
```
ashovlin, line 124: Could we return if `targetFramework` is null or empty? Not too concerned with the performance, but it'd help readability to not have think about how the missing value is handled through all of this?

src/AWS.Deploy.Orchestration/DeploymentBundleHandler.cs
```diff
@@ -108,21 +112,70 @@ public async Task PushDockerImageToECR(Recommendation recommendation, string rep
...
+                if (beanstalkPlatformSetting != null)
+                {
+                    var beanstalkPlatformSettingValue = _optionSettingHandler.GetOptionSettingValue<string>(recommendation, beanstalkPlatformSetting);
+                    var beanstalkPlatformSettingValueSplit = beanstalkPlatformSettingValue?.Split("/");
+                    if (beanstalkPlatformSettingValueSplit?.Length != 3)
+                        throw new InvalidElasticBeanstalkPlatformException(DeployToolErrorCode.InvalidElasticBeanstalkPlatform, $"The selected Elastic Beanstalk platform version '{beanstalkPlatformSettingValue}' is invalid.");
+                    var beanstalkPlatformName = beanstalkPlatformSettingValueSplit[1];
+                    if (!Version.TryParse(beanstalkPlatformSettingValueSplit[2], out var beanstalkPlatformVersion))
+                        throw new InvalidElasticBeanstalkPlatformException(DeployToolErrorCode.InvalidElasticBeanstalkPlatform, $"The selected Elastic Beanstalk platform version '{beanstalkPlatformSettingValue}' is invalid.");
```
ashovlin, line 145: Is this format part of Elastic Beanstalk's "contract"? I would hate to penalize users by throwing an exception if EB ever changed the convention of these strings. 

I'm wondering if it'd be safer to just proceed and possibly let the deployment fail, since users could then set `Self Contained Build` themselves.

src/AWS.Deploy.Orchestration/DeploymentBundleHandler.cs
ashovlin: Also, are we skipping all of this if it's already set to self contained manually? If not, could we?

96malhar (approved):
Approved assuming you address Alex's feedback

src/AWS.Deploy.Orchestration/DeploymentBundleHandler.cs
philasmar: Done

src/AWS.Deploy.Orchestration/DeploymentBundleHandler.cs
philasmar: Done

ashovlin (approved):

PR #850, opened 2024-08-28
fix: Generated CDK projects referencing version of System.Text.Json

normj:
*Issue #, if available:*
https://github.com/aws/aws-toolkit-visual-studio/issues/462

*Description of changes:*
The generated CDK projects have `TreatWarningsAsErrors` as true which makes deployments start failing immediately when a dependency has a listed vulnerability. In this case System.Text.Json has a listed failure which causes the generated CDK project fail to build preventing deployment.

This PR updates the version of `System.Text.Json` which was actually coming in as a transitive dependency from `Microsoft.Extensions.Configuration.Json`. It also removes the `TreatWarningsAsErrors` property from the generated projects because we have no control when a package might be listed as vulnerable and we don't one deployments to just start failing when that happens. 

### DIFF
--- a/src/AWS.Deploy.Recipes/CdkTemplates/ConsoleAppECSFargateService/ConsoleAppEcsFargateService.csproj
+++ b/src/AWS.Deploy.Recipes/CdkTemplates/ConsoleAppECSFargateService/ConsoleAppEcsFargateService.csproj
@@ -4,8 +4,8 @@
     <OutputType>Exe</OutputType>
     <TargetFramework>net6.0</TargetFramework>
     <RollForward>Major</RollForward>
-    <TreatWarningsAsErrors>true</TreatWarningsAsErrors>
     <Nullable>enable</Nullable>
+    <TreatWarningsAsErrors>true</TreatWarningsAsErrors>
     <AWSProjectType>DeploymentProject</AWSProjectType>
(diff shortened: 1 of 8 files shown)


philasmar (approved):

afroz429 (approved):

src/AWS.Deploy.Recipes/CdkTemplates/ConsoleAppECSFargateService/ConsoleAppEcsFargateService.csproj
```diff
@@ -4,8 +4,8 @@
     <OutputType>Exe</OutputType>
     <TargetFramework>net6.0</TargetFramework>
     <RollForward>Major</RollForward>
-    <TreatWarningsAsErrors>true</TreatWarningsAsErrors>
     <Nullable>enable</Nullable>
+    <TreatWarningsAsErrors>true</TreatWarningsAsErrors>
```
dscpinheiro, line 8: Should this `TreatWarningsAsErrors` be removed as well? You removed it from the other templates.

philasmar: Good catch. I removed it.

dscpinheiro (approved):

PR #853, opened 2024-09-01
chore: add ability to select a supported architecture

philasmar:
*Issue #, if available:*
DOTNET-5124

*Description of changes:*
* Added a new `SupportedArchitectures` property to recipes which has 2 allowed values: `X86_64` and `Arm64`. I have set all our recipes to `X86_64` while we work on adding `Arm64` for the ones that support it.
* Added a new option setting item `EnvironmentArchitecture` which is part of the Deployment Bundle Settings. This option setting has a type hint that checks the `SupportedArchitectures` property on the recipe and prompts the user to chose one of them. By default, the architecture is `X86_64`.

image

### DIFF
--- a/src/AWS.Deploy.CLI/Commands/DeployCommand.cs
+++ b/src/AWS.Deploy.CLI/Commands/DeployCommand.cs
@@ -726,28 +726,29 @@ namespace AWS.Deploy.CLI.Commands
 
             object currentValue = _optionSettingHandler.GetOptionSettingValue(recommendation, setting);
             object? settingValue = null;
-            if (setting.AllowedValues?.Count > 0)
+
+            if (setting.TypeHint.HasValue && _typeHintCommandFactory.GetCommand(setting.TypeHint.Value) is var typeHintCommand && typeHintCommand != null)
             {
-                var userInputConfig = new UserInputConfiguration<string>(
-                    idSelector: x => x,
-                    displaySelector: x => setting.ValueMapping.ContainsKey(x) ? setting.ValueMapping[x] : x,
-                    defaultSelector: x => x.Equals(currentValue))
-                {
-                    CreateNew = false
-                };
-
-                var userResponse = _consoleUtilities.AskUserToChooseOrCreateNew(setting.AllowedValues, string.Empty, userInputConfig);
-                settingValue = userResponse.SelectedOption;
-
-                // If they didn't change the value then don't store so we can rely on using the default in the recipe.
-                if (Equals(settingValue, currentValue))
-                    return;
+                settingValue = await typeHintCommand.Execute(recommendation, setting);
             }
             else
             {
-                if (setting.TypeHint.HasValue && _typeHintCommandFactory.GetCommand(setting.TypeHint.Value) is var typeHintCommand && typeHintCommand != null)
+                if (setting.AllowedValues?.Count > 0)
                 {
-                    settingValue = await typeHintCommand.Execute(recommendation, setting);
+                    var userInputConfig = new UserInputConfiguration<string>(
+                        idSelector: x => x,
+                        displaySelector: x => setting.ValueMapping.ContainsKey(x) ? setting.ValueMapping[x] : x,
+                        defaultSelector: x => x.Equals(currentValue))
+                    {
+                        CreateNew = false
+                    };
+
+                    var userResponse = _consoleUtilities.AskUserToChooseOrCreateNew(setting.AllowedValues, string.Empty, userInputConfig);
+                    settingValue = userResponse.SelectedOption;
+
+                    // If they didn't change the value then don't store so we can rely on using the default in the recipe.
+                    if (Equals(settingValue, currentValue))
+                        return;
                 }
                 else
                 {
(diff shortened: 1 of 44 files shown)


src/AWS.Deploy.CLI/Commands/DeployCommand.cs
```diff
@@ -726,28 +726,29 @@ private async Task ConfigureDeploymentFromCli(Recommendation recommendation, Opt
...
-                    CreateNew = false
-                };
 
-                var userResponse = _consoleUtilities.AskUserToChooseOrCreateNew(setting.AllowedValues, string.Empty, userInputConfig);
-                settingValue = userResponse.SelectedOption;
-
-                // If they didn't change the value then don't store so we can rely on using the default in the recipe.
-                if (Equals(settingValue, currentValue))
-                    return;
+            if (setting.TypeHint.HasValue && _typeHintCommandFactory.GetCommand(setting.TypeHint.Value) is var typeHintCommand && typeHintCommand != null)
```
philasmar, line 730: I switched the order of the `if` statement to give priority to any Type Hint that is available on an option setting as that should preceed everything else.

normj: Is there a circumstance that an option would have both an AllowsValues collection and a typehint? Are we introducing that for this change and if so we need to know if there are toolkit ramifications.

src/AWS.Deploy.Orchestration/Orchestrator.cs
```diff
@@ -244,6 +244,13 @@ public async Task ApplyAllReplacementTokens(Recommendation recommendation, strin
                 recommendation.AddReplacementToken(Constants.RecipeIdentifier.REPLACE_TOKEN_DEFAULT_CONTAINER_PORT, defaultPort);
                 recommendation.DeploymentBundle.DockerfileHttpPort = defaultPort;
             }
+            if (recommendation.ReplacementTokens.ContainsKey(Constants.RecipeIdentifier.REPLACE_TOKEN_DEFAULT_ENVIRONMENT_ARCHITECTURE))
+            {
+                if (_dockerEngine == null)
```
normj, line 249: Why is docker required for architecture? What if we are building a deployment package for Beanstalk?

normj (commented):
I'm assuming it is future PRs to the feature branch that will update the CDK projects to support ARM setup.

src/AWS.Deploy.Orchestration/Orchestrator.cs
philasmar: Docker is not required. I removed the check.

src/AWS.Deploy.CLI/Commands/DeployCommand.cs
philasmar: Originally I didn't have a type hint for the new architecture setting and then I added a typehint for more customization which didn't work, which is when I noticed that the TypeHint should override any default behavior we have on the setting. So for correctness, I have updated the behavior. @awschristou Do you know if there are any issues in the toolkit if we update this behavior? (I don't think there would be any issues currently since we don't have a setting that exercises this user case)

src/AWS.Deploy.CLI/Commands/TypeHints/EnvironmentArchitectureCommand.cs
```diff
@@ -0,0 +1,66 @@
...
+        {
+            var resourceTable = new TypeHintResourceTable
+            {
+                Columns = new List<TypeHintResourceColumn>()
+                {
+                    new TypeHintResourceColumn("Architecture")
+                }
+            };
+
+            foreach (var value in recommendation.Recipe.SupportedArchitectures ?? new List<SupportedArchitecture> { SupportedArchitecture.X86_64 })
```
awschristou, line 34: You may want to update this , DeploymentBundle.cs, and Orchestrator.cs to use a constant representing the default architecture. This will make it easier to find "all the places" where fallback/initialization handling is applied.

src/AWS.Deploy.CLI/Commands/DeployCommand.cs
awschristou: I'm a little rusty on this. I believe that TypeHint is a cue for the Toolkit to use custom UIs for a specific property instead of a generic property editor.

I was looking through the Toolkit Code, and AllowedValues is only used in situations where the ValueMapping field does not contain any values.

I think that this should be okay, but I recommend testing out the updated code in the Toolkit to be sure.

philasmar: I tested the changes in the toolkit, though not exhaustively, and there were no issues. Also, once the feature is completed, we will do a bug bash for the toolkit to verify the changes prior to release.

src/AWS.Deploy.CLI/Commands/TypeHints/EnvironmentArchitectureCommand.cs
philasmar: Done

normj (approved):

96malhar (approved):

PR #861, opened 2024-09-18
fix: fix an issue causing container deployments to fail when run on an ARM-based system

philasmar:
*Issue #, if available:*
DOTNET-7715

*Description of changes:*
* Updated the default `Dockerfile` template to include a fix for the hanging `dotnet restore` in docker build. This follows the following Microsoft blog https://devblogs.microsoft.com/dotnet/improving-multiplatform-container-support/. This fix only applies to .NET7 and newer since the `-a` argument was added in .NET7.
* Added a 2nd `Dockerfile` template which maintains the current behavior for .NET6 and older.
* Fixed an issue where the correct version of `AWS.Deploy.Recipes.CDK.Common` is not used in the CDK projects which was broken after we dropped `Nerdback.GitVersioning`. This issue does not currently exist in production. It is localized to the `dev` branch.
* This update will change the behavior of existing customers with generated dockerfiles. If they used the deploy tool to generate a dockerfile, if they try to build that image on an ARM-based system, the docker build will hang. As opposed to the previous behavior where the docker build succeeds but the deployment fails. This change does not break an existing use case, but changes where the issue occurs. The fix is to simply delete the dockerfile and let the deploy tool generate a new one.

### DIFF
--- a/src/AWS.Deploy.DockerEngine/ProjectUtilities.cs
+++ b/src/AWS.Deploy.DockerEngine/ProjectUtilities.cs
@@ -1,7 +1,6 @@
 // Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
 // SPDX-License-Identifier: Apache-2.0
 
-using System.IO;
 using System.Reflection;
 using AWS.Deploy.Common;
 using AWS.Deploy.Common.Extensions;
@@ -12,6 +11,7 @@ namespace AWS.Deploy.DockerEngine
     {
         private const string DockerFileConfig = "AWS.Deploy.DockerEngine.Properties.DockerFileConfig.json";
         private const string DockerfileTemplate = "AWS.Deploy.DockerEngine.Templates.Dockerfile.template";
+        private const string DockerfileTemplate_Net6 = "AWS.Deploy.DockerEngine.Templates.Dockerfile.Net6.template";
 
         /// <summary>
         /// Retrieves the Docker File Config
@@ -31,9 +31,22 @@ namespace AWS.Deploy.DockerEngine
         /// <summary>
         /// Reads dockerfile template file
         /// </summary>
-        internal static string ReadTemplate()
+        internal static string ReadTemplate(string? targetFramework)
         {
-            var template = Assembly.GetExecutingAssembly().ReadEmbeddedFile(DockerfileTemplate);
+            string templateLocation;
+            switch (targetFramework)
+            {
+                case "net6.0":
+                case "net5.0":
+                case "netcoreapp3.1":
+                    templateLocation = DockerfileTemplate_Net6;
+                    break;
+
+                default:
+                    templateLocation = DockerfileTemplate;
+                    break;
+            }
+            var template = Assembly.GetExecutingAssembly().ReadEmbeddedFile(templateLocation);
 
             if (string.IsNullOrWhiteSpace(template))
             {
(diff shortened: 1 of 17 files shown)


src/AWS.Deploy.DockerEngine/ProjectUtilities.cs
```diff
@@ -31,9 +31,22 @@ internal static string ReadDockerFileConfig()
         /// <summary>
         /// Reads dockerfile template file
         /// </summary>
-        internal static string ReadTemplate()
+        internal static string ReadTemplate(string? targetFramework)
         {
-            var template = Assembly.GetExecutingAssembly().ReadEmbeddedFile(DockerfileTemplate);
+            string templateLocation;
+            switch (targetFramework)
```
philasmar, line 37: I am specifically checking the following target framework to maintain consistency with other parts of the codebase that also have the same check for different purposes.

src/AWS.Deploy.DockerEngine/Templates/Dockerfile.Net6.template
```diff
@@ -0,0 +1,26 @@
+FROM {docker-base-image} AS base{non-root-user}
```
philasmar, line 1: The fix to container deployments only applies to .NET 7 and newer, so this dockerfile template is the version we currently use and will continue to be used for .NET6 and older targets.

src/AWS.Deploy.Orchestration/AWS.Deploy.Orchestration.csproj
```diff
@@ -7,10 +7,6 @@
     <RootNamespace>AWS.Deploy.Orchestration</RootNamespace>
   </PropertyGroup>
 
-  <ItemGroup>
```
philasmar, line 10: I added a project reference to `AWS.Deploy.Recipes.CDK.Common` so this reference is no longer needed.

src/AWS.Deploy.Orchestration/DeploymentBundleHandler.cs
```diff
@@ -74,6 +74,11 @@ public async Task BuildDockerImage(CloudApplication cloudApplication, Recommenda
             DockerUtilities.TryGetAbsoluteDockerfile(recommendation, _fileManager, _directoryManager, out var dockerFile);
 
             var dockerBuildCommand = $"docker build -t {imageTag} -f \"{dockerFile}\"{buildArgs} .";
+            if (RuntimeInformation.OSArchitecture != Architecture.X64)
```
philasmar, line 77: We don't currently support running tests on non X64 architectures, so this code path will not be tested in our CI. However, this code path is covered using existing tests when run on ARM.

src/AWS.Deploy.Orchestration/TemplateEngine.cs
```diff
@@ -67,7 +67,7 @@ public void GenerateCDKProjectFromTemplate(Recommendation recommendation, Orches
             var templateParameters = new Dictionary<string, string> {
                 // CDK Template projects can parameterize the version number of the AWS.Deploy.Recipes.CDK.Common package. This avoid
                 // projects having to be modified every time the package version is bumped.
-                { "AWSDeployRecipesCDKCommonVersion", FileVersionInfo.GetVersionInfo(typeof(Constants.CloudFormationIdentifier).Assembly.Location).ProductVersion
+                { "AWSDeployRecipesCDKCommonVersion", FileVersionInfo.GetVersionInfo(typeof(AWS.Deploy.Recipes.CDK.Common.CDKRecipeSetup).Assembly.Location).ProductVersion
```
philasmar, line 70: Since we dropped Nerdbank.GitVersioning, this code was bugged as it always returned 1.0.0. This is because `Constants.CloudFormationIdentifier` is part of `AWS.Deploy.Orchestration` which is not versioned. This is why I added a project reference to `AWS.Deploy.Recipes.CDK.Common` and used `AWS.Deploy.Recipes.CDK.Common.CDKRecipeSetup` which is versioned by `AutoVer`.

96malhar (approved):

normj (approved):

Issue #877, opened 2024-10-15
Add support for LaunchTemplates in aws-dotnet-deploy tool. LaunchConfigurations are not supported anymore


chitikjb:
### Describe the bug

Starting on October 1, 2024, new accounts will not be able to create new launch configurations by using the console, API, CLI, and CloudFormation. https://docs.aws.amazon.com/autoscaling/ec2/userguide/create-launch-config.html

This created an issue when I used aws-dotnet-deploy tool in my devops pipeline to deploy a .NET 8 webAPI to elastic beanstalk (linux) environment in a brand new AWS account. Currently the tool only supports LaunchConfiguration and not LaunchTemplates, that causes the issue.

### Regression Issue

- [ ] Select this option if this issue appears to be a regression.

### Expected Behavior

The deployment to go smoothly

### Current Behavior

The deployment fails because LaunchConfigurations are not supported anymore in new aws accounts and the tool uses LaunchConfigurations in it's autoscaling group for load balanced EB deployments.

### Reproduction Steps

use aws-dotnet-deploy tool to deploy a .NET 8 Web application/WebAPI to a load balanced Elastic Beanstalk environment.

### Version used

1.26.1

### Operating System and version

Amazon Linux 2

ashishdhingra:
Yes, this is something that we should address.

bhoradc:
The changes are addressed in Pull Request #877. Marking as completed.

PR #878, opened 2024-10-16
Add support for .NET 9 and support for disabling IMDS v1

normj:
*Issue #, if available:*
https://github.com/aws/aws-dotnet-deploy/issues/877

*Description of changes:*
EC2 made a change that new accounts created after Oct 1 2024. This means today if an account created after Oct 1st tries to deploy to Beanstalk using our tooling it will fail because it can't create a LaunchConfiguration.

To force Beanstalk you need to set one of the following 4 settings.
* RootVolumeType
* BlockDeviceMappings
* DisableIMDSv1 -> `true`
* EnableSpot
(For reference the Beantalk Developer Guide: https://docs.aws.amazon.com/elasticbeanstalk/latest/dg/environments-cfg-autoscaling-launch-templates.html)

None of these setting are currently settable in the tooling. Ideally we should add support for now I added `DisableIMDSv1`. The other settings require more settings to be set when enabled and there isn't an obvious default.

The  `DisableIMDSv1` is a boolean property but we have to be careful not to automatically set this to `true` on existing environments when users do a redeployment. That could break environments. The compromise I did was modeled the property as 3 state value: Enabled, Disabled and Default. 

The `Default` value is the the default for the new setting in the recipe. Using `Default` means in the CDK project to set `DisableIMDSv1` for new deployments but for redeployments leave the setting alone. That way existing users when they upgrade to the new version of the tooling no change will be made to their environments when they redeploy. This required adding metadata being passed into the CDK project whether it was being run for a new deployment or not.

I decided to make the changes to add .NET 9 support while I was waiting for test deployments to finish. The change was just to add the .NET 9 target in the recipe rules files and update the docker manifest.

### Testing

I have run all of the unit and integ tests. Before making any changes I created a new AWS account and reproduced the failure situation. After making the changes I was able to do deployments and redeployments with both my old account and new account. I also ran through .NET 9 deployments.

### DIFF
--- a/src/AWS.Deploy.Orchestration/DeploymentBundleHandler.cs
+++ b/src/AWS.Deploy.Orchestration/DeploymentBundleHandler.cs
@@ -133,9 +133,9 @@ namespace AWS.Deploy.Orchestration
                 if (string.IsNullOrEmpty(targetFramework))
                     return;
 
-                // Elastic Beanstalk doesn't currently have .NET 7 preinstalled.
-                var unavailableFramework = new List<string> { "net7.0" };
-                var frameworkNames = new Dictionary<string, string> { { "net7.0", ".NET 7" } };
+                // Elastic Beanstalk doesn't currently have .NET 7 and 9 preinstalled.
+                var unavailableFramework = new List<string> { "net7.0", "net9.0" };
+                var frameworkNames = new Dictionary<string, string> { { "net7.0", ".NET 7" }, { "net9.0", ".NET 9" } };
                 if (unavailableFramework.Contains(targetFramework))
                 {
                     _interactiveService.LogInfoMessage($"Using self-contained publish since AWS Elastic Beanstalk does not currently have {frameworkNames[targetFramework]} preinstalled");
--- a/src/AWS.Deploy.Recipes/CdkTemplates/AspNetAppElasticBeanstalkLinux/Generated/Recipe.cs
+++ b/src/AWS.Deploy.Recipes/CdkTemplates/AspNetAppElasticBeanstalkLinux/Generated/Recipe.cs
@@ -33,6 +33,10 @@ namespace AspNetAppElasticBeanstalkLinux
         public const string LOADBALANCERTYPE_APPLICATION = "application";
         public const string LOADBALANCERSCHEME_PUBLIC = "public";
 
+        public const string IMDS_V1_DEFAULT = "Default";
+        public const string IMDS_V1_DISABLED = "Disabled";
+        public const string IMDS_V1_ENABLED = "Enabled";
+
         public const string REVERSEPROXY_NGINX = "nginx";
         
         public const string ENHANCED_HEALTH_REPORTING = "enhanced";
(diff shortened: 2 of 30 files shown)


src/AWS.Deploy.Orchestration/DeploymentBundleHandler.cs
```diff
@@ -134,8 +134,8 @@ private void SwitchToSelfContainedBuildIfNeeded(Recommendation recommendation)
                     return;
 
                 // Elastic Beanstalk doesn't currently have .NET 7 preinstalled.
```
philasmar, line 136: nit: forgot to add .NET9 to the comment

src/AWS.Deploy.Recipes/CdkTemplates/AspNetAppElasticBeanstalkLinux/Generated/Recipe.cs
```diff
@@ -238,6 +242,18 @@ private void ConfigureBeanstalkEnvironment(Configuration settings, string beanst
                    }
                 };
 
+            if (newDeployment ||
```
philasmar, line 245: if `newDeployment` is true, then the `if` statement is true without checking the 2nd condition.
If `newDeployment` is false, the 2nd condition will also be false since you are using &&

src/AWS.Deploy.Recipes/CdkTemplates/AspNetAppElasticBeanstalkWindows/Generated/Recipe.cs
```diff
@@ -238,6 +242,18 @@ private void ConfigureBeanstalkEnvironment(Configuration settings, string beanst
                    }
                 };
 
+            if (newDeployment ||
```
philasmar, line 245: same comment

philasmar (changes requested):

philasmar:
Could you also add a `change file` to this PR?

src/AWS.Deploy.Orchestration/DeploymentBundleHandler.cs
normj: Fixed

src/AWS.Deploy.Recipes/CdkTemplates/AspNetAppElasticBeanstalkLinux/Generated/Recipe.cs
normj: The second condition was supposed to be `!newDeployment` for if doing a redeployment then skip if disable IMDSv1 is set to default. I had done some renaming refactoring where the variable used to be called `performRedeployment` and switched the naming to be `newDeployment` thinking that looks cleaner but forgot to update the inverse logic here.

src/AWS.Deploy.Recipes/CdkTemplates/AspNetAppElasticBeanstalkWindows/Generated/Recipe.cs
normj: Fixed

philasmar (approved):

96malhar (approved):

PR #905, opened 2025-04-16
feat: add validator for beanstalk when switching architecture on redeployment

philasmar:
*Issue #, if available:*
DOTNET-8065

*Description of changes:*
* Add a recipe validator for the Beanstalk recipe which checks if the `Environment Architecture` is compatible with the selected `Instance Type`. If no `Instance Type` is selected and CDK is selecting the instance, I am querying Beanstalk to fetch the Instance Type in order to check the architecture.

### DIFF
--- a/src/AWS.Deploy.Common/Recipes/Validation/RecipeValidators/BeanstalkInstanceTypeValidator.cs
+++ b/src/AWS.Deploy.Common/Recipes/Validation/RecipeValidators/BeanstalkInstanceTypeValidator.cs
@@ -10,11 +10,18 @@ using AWS.Deploy.Common.Data;
 
 namespace AWS.Deploy.Common.Recipes.Validation;
 
+/// <summary>
+/// Validates that the chosen Elastic Beanstalk instance type(s) support
+/// the target environment architecture in a deployment recommendation.
+/// </summary>
 public class BeanstalkInstanceTypeValidator : IRecipeValidator
 {
     private readonly IOptionSettingHandler _optionSettingHandler;
     private readonly IAWSResourceQueryer _awsResourceQueryer;
 
+    /// <summary>
+    /// Initializes a new instance of the <see cref="BeanstalkInstanceTypeValidator"/> class.
+    /// </summary>
     public BeanstalkInstanceTypeValidator(IAWSResourceQueryer awsResourceQueryer, IOptionSettingHandler optionSettingHandler)
     {
         _awsResourceQueryer = awsResourceQueryer;
@@ -25,10 +32,13 @@ public class BeanstalkInstanceTypeValidator : IRecipeValidator
     public string ApplicationNameOptionSettingsId { get; set; } = "BeanstalkApplication.ApplicationName";
     public string EnvironmentNameOptionSettingsId { get; set; } = "BeanstalkEnvironment.EnvironmentName";
 
+    /// <summary>
+    /// Validates that the instance type(s) configured for an Elastic Beanstalk
+    /// deployment recommendation support the deployment bundle’s environment architecture.
+    /// </summary>
     public async Task<ValidationResult> Validate(Recommendation recommendation, IDeployToolValidationContext deployValidationContext)
     {
         string? instanceType;
-
         try
         {
             instanceType = _optionSettingHandler.GetOptionSettingValue<string>(recommendation, _optionSettingHandler.GetOptionSetting(recommendation, InstanceTypeOptionSettingsId));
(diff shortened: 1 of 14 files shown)


src/AWS.Deploy.Common/Recipes/Validation/RecipeValidators/BeanstalkInstanceTypeValidator.cs
```diff
@@ -0,0 +1,113 @@
...
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+using AWS.Deploy.Common.Data;
+
+namespace AWS.Deploy.Common.Recipes.Validation;
+
+public class BeanstalkInstanceTypeValidator : IRecipeValidator
```
GarrettBeatty, line 13: can you add docs for this class and other classes

src/AWS.Deploy.Common/Recipes/Validation/RecipeValidators/BeanstalkInstanceTypeValidator.cs
```diff
@@ -0,0 +1,113 @@
...
+    {
+        _awsResourceQueryer = awsResourceQueryer;
+        _optionSettingHandler = optionSettingHandler;
+    }
+
+    public string InstanceTypeOptionSettingsId { get; set; } = "InstanceType";
+    public string ApplicationNameOptionSettingsId { get; set; } = "BeanstalkApplication.ApplicationName";
+    public string EnvironmentNameOptionSettingsId { get; set; } = "BeanstalkEnvironment.EnvironmentName";
+
+    public async Task<ValidationResult> Validate(Recommendation recommendation, IDeployToolValidationContext deployValidationContext)
```
GarrettBeatty, line 28: can we split `Validate` into multiple sub functions - it is pretty long right now

test/AWS.Deploy.CLI.IntegrationTests/WebAppNoDockerFileTests.cs
```diff
@@ -164,7 +164,29 @@ public async Task BeanstalkArmDeployment()
 
             // Verify stack exists in list of deployments
             var listStdOut = _interactiveService.StdOutReader.ReadAllLines().Select(x => x.Split()[0]).ToList();
-            Assert.Contains(listStdOut, (deployment) => _stackName.Equals(deployment));
+            Assert.Contains(listStdOut, (deployment) => _stackName.Equals(deployment));
```
GarrettBeatty, line 167: do we want/need a test for the opposite of this - arm to x86?

GarrettBeatty (changes requested):

src/AWS.Deploy.Common/Recipes/Validation/RecipeValidators/BeanstalkInstanceTypeValidator.cs
philasmar: Done

src/AWS.Deploy.Common/Recipes/Validation/RecipeValidators/BeanstalkInstanceTypeValidator.cs
philasmar: I didn't split the function as I am returning validation results as well as retrieving values at every step and didn't want to overengineer it in order to make the split. However, I did add comments at every step to explain the logic.

test/AWS.Deploy.CLI.IntegrationTests/WebAppNoDockerFileTests.cs
philasmar: I decided against it since all our deployments default to X86 and didn't want to add more deployments/redeployments to further slow down our already slow test suite. The logic is the same for both x86 and arm so I didn't think it would be a must as i am already testing the code path. If you feel strongly about it, i could add that test.

src/AWS.Deploy.Common/Recipes/Validation/RecipeValidators/BeanstalkInstanceTypeValidator.cs
```diff
@@ -0,0 +1,132 @@
...
+            // We check if the selected instance types support the architecture we are trying to deploy to.
+            if (!environmentArchitectures.Contains(recommendation.DeploymentBundle.EnvironmentArchitecture.ToString(), StringComparer.InvariantCultureIgnoreCase))
+            {
+                return await ValidationResult.FailedAsync(
+                    $"The Elastic Beanstalk application is currently using the Instance Types '{string.Join(",", environmentInstanceTypes)}' " +
+                    $"which do not support the currently selected Environment Architecture '{recommendation.DeploymentBundle.EnvironmentArchitecture}'. " +
+                    "Please select an Instance Type that supports the currently selected Environment Architecture.");
+            }
+        }
+        else
```
GarrettBeatty, line 115: this `else` statement handles 3 scenarios right?

1. the user tries to change the instance type without changing the architecture
2. The user tries to change the architecture without changing the instance type (but they still explicitly set the existing one)
3. the user tries to change the architecture and instance type at the same time

src/AWS.Deploy.Common/Recipes/Validation/ValidatorFactory.cs
```diff
@@ -68,7 +68,8 @@ public ValidatorFactory(IServiceProvider serviceProvider)
         private static readonly Dictionary<RecipeValidatorList, Type> _recipeValidatorTypeMapping = new()
         {
             { RecipeValidatorList.FargateTaskSizeCpuMemoryLimits, typeof(FargateTaskCpuMemorySizeValidator) },
-            { RecipeValidatorList.ValidDockerfilePath, typeof(DockerfilePathValidator) }
+            { RecipeValidatorList.ValidDockerfilePath, typeof(DockerfilePathValidator) },
+            { RecipeValidatorList.BeanstalkInstanceType, typeof(BeanstalkInstanceTypeValidator) }
```
GarrettBeatty, line 72: this validator is configured to always run, regardless of what settings the user actually changed right?

src/AWS.Deploy.Common/Recipes/Validation/RecipeValidators/BeanstalkInstanceTypeValidator.cs
philasmar: Yes that is true, but more broadly this always checks that the instance type and environment architecture are compatible in the case where an instance type is explicitly defined.

src/AWS.Deploy.Common/Recipes/Validation/ValidatorFactory.cs
philasmar: Yes, Recipe validators always run prior to the start of the deployment. They are not tied to a specific setting. They are usually used to check if multiple settings are compatible.

GarrettBeatty (approved):

AlexDaines (approved):

PR #915, opened 2025-04-30
chore: self-contain unsupported .net versions on beanstalk

philasmar:
*Issue #, if available:*
DOTNET-8094

*Description of changes:*
* Instead of tracking the unsupported dotnet versions, we are now tracking the supported ones and everything else will be deployed as self-contained.
* We are also tracking the retired .net versions to provide the user messaging informing them that they are using an out of support version

### DIFF
--- a/.autover/changes/1c879063-ec4b-4330-80a2-3f2116812150.json
+++ b/.autover/changes/1c879063-ec4b-4330-80a2-3f2116812150.json
@@ -1,0 +1,11 @@
+{
+  "Projects": [
+    {
+      "Name": "AWS.Deploy.CLI",
+      "Type": "Patch",
+      "ChangelogMessages": [
+        "Automatically deploy unsupported .NET versions using a self-contained build to Elastic Beanstalk"
+      ]
+    }
+  ]
+}
--- a/.autover/changes/1c879063-ec4b-4330-80a2-3f2116812150.json
+++ b/.autover/changes/1c879063-ec4b-4330-80a2-3f2116812150.json
@@ -2,7 +2,7 @@
   "Projects": [
     {
       "Name": "AWS.Deploy.CLI",
-      "Type": "Patch",
+      "Type": "Minor",
       "ChangelogMessages": [
         "Automatically deploy unsupported .NET versions using a self-contained build to Elastic Beanstalk"
       ]
--- a/src/AWS.Deploy.Orchestration/DeploymentBundleHandler.cs
+++ b/src/AWS.Deploy.Orchestration/DeploymentBundleHandler.cs
@@ -143,7 +143,7 @@ namespace AWS.Deploy.Orchestration
                     if (retiredFrameworks.Contains(targetFramework))
                     {
                         _interactiveService.LogErrorMessage($"The version of .NET that you are targeting has reached its end-of-support and has been retired by Elastic Beanstalk");
-                        _interactiveService.LogInfoMessage($"Using self-contained publish to deploy your application on a newer Elastic Beanstalk platform with a supported version of .NET");
+                        _interactiveService.LogInfoMessage($"Using self-contained publish to include the out of support version of .NET used by this application with the deployment bundle");
                     }
                     else
                     {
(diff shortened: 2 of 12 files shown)


src/AWS.Deploy.Orchestration/DeploymentBundleHandler.cs
```diff
@@ -135,12 +135,20 @@ private void SwitchToSelfContainedBuildIfNeeded(Recommendation recommendation)
                 if (string.IsNullOrEmpty(targetFramework))
                     return;
 
-                // Elastic Beanstalk doesn't currently have .NET 7 and 9 preinstalled.
-                var unavailableFramework = new List<string> { "net7.0", "net9.0" };
-                var frameworkNames = new Dictionary<string, string> { { "net7.0", ".NET 7" }, { "net9.0", ".NET 9" } };
-                if (unavailableFramework.Contains(targetFramework))
+                // Elastic Beanstalk currently has .NET 8 and 9 supported on their platforms.
+                var supportedFrameworks = new List<string> { "net8.0", "net9.0" };
```
GarrettBeatty, line 139: is there any way to query beanstalk to see what their supported versions are instead of updating this list every time?

.autover/changes/1c879063-ec4b-4330-80a2-3f2116812150.json
```diff
@@ -0,0 +1,11 @@
+{
+  "Projects": [
+    {
+      "Name": "AWS.Deploy.CLI",
+      "Type": "Patch",
```
GarrettBeatty, line 5: should this be minor version since we are changing the behavior slightly?

src/AWS.Deploy.Orchestration/DeploymentBundleHandler.cs
```diff
@@ -135,12 +135,20 @@ private void SwitchToSelfContainedBuildIfNeeded(Recommendation recommendation)
...
+                var retiredFrameworks = new List<string> { "netcoreapp3.1", "net5.0", "net6.0", "net7.0" };
+                if (!supportedFrameworks.Contains(targetFramework))
                 {
-                    _interactiveService.LogInfoMessage($"Using self-contained publish since AWS Elastic Beanstalk does not currently have {frameworkNames[targetFramework]} preinstalled");
+                    if (retiredFrameworks.Contains(targetFramework))
+                    {
+                        _interactiveService.LogErrorMessage($"The version of .NET that you are targeting has reached its end-of-support and has been retired by Elastic Beanstalk");
+                        _interactiveService.LogInfoMessage($"Using self-contained publish to deploy your application on a newer Elastic Beanstalk platform with a supported version of .NET");
```
normj, line 146: I find this sentence to the user confusing. I think saying "newer Elastic Beanstalk platform with a supported version of .NET" makes me think you are trying to tell me that newer version is going to be used somehow. Maybe something like the following:

```
Using self-contained publish to include the out support version of .NET used by this application with the deployment bundle.
```

normj (changes requested):

.autover/changes/1c879063-ec4b-4330-80a2-3f2116812150.json
philasmar: we are fixing the behavior so this should be fine to do a patch. however, since we mostly always do Minors for this package, i'll change to minor.

src/AWS.Deploy.Orchestration/DeploymentBundleHandler.cs
philasmar: Not aware of anywhere besides their documentation. We can't rely on the platform name since they also have .NET Core which supports different versions.

src/AWS.Deploy.Orchestration/DeploymentBundleHandler.cs
philasmar: Fixed

normj (approved):

GarrettBeatty (approved):

